PR #15, opened 2025-08-20
Added scripts to generate tpch data via tpchgen-cli.  

misiugodfrey:
Alters our existing data generation process for TPC-H data to generate data using tpchgen + pyarrow.parquet rewrite steps, which despite the rewrite is still substantially faster than duckdb, and has a lower memory footprint (exact differences will depend on the number of threads used).

The second part is a new python script (based on `rewrite_parquet.py`) that will rewrite the parquet files in a directory to conform with the current configurations:
- Rewrites decimals as doubles
- Rewrites some int64 columns as int32 (ones that duckdb generates as int32)

This should slot directly into the existing data generation script, with the only UI changes being new flags that can be specified.

Example usage: `python generate_data_files.py --benchmark-type tpch --data-dir-path /Data/tpch-sf100 --scale-factor 100 --verbose --convert-decimals-to-floats`

Future work:
Additional configurations can be added to the parquet rewrite as we decide on them.
Change the rewrite code to use cudf?


simoneves (changes requested):
Adapt to use the new Python "set up environment using venv or miniforge" stuff from #36

paul-aiyedun (approved):
Looks good to me.

simoneves (approved):
LGTM no more concerns, apologies for delay

PR #28, opened 2025-09-03
Add thread count option to CPU and GPU Presto start scripts

paul-aiyedun:


paul-aiyedun:
> Can we use a common flag for specifying `num_threads` parameter? `build_velox.sh` script uses `-j` but this script uses `-t` . I would prefer `-j` since it looks like a universal flag for denoting `num_threads`

I have updated the flag name to match the other scripts.

paul-aiyedun:
> LGTM, my only recommendation would be potentially to lower the default value.

The default value is based on `nproc`, so how high or low this value is will depend on the machine.

mattgara:
> Some time ago I proposed `max(nproc, 8)` as the default.

Surely you mean `min(nproc,8)` :-) but otherwise +1

simoneves:
I'm retracting this idea as it's a bad one whether `min` or `max`! :)

The important points are
(1) On a GHA Runner, use ALL available cores (because the VM will only present as many as it has)
(2) On a dev machine, use 3/4 of the cores, so that the machine remains usable during a build
So perhaps default to (2) but enable (1) with an option, which is set in the GHA Actions?

Avinash-Raj:
> Some time ago I proposed `max(nproc, 8)` as the default. This may not still be ideal in the case of laptops with smaller core counts, but we also need to make sure we're using ALL the cores presented by a GHA Runner VM.
> 
> EDIT: I meant `min(nproc, 8)`. As many as are available, but no more than 8.

As @simoneves suggested, we have to modify our existing scripts to accept `NUM_THREADS` value from env var. So that from CI workflow, we can make use of all the cores by calling the target script by `NUM_THREADS=$(nproc) ./build_velox.sh`

mattgara:
> (2) On a dev machine, use 3/4 of the cores, so that the machine remains usable during a build
> So perhaps default to (2) but enable (1) with an option, which is set in the GHA Actions?

I think the practical issue with using 3/4 of cores on anything less than a server machine is you will likely start hitting OOMs when you start linking. 

But this is less an issue for Presto builds which are comparatively light when compared with full Velox builds. 

A parrallel issue on a server machine with using 3/4 cores is if its a shared resource that might be too much. Users will have to pay attention to this themselves I suppose.

simoneves:
In other words, there is no good answer to this!

If we agree on different behavior for interactive vs batch then how about:
  Interactive: Half the cores (leaves machine usable and hopefully doesn't OOM)
  Batch: All the presented cores with a max of (say) 16

PR #30, opened 2025-09-03
Velox Build & Test Script Fixes

simoneves:


velox/docker/adapters_build.dockerfile
```diff
@@ -64,12 +65,21 @@ RUN if [ "$VELOX_ENABLE_BENCHMARKS" = "ON" ]; then \
       echo "Skipping nsys installation (VELOX_ENABLE_BENCHMARKS=OFF)"; \
     fi
 
+# Do the build using a cached BUILD_BASE_DIR (to allow incremental builds)
+# Then copy the results to /tmp
 RUN --mount=type=bind,source=velox,target=/workspace/velox,ro \
-    --mount=type=cache,target=/buildcache,sharing=locked,rw \
+    --mount=type=cache,target=/opt/velox-build,sharing=locked,rw \
```
simoneves, line 71: We now use a cached mount for the final path, so that it is directly persistent

velox/docker/adapters_build.dockerfile
```diff
@@ -64,12 +65,21 @@ RUN if [ "$VELOX_ENABLE_BENCHMARKS" = "ON" ]; then \
...
-    mkdir -p /opt/velox-build/release && \
-    cp -a "/buildcache/release/." "/opt/velox-build/release/"
+    # Build release into /opt/velox-build (cached mount)
+    make release EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS[*]}" BUILD_BASE_DIR="/opt/velox-build" && \
+    # Copy resulting build tree to /tmp
+    dnf install -y rsync && \
+    rsync -a /opt/velox-build /tmp
+
+# Move the build back from /tmp to the final non-cached BUILD_BASE_DIR
+RUN --mount=type=bind,source=velox,target=/workspace/velox,ro \
```
simoneves, line 81: ...and then, as a separate `RUN` step *without* the cached mount, we `mv` it back again.

This resolves both issues:
(1) `ctest` can find the tests
(2) You can do incremental rebuilds during dev, as `/opt/velox-build` is cached

velox/docker/adapters_build.dockerfile
```diff
@@ -37,6 +37,7 @@ ENV VELOX_DEPENDENCY_SOURCE=SYSTEM \
                       -DVELOX_ENABLE_FAISS=ON" \
     LD_LIBRARY_PATH="/opt/velox-build/release/lib:\
 /opt/velox-build/release/_deps/cudf-build:\
+/opt/velox-build/release/_deps/rmm-build:\
```
mattgara, line 40: FYI: this is also fixed in https://github.com/rapidsai/velox-testing/pull/23

velox/docker/adapters_build.dockerfile
```diff
@@ -64,12 +65,21 @@ RUN if [ "$VELOX_ENABLE_BENCHMARKS" = "ON" ]; then \
...
     set -euxo pipefail && \
-    # Build release into /buildcache
-    make release EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS[*]}" BUILD_BASE_DIR="/buildcache" && \
-    # Copy release to /opt/velox-build/release
-    mkdir -p /opt/velox-build/release && \
-    cp -a "/buildcache/release/." "/opt/velox-build/release/"
+    # Build release into /opt/velox-build (cached mount)
+    make release EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS[*]}" BUILD_BASE_DIR="/opt/velox-build" && \
+    # Copy resulting build tree to /tmp
+    dnf install -y rsync && \
```
mattgara, line 77: This (`rsync` install) should probably be a separate `RUN` step as its not dependent on any of this.

velox/docker/adapters_build.dockerfile
mattgara: Can you elaborate on what issue this resolves? 

Reading this, I can't see how this would be different than doing the copy or `mv` at the end of the build command above (similar to what was being done before?)

mattgara (changes requested):
I don't have any specific changes requested, but I'd like to understand what issue this resolves and how before going forward.

PR #39, opened 2025-09-13
Set the NVIDIA Docker runtime only for Velox GPU builds

Avinash-Raj:
This PR sets the default Docker runtime for Velox as nvidia. For CPU builds (invoked through `build_velox.sh`), the runtime is automatically changed to runc. This results in using different runtimes for Velox CPU (runc) and GPU (nvidia) builds.


.github/workflows/velox-test.yml
```diff
@@ -88,11 +88,10 @@ jobs:
           name: velox_cpu_build_log
           path: ${{ github.workspace }}/velox-testing/velox/scripts/build_velox_cpu.log
 
-      - &run-velox-tests
-        name: Run Velox Tests
+      - name: Run Velox Tests for CPU
         run: |
           cd ${{ github.workspace }}/velox-testing/velox/scripts
-          ./test_velox.sh
+          ./test_velox.sh --cpu
```
simoneves, line 94: Did you have to remove the Anchor because you needed to pass the parameter?

.github/workflows/velox-test.yml
```diff
@@ -121,4 +120,7 @@ jobs:
           name: velox_gpu_build_log
           path: ${{ github.workspace }}/velox-testing/velox/scripts/build_velox_gpu.log
 
-      - *run-velox-tests
+      - name: Run Velox Tests for GPU
+        run: |
+          cd ${{ github.workspace }}/velox-testing/velox/scripts
+          ./test_velox.sh --gpu
```
simoneves, line 126: (ditto)

velox/scripts/build_velox.sh
```diff
@@ -72,10 +73,12 @@ parse_args() {
         ;;
       --cpu)
         BUILD_WITH_VELOX_ENABLE_CUDF="OFF"
+        DOCKER_RUNTIME="runc"
```
simoneves, line 76: Is this because you need SOMETHING to put inside a `runtime:` block, and this is effectively the default? Is there a way to elide the `runtime:` block completely? No worries if not.

simoneves (approved):
LGTM just some questions for understanding

velox/scripts/test_velox.sh
```diff
@@ -3,13 +3,18 @@ set -euo pipefail
...
+DOCKER_RUNTIME="nvidia"
+
 print_help() {
   cat <<EOF
 Usage: $(basename "$0") [OPTIONS]
 
 Runs tests on the Velox adapters using ctest with parallel execution.
 
 Options:
+  --cpu                 Run tests with CPU runtime (sets DOCKER_RUNTIME=runc).
```
mattgara, line 16: I think this may start to get confusing with several layers of `--cpu` and `--gpu` flags in downstream taks.

Can we make this a command line option `--docker-runtime` and allow it three options (`cpu`, `gpu`, `auto`.) Default will be `auto`.

For `auto` this can attempt to auto-detect if an appropriate GPU is on the system ( via `nvidia-smi`) and if so set this to `gpu` otherwise `cpu`. 

Can we please add the same option to `benchmark_velox.sh` as well?

mattgara (commented):
Great find with the `DOCKER_RUNTIME` variable! Overall this looks like an elegant solution to me.

My only comment is to make this more automated to remove cognitive load from the user (by setting smart defaults) but otherwise looks great.

velox/scripts/build_velox.sh
Avinash-Raj: Yes, but let me correct it to use `runtime` only for GPU HW . 

> Docker Engine uses containerd for managing the container lifecycle, which includes creating, starting, and stopping containers. By default, containerd uses runc as its container runtime.

https://docs.docker.com/engine/daemon/alternative-runtimes/

velox/scripts/test_velox.sh
Avinash-Raj: I agree with your point. let me work on it.

Avinash-Raj:
[Workflow job](https://github.com/rapidsai/velox-testing/actions/runs/17739064878/job/50408202426) runs successfully using the code from this PR.

velox/scripts/test_velox.sh
Avinash-Raj: PR Updated. let me re-run the CI job.

Avinash-Raj:
@mattgara Fixed! also I just triggered a[ CI job](https://github.com/rapidsai/velox-testing/actions/runs/17803501509) using merged_prs as source.

velox/scripts/benchmark_velox.sh
```diff
@@ -30,6 +32,7 @@ Benchmark Options:
   -q, --queries "1 2 ..."                 Query numbers to run, specified as a space-separated list of query numbers (default: all queries for benchmark type)
   -d, --device-type "cpu gpu"             Devices to test: cpu, gpu, or "cpu gpu" (default: "cpu gpu")  
   -p, --profile BOOL                      Enable profiling: true or false (default: false)
+  --docker-runtime MODE                   Docker runtime mode: cpu|gpu|auto (default: auto)
```
paul-aiyedun, line 35: Why do we need this as a separate flag from `device-type`? Also, what is the use case for the `auto` option?

velox/scripts/test_velox.sh
```diff
@@ -48,9 +60,16 @@ parse_args() {
 
 parse_args "$@"
 
+# Resolve docker runtime from mode and export DOCKER_RUNTIME
+set_docker_runtime_from_mode "$DOCKER_RUNTIME_MODE"
```
mattgara, line 64: Will this fail because of `set -e` when this function returns 1? 

Is that the behaviour we want? Should we just fallback on CPU mode in that case?

mattgara (commented):
Just had one question but LGTM otherwise.

velox/scripts/test_velox.sh
Avinash-Raj: Good point, updated the code to fallback to the default runtime "runc"

velox/scripts/benchmark_velox.sh
paul-aiyedun: >device-type is specific to running benchmarks where as this is specific to the docker runtime

In what case would we want a docker runtime that does not match the specified `device-type`?

>auto option automatically detects the runtime on the fly by checking nvidia-smi command and set the DOCKER_RUNTIME var based upon it's results.,

I understand what it does. My question is where the auto-detection requirement is coming from. When running these benchmarks, we typically know what device type we are targeting. This appears to add complexity that may not be needed.

paul-aiyedun (changes requested):

Avinash-Raj:
@paul-aiyedun The reason why we put `auto` option is not to affect existing benchmark script if the user failed to pass the runtime explicitly, it has to automatically detect the corresponding runtime on the fly.

velox/docker/docker-compose.adapters.benchmark.yml
```diff
@@ -2,12 +2,12 @@ services:
   velox-benchmark:
     container_name: velox-benchmark
     image: velox-adapters-build:latest
-    runtime: nvidia
+    runtime: ${DOCKER_RUNTIME:-nvidia}
```
mattgara, line 5: Actually, thinking about this some more, we could probably remove the `docker-runtime` CLI option in downstream tasks, and just leave these lines in this PR. 

Then in the github runners that lack GPU capability, the can simply export `DOCKER_RUNTIME` to `runc`.

Avinash-Raj: @mattgara PR updated as per your instructions.

velox/scripts/test_velox.sh
```diff
@@ -65,7 +65,11 @@ parse_args "$@"
 
 echo "Running tests on Velox adapters..."
 echo ""
-test_cmd="ctest -j ${NUM_THREADS} --label-exclude cuda_driver --output-on-failure --no-tests=error --stop-on-failure"
+if [[ "$DOCKER_RUNTIME" == "runc" ]]; then
```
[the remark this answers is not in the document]
mattgara: @Avinash-Raj @paul-aiyedun I may be missing something, but the way I understand this is that Paul is suggesting to add the `device-type` to use in the condition in this `if` statement. IIUC, this makes sense to me.

Avinash-Raj:
@mattgara @simoneves PR updated as per your comments. Can we merge now?

PR #38, opened 2025-09-10
CI - Follow Up Velox Benchmark workflow

Avinash-Raj:

PR #41, opened 2025-09-16
Standardized way for applying patches

Avinash-Raj:


velox/scripts/test_velox.sh
```diff
@@ -50,7 +50,7 @@ parse_args "$@"
 
 echo "Running tests on Velox adapters..."
 echo ""
-test_cmd="ctest -j ${NUM_THREADS} --label-exclude cuda_driver --output-on-failure --no-tests=error --stop-on-failure"
+test_cmd="ctest -j ${NUM_THREADS} -R cudf -V"
```
simoneves, line 53: This seems unrelated to this PR. Accidental commit from some other dev?

.github/workflows/velox-test.yml
```diff
@@ -48,32 +54,13 @@ jobs:
...
-          path: velox
-
-      - &get-velox-sha
-        name: Get Velox commit SHA
-        id: velox-sha
-        run: |
-          VELOX_REF="${{ inputs.velox_commit || github.event.inputs.velox_commit }}"
-          echo "Input ref: $VELOX_REF"
-          
-          if [[ "$VELOX_REF" == "main" || "$VELOX_REF" == "master" ]]; then
```
simoneves, line 66: Where is this logic now? Is it no longer needed? Why?

simoneves (commented):
Implementation looks OK, except:
(1) Where did the checkout `main` logic go?
(2) `ctest` parameter removal seems unrelated

My main concern, though, is we're committing patch files to the repo, but accepting SHAs to build as a parameter. While it's likely that the requirement for a particular patch file may have a longer lifetime than a particular SHA (i.e. we may progress through a sequence of known-good SHAs with the same patches) this still seems wrong.

It would be better if the patch files were in the `velox` repo, so that their presence or absence was tightly linked with each Velox commit, but that's not really practical either. Even better would be if Velox didn't need any patches AT ALL! :)

Discuss.

velox/scripts/test_velox.sh
Avinash-Raj: sorry that's me only, eager to test it out with the above change.

Avinash-Raj: Fixed!

Avinash-Raj:
@simoneves 

In order to test this PR, I duplicated code from https://github.com/rapidsai/velox-testing/pull/38 PR to this one. 
That's why you can see the commits that are un-related to the PR description.

This PR should have been merged after benchmark followup PR https://github.com/rapidsai/velox-testing/pull/38 . Better if we quicken the PR approval process.

Avinash-Raj:
ok, great! If we all agree then I'll close this PR.

simoneves:
I seeded a discussion in Slack as to a better workflow that avoids the need for patching. I never liked the idea of doing it. If that would be a better overall workflow, then this PR is indeed probably superfluous... (sorry)

PR #44, opened 2025-09-16
Misiug/min python

misiugodfrey:
Some features in `generate_data_files.py` requires python 3.12, but no formal requirement was put in the scripts.

Presumably we can automatically add python 3.12 to the environment, but I figured we should add the requirements here first in case anyone else runs into this issue.

### DIFF
--- a/benchmark_data_tools/generate_data_files.py
+++ b/benchmark_data_tools/generate_data_files.py
@@ -93,7 +93,7 @@ def rearrange_directory(raw_data_path, num_partitions):
     tables = []
     for p_file in parquet_files:
         tables.append(p_file.replace(".parquet", ""))
-    
+
     for table in tables:
         Path(f"{raw_data_path}/{table}").mkdir(parents=True, exist_ok=True)
 
@@ -131,7 +131,7 @@ def get_select_query(table_name, convert_decimals_to_floats):
             get_column_projection(column_metadata, convert_decimals_to_floats)
             for column_metadata in column_metadata_rows
         ]
-        query = f"SELECT {",".join(column_projections)} FROM {table_name}"
+        query = f"SELECT {','.join(column_projections)} FROM {table_name}"
     else:
         query = f"SELECT * FROM {table_name}"
     return query
@@ -145,8 +145,8 @@ def get_column_projection(column_metadata, convert_decimals_to_floats):
     return projection
 
 if __name__ == "__main__":
-    if (sys.version_info < (3, 12)):
-        sys.exit(f"Error: generate_data_files.py requires minimum Python 3.12")
+    if (sys.version_info < (3, 10)):
+        sys.exit(f"Error: generate_data_files.py requires minimum Python 3.10")
 
     parser = argparse.ArgumentParser(
         description="Generate benchmark parquet data files for a given scale factor. "
--- a/benchmark_data_tools/generate_data_files.py
+++ b/benchmark_data_tools/generate_data_files.py
@@ -145,9 +145,6 @@ def get_column_projection(column_metadata, convert_decimals_to_floats):
     return projection
 
 if __name__ == "__main__":
-    if (sys.version_info < (3, 10)):
-        sys.exit(f"Error: generate_data_files.py requires minimum Python 3.10")
-
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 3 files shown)


simoneves (commented):
Are you sure we can safely assume this? 3.10 is the default for Ubuntu 22, so if there are any lab machines still running that, then this will fail.

paul-aiyedun:
> > What is it that requires 3.12? Can it be adapted to not?
> 
> It is this line here:
> 
> ```
> query = f"SELECT {",".join(column_projections)} FROM {table_name}"
>                        ^
> SyntaxError: f-string: expecting '}'
> ```
> 
> I'm not sure exactly what the change in python is that allows this syntax, but considering how basic a line this is, I'm sure I can rewrite it to work in python 3.10. If that's the default for lab machines, then that's probably a safer bet. @simoneves Would you prefer I pin to 3.10 and add changes to make it all work with that version?

I think changing `","` to `','` should address this issue.

benchmark_data_tools/generate_data_files.py
```diff
@@ -130,7 +131,7 @@ def get_select_query(table_name, convert_decimals_to_floats):
             get_column_projection(column_metadata, convert_decimals_to_floats)
             for column_metadata in column_metadata_rows
         ]
-        query = f"SELECT {",".join(column_projections)} FROM {table_name}"
+        query = f"SELECT {','.join(column_projections)} FROM {table_name}"
```
misiugodfrey, line 134: Required for 3.10 syntax.

benchmark_data_tools/generate_table_schemas.py
```diff
@@ -23,9 +23,10 @@ def get_table_schema(benchmark_type, table_name, schema_name, convert_decimals_t
         f"{' ' * 4}{get_column_definition(column_metadata, convert_decimals_to_floats)}"
         for column_metadata in column_metadata_rows
     ]
-    schema = (f"CREATE TABLE hive.{schema_name}.{table_name} "
-              f"(\n{",\n".join(columns_ddl_list)}\n) "
-              f"WITH (FORMAT = 'PARQUET', EXTERNAL_LOCATION = 'file:{{file_path}}')")
+    nl = '\n'
```
misiugodfrey, line 26: 3.10 does not allow backslashes inside f-string braces.  Need to outline to a variable.

benchmark_data_tools/generate_table_schemas.py
```diff
@@ -23,9 +23,10 @@ def get_table_schema(benchmark_type, table_name, schema_name, convert_decimals_t
...
+    schema = f"CREATE TABLE hive.{schema_name}.{table_name} \
```
misiugodfrey, line 27: The brace syntax used here was not supported in 3.10, needed to convert this to a single string with backslashes.

simoneves (approved):
LGTM if this now works on 3.10, which yes, I think is safer than requiring 3.12.

benchmark_data_tools/generate_table_schemas.py
paul-aiyedun: I think it might be cleaner to move `",\n".join(columns_ddl_list)` into a separate variable and use that in the f-string.

misiugodfrey: I can outline it into a variable, but I still need to move the newline into a separate variable because 3.10 does not support backslash inside f-string epressions:
```
columns_text = f"{',\n'.join(columns_ddl}"
                                              ^
SyntaxError: f-string expression part cannot include a backslash
```

benchmark_data_tools/generate_data_files.py
```diff
@@ -5,6 +5,7 @@
 import os
 import shutil
 import math
+import sys
```
paul-aiyedun, line 8: Unused import.

benchmark_data_tools/generate_table_schemas.py
paul-aiyedun: We don't need the f-string in this case i.e. `columns_text = ",\n".join(columns_ddl_list)` should work.

paul-aiyedun (approved):
I had a couple minor comments, but changes look good to me.

PR #45, opened 2025-09-16
Add support for running integration tests against existing schemas

paul-aiyedun:
* Extend integration test script to allow testing against different hostnames, port numbers, usernames, and schemas.

* Update integration test script to use functions in py_env_functions.sh.

### DIFF
--- a/presto/scripts/run_integ_test.sh
+++ b/presto/scripts/run_integ_test.sh
@@ -143,10 +143,6 @@ if [[ -n ${SCHEMA_NAME} ]]; then
   PYTEST_ARGS+=("--schema-name ${SCHEMA_NAME}")
 fi
 
-source ./common_functions.sh
-
-wait_for_worker_node_registration "$HOST_NAME" "$PORT"
-
 source ../../scripts/py_env_functions.sh
 
 trap delete_python_virtual_env EXIT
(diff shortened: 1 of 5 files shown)


presto/scripts/run_integ_test.sh
```diff
@@ -96,8 +127,32 @@ if [[ -n ${QUERIES} ]]; then
...
+fi
+
+if [[ -n ${SCHEMA_NAME} ]]; then
+  PYTEST_ARGS+=("--schema-name ${SCHEMA_NAME}")
+fi
+
 source ./common_functions.sh
 
-wait_for_worker_node_registration
+wait_for_worker_node_registration "$HOST_NAME" "$PORT"
```
simoneves, line 148: Can this move to the end of this new section, so that any Python installs can happen while waiting for the worker to start up?

simoneves (approved):
LGTM with one suggestion for overlapping work (probably only a few seconds, but...)

presto/scripts/run_integ_test.sh
paul-aiyedun: Good point. Moved.

misiugodfrey (approved):

mattgara (approved):
LGTM

simoneves (approved):
LGTM

PR #46, opened 2025-09-17
Changed tests to build duckdb tables from parquet files if they exist

misiugodfrey:
The existing integration tests created presto tables backed by parquet files in the `data` directory and ran the tpch queries against them.  These queries were also run against duckdb, where the tables were generated by duckdb's internal tpch generation tools.  The results of both queries were then compared.

This leads to incorrect comparisons when the source parquet files were generated using a different source than duckdb, as the random seed used, as well as the dictionary used to generate random content for the tables will not necessarily be the same.  Particularly the *_address and *_comment columns are almost completely different when generated via different tpch generation tools, therefore the comparison was failing.

This patch changes the integration tests so that the duckdb tables are generated by backing against the same parquet files that are used to back the presto tables, therefore guaranteeing the same data is used for both comparisons.  The query results now line up for larger scale factors.

### DIFF
--- a/benchmark_data_tools/duckdb_utils.py
+++ b/benchmark_data_tools/duckdb_utils.py
@@ -13,5 +13,21 @@ def init_benchmark_tables(benchmark_type, scale_factor):
 
     duckdb.sql(f"INSTALL {benchmark_type}; LOAD {benchmark_type}; CALL {function_name}(sf = {scale_factor});")
 
+def init_benchmark_tables_from_parquet(benchmark_type, scale_factor, data_path):
+    tables = duckdb.sql("SHOW TABLES").fetchall()
+    assert len(tables) == 0
+
+    if benchmark_type == "tpch":
+        duckdb.sql(f"CREATE TABLE customer AS SELECT * FROM '{data_path}/customer/*.parquet';")
+        duckdb.sql(f"CREATE TABLE lineitem AS SELECT * FROM '{data_path}/lineitem/*.parquet';")
+        duckdb.sql(f"CREATE TABLE nation AS SELECT * FROM '{data_path}/nation/*.parquet';")
+        duckdb.sql(f"CREATE TABLE orders AS SELECT * FROM '{data_path}/orders/*.parquet';")
+        duckdb.sql(f"CREATE TABLE part AS SELECT * FROM '{data_path}/part/*.parquet';")
+        duckdb.sql(f"CREATE TABLE partsupp AS SELECT * FROM '{data_path}/partsupp/*.parquet';")
+        duckdb.sql(f"CREATE TABLE region AS SELECT * FROM '{data_path}/region/*.parquet';")
+        duckdb.sql(f"CREATE TABLE supplier AS SELECT * FROM '{data_path}/supplier/*.parquet';")
+    else:
+        init_benchmark_tables(benchmark_type, scale_factor)
+
 def is_decimal_column(column_type):
     return bool(re.match(r"^DECIMAL\(\d+,\d+\)$", column_type))
--- a/benchmark_data_tools/generate_data_files.py
+++ b/benchmark_data_tools/generate_data_files.py
@@ -92,7 +92,7 @@ def rearrange_directory(raw_data_path, num_partitions):
     tables = []
     for p_file in parquet_files:
         tables.append(p_file.replace(".parquet", ""))
-    
+
     for table in tables:
         Path(f"{raw_data_path}/{table}").mkdir(parents=True, exist_ok=True)
 
@@ -147,21 +147,21 @@ if __name__ == "__main__":
     parser = argparse.ArgumentParser(
         description="Generate benchmark parquet data files for a given scale factor. "
                     "Only the TPC-H and TPC-DS benchmarks are currently supported.")
-    parser.add_argument("--benchmark-type", type=str, required=True, choices=["tpch", "tpcds"],
+    parser.add_argument("-b", "--benchmark-type", type=str, required=True, choices=["tpch", "tpcds"],
                         help="The type of benchmark to generate data for.")
-    parser.add_argument("--data-dir-path", type=str, required=True,
+    parser.add_argument("-d", "--data-dir-path", type=str, required=True,
                         help="The path to the directory that will contain the benchmark data files. "
                              "This directory will be created if it does not already exist.")
-    parser.add_argument("--scale-factor", type=str, required=True,
+    parser.add_argument("-s", "--scale-factor", type=str, required=True,
                         choices=["0.01", "0.1", "1", "10", "100", "1000"],
                         help="The scale factor of the generated dataset.")
-    parser.add_argument("--convert-decimals-to-floats", action="store_true", required=False,
+    parser.add_argument("-c", "--convert-decimals-to-floats", action="store_true", required=False,
                         default=False, help="Convert all decimal columns to float column type.")
     parser.add_argument("--use-duckdb", action="store_true", required=False,
                         default=False, help="Use duckdb instead of tpchgen")
-    parser.add_argument("--num-threads", type=int, required=False,
+    parser.add_argument("-j", "--num-threads", type=int, required=False,
                         default=4, help="Number of threads to generate data with tpchgen")
-    parser.add_argument("--verbose", action="store_true", required=False,
+    parser.add_argument("-v", "--verbose", action="store_true", required=False,
                         default=False, help="Extra verbose logging")
     parser.add_argument("--max-rows-per-file", type=int, required=False,
                         default=100_000_000, help="Limit number of rows in each file (creates more partitions)")
--- a/presto/scripts/run_integ_test.sh
+++ b/presto/scripts/run_integ_test.sh
@@ -32,7 +32,7 @@ EOF
 
 KEEP_TABLES=false
 
-parse_args() { 
+parse_args() {
   while [[ $# -gt 0 ]]; do
     case $1 in
       -h|--help)
--- a/presto/testing/integration_tests/test_utils.py
+++ b/presto/testing/integration_tests/test_utils.py
@@ -13,7 +13,7 @@ import json
 import pytest
 import sqlglot
 
-from duckdb_utils import init_benchmark_tables
+from duckdb_utils import init_benchmark_tables_from_parquet
 
 
 def get_queries(benchmark_type):
@@ -52,7 +52,7 @@ def compare_results(presto_rows, duckdb_rows, types, is_sorted_query):
 
 
 def init_duckdb_tables(benchmark_type):
-    init_benchmark_tables(benchmark_type, get_scale_factor(benchmark_type))
+    init_benchmark_tables_from_parquet(benchmark_type, get_scale_factor(benchmark_type), get_abs_file_path(f"data/{benchmark_type}"))
 
 
 def execute_duckdb_query(query):
--- a/presto/testing/integration_tests/tpch_test.py
+++ b/presto/testing/integration_tests/tpch_test.py
@@ -9,8 +9,11 @@ BENCHMARK_TYPE = "tpch"
 @pytest.fixture(scope="module")
 def tpch_queries():
     queries = test_utils.get_queries(BENCHMARK_TYPE)
+    scale_factor = test_utils.get_scale_factor(BENCHMARK_TYPE)
+    value_ratio = 0.0001 / float(scale_factor)
     # Referencing the CTE defined "supplier_no" alias in the parent query causes issues on presto.
     queries["Q15"] = queries["Q15"].replace(" AS supplier_no", "").replace("supplier_no", "l_suppkey")
+    queries["Q11"] = queries["Q11"].replace("0.0001000000", str(value_ratio))
     return queries
 
 


benchmark_data_tools/generate_data_files.py
```diff
@@ -147,21 +147,21 @@ def get_column_projection(column_metadata, convert_decimals_to_floats):
     parser = argparse.ArgumentParser(
         description="Generate benchmark parquet data files for a given scale factor. "
                     "Only the TPC-H and TPC-DS benchmarks are currently supported.")
-    parser.add_argument("--benchmark-type", type=str, required=True, choices=["tpch", "tpcds"],
+    parser.add_argument("-b", "--benchmark-type", type=str, required=True, choices=["tpch", "tpcds"],
```
misiugodfrey, line 150: Added more convenient short-form options as I was running this script directly a lot.

presto/testing/integration_tests/tpch_test.py
```diff
@@ -9,8 +9,11 @@
 @pytest.fixture(scope="module")
 def tpch_queries():
     queries = test_utils.get_queries(BENCHMARK_TYPE)
+    scale_factor = test_utils.get_scale_factor(BENCHMARK_TYPE)
+    value_ratio = 0.0001 / float(scale_factor)
```
misiugodfrey, line 13: This query needed to be updated based on the scale factor.

benchmark_data_tools/duckdb_utils.py
```diff
@@ -13,5 +13,21 @@ def init_benchmark_tables(benchmark_type, scale_factor):
 
     duckdb.sql(f"INSTALL {benchmark_type}; LOAD {benchmark_type}; CALL {function_name}(sf = {scale_factor});")
 
+def init_benchmark_tables_from_parquet(benchmark_type, scale_factor, data_path):
+    tables = duckdb.sql("SHOW TABLES").fetchall()
+    assert len(tables) == 0
+
+    if benchmark_type == "tpch":
```
misiugodfrey, line 20: We could theoretically try to grab the table names from a schema file or something, but I think that might be overkill for a simple task like this?

presto/testing/integration_tests/tpch_test.py
paul-aiyedun: Can you please expand on this?

benchmark_data_tools/duckdb_utils.py
paul-aiyedun: I think we should avoid hard-coding table names whenever possible.

presto/testing/integration_tests/tpch_test.py
misiugodfrey: The "fraction" value in this query is dependent on the scale factor of the data, and should be defined as "0.0001 / scale_factor", but in our queries it is hard-coded as "0.0001".  Likely because duckdb has this same issue: https://github.com/duckdb/duckdb/issues/17965

If we try to run at higher scale factors without adjusting this value we will get incorrect results.

presto/testing/integration_tests/common_fixtures.py
```diff
@@ -25,34 +26,78 @@ def presto_cursor(request):
...
-    schema = schema if schema else f"{benchmark_type}_test"
+    schema = f"{benchmark_type}_test"
+    schema = request.config.getoption("--schema-name") if request.config.getoption("--schema-name") else schema
     conn = prestodb.dbapi.connect(host=hostname, port=port, user=user, catalog="hive", schema=schema)
     return conn.cursor()
 
-
 @pytest.fixture(scope="module")
-def setup_and_teardown(request, presto_cursor):
+def get_scale_factor(request, presto_cursor):
```
misiugodfrey, line 35: Scale factor can be determined dynamically by looking at the data, avoid the need for an option and possible mis-matches between data an option.

presto/testing/integration_tests/common_fixtures.py
```diff
@@ -25,34 +26,78 @@ def presto_cursor(request):
...
+    else:
+        # default assumed location for metadata file.
+        meta_file = test_utils.get_abs_file_path(f"data/{benchmark_type}/metadata.json")
+    if meta_file == "":
+        raise pytest.UsageError("Could not find metadata file in data repository")
+    return test_utils.get_scale_factor_from_file(meta_file)
 
-    has_schema_name = bool(request.config.getoption("--schema-name"))
-    scale_factor = request.config.getoption("--scale-factor")
+def validate_options(request):
```
misiugodfrey, line 61: Added support to make sure the data directory we are looking at exists.

presto/testing/integration_tests/common_fixtures.py
```diff
@@ -25,34 +26,78 @@ def presto_cursor(request):
...
+    should_create_tables = create_schema or not has_schema_name
     if should_create_tables:
-        schema_name = f"{benchmark_type}_test"
         schemas_dir = test_utils.get_abs_file_path(f"schemas/{benchmark_type}")
-        data_sub_directory = f"integration_test/{benchmark_type}"
-        create_hive_tables.create_tables(presto_cursor, schema_name, schemas_dir, data_sub_directory)
+        print(f"data_dir: {data_dir}")
+        create_hive_tables.create_tables(presto_cursor, schema_name, schemas_dir, data_dir)
+
+    # duckdb will need to know the name of each table in a hive schema, as well as the path to the parquet directory they are based on.
```
misiugodfrey, line 91: The main fix here is to make sure duckdb tables are backed by the same files as the ones we are providing for presto tables.  This means they could live somewhere other than the default data directory (if a pre-existing schema was provided), so we need to query the presto tables to see what data they are pointing at and then use those same directories.

presto/testing/integration_tests/schemas/tpch/customer.sql
```diff
@@ -1,4 +1,4 @@
-CREATE TABLE hive.tpch_test.customer (
+CREATE TABLE hive.{schema}.customer (
```
misiugodfrey, line 1: Since we can create new schema now we need to parameterize the schema in the sql files.

scripts/py_env_functions.sh
```diff
@@ -75,12 +75,12 @@ function init_python_virtual_env() {
 }
 
 function delete_python_virtual_env() {
-  if command -v conda &> /dev/null; then
-    echo "Deactivating conda environment"
-    conda deactivate
-  elif command -v deactivate &> /dev/null; then
+  if command -v deactivate &> /dev/null; then
```
misiugodfrey, line 78: I needed to reverse the order here to reflect the order used by `init_python_virtual_env`.  Otherwise, if you have a conda environment setup by default it will set up the venv on init, but teardown the via conda and error out because there was no `conda init` call in the script.

mattgara (commented):
Overall LGTM, just had some cleanup comments.

presto/scripts/run_integ_test.sh
```diff
@@ -32,24 +32,26 @@ OPTIONS:
     -h, --hostname          Hostname of the Presto coordinator.
     -p, --port              Port number of the Presto coordinator.
     -u, --user              User who queries will be executed as.
-    -s, --schema-name       Name of the schema containing the tables that will be queried.
-    -f, --scale-factor      Scale factor of the schema containing the tables that will be queried. This must be
-                            set when (and only when) --schema-name is specified.
+    -s, --schema-name       Name of the schema containing the tables that will be queried (default is {benchmark_type}_test).
+    -c, --create-schema     Create a new schema with --schema-name
```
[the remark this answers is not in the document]
paul-aiyedun: I don't think this should be a concern of the integration test. We have setup scripts that handle schema/table creation outside the default path.

presto/scripts/run_integ_test.sh
```diff
@@ -32,24 +32,26 @@ OPTIONS:
...
+    -d, --data-dir          What directory (contained within presto/testing/integration_tests/data/) to get data from (default is value of {benchmark-type})
```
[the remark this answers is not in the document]
paul-aiyedun: I believe the `PRESTO_DATA_DIR` environment variable is what we use to decide the path for custom schema/table setup. Can we use that here for consistency?

paul-aiyedun (commented):
The approach of testing with DuckDB tables that use the same data files makes sense to me. However, I think we should align this with how the existing schema/table setup scripts work vs creating new files under the `integration_test` path.

presto/docker/config/etc_worker/config_java.properties
```diff
@@ -1,3 +1,8 @@
 coordinator=false
 http-server.http.port=8080
 discovery.uri=http://presto-coordinator:8080
+
```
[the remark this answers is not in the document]
paul-aiyedun: Are these not coordinator specific configurations?

presto/scripts/run_integ_test.sh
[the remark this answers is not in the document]
misiugodfrey: I've removed the option and it will now use PRESTO_DATA_DIR instead.

presto/testing/integration_tests/queries/tpch/queries.json
```diff
@@ -9,7 +9,7 @@
...
-  "Q11": "SELECT ps_partkey, sum(ps_supplycost * ps_availqty) AS value FROM partsupp, supplier, nation WHERE ps_suppkey = s_suppkey AND s_nationkey = n_nationkey AND n_name = 'GERMANY' GROUP BY ps_partkey HAVING sum(ps_supplycost * ps_availqty) > ( SELECT sum(ps_supplycost * ps_availqty) * 0.0001000000 FROM partsupp, supplier, nation WHERE ps_suppkey = s_suppkey AND s_nationkey = n_nationkey AND n_name = 'GERMANY') ORDER BY value DESC",
+  "Q11": "SELECT ps_partkey, sum(ps_supplycost * ps_availqty) AS value FROM partsupp, supplier, nation WHERE ps_suppkey = s_suppkey AND s_nationkey = n_nationkey AND n_name = 'GERMANY' GROUP BY ps_partkey HAVING sum(ps_supplycost * ps_availqty) > ( SELECT sum(ps_supplycost * ps_availqty) * <SF_PLACEHOLDER> FROM partsupp, supplier, nation WHERE ps_suppkey = s_suppkey AND s_nationkey = n_nationkey AND n_name = 'GERMANY') ORDER BY value DESC",
```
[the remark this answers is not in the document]
misiugodfrey: I've changed this to use the syntax `{SF_FRACTION}` to be replaced in python.  Can you directly put the calculation in like you quoted above?  I'm not aware of that sytax.

presto/scripts/run_integ_test.sh
```diff
@@ -32,24 +32,26 @@ OPTIONS:
     -h, --hostname          Hostname of the Presto coordinator.
     -p, --port              Port number of the Presto coordinator.
     -u, --user              User who queries will be executed as.
-    -s, --schema-name       Name of the schema containing the tables that will be queried.
-    -f, --scale-factor      Scale factor of the schema containing the tables that will be queried. This must be
```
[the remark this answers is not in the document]
misiugodfrey: I've spoken with Paul, and I think the best solution here is to auto-detect but make sure there is clear error messaging if the metadata file is missing.  If the user can add their own metadata file if one is missing then this seems like the best solution.

presto/scripts/run_integ_test.sh
```diff
@@ -164,24 +164,14 @@ if [[ -n ${USER} ]]; then
...
-  fi
-  PYTEST_ARGS+=("--schema-name ${SCHEMA_NAME}")
-fi
-
 if [[ -n ${SCALE_FACTOR} ]]; then
-  if [[ -z ${SCHEMA_NAME} ]]; then
-    echo "Error: Scale factor should be set only when --schema-name is specified."
-    print_help
-    exit 1
-  fi
```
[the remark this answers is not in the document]
misiugodfrey: I've change it as you said.  We now check for the metadata file and error out if it's missing with instructions on how to add one.

presto/testing/integration_tests/common_fixtures.py
```diff
@@ -30,30 +32,60 @@ def presto_cursor(request):
...
+        meta_file = test_utils.get_abs_file_path(f"{location}/../metadata.json")
+    else:
+        # default assumed location for metadata file.
+        meta_file = test_utils.get_abs_file_path(f"data/{benchmark_type}/metadata.json")
+    if meta_file == "":
+        raise pytest.UsageError("Could not find metadata file in data repository")
+    return test_utils.get_scale_factor_from_file(meta_file)
 
-    if scale_factor and not has_schema_name:
-        raise pytest.UsageError("--scale-factor should be set only when --schema-name is specified")
```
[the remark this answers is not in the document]
misiugodfrey: Removed.

presto/docker/config/etc_worker/config_java.properties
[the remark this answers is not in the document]
paul-aiyedun: In my testing, this worked with only an update to the coordinator config. Can you please verify?

presto/testing/integration_tests/test_utils.py
```diff
@@ -42,40 +41,39 @@ def execute_query_and_compare_results(presto_cursor, queries, query_id):
...
 
-def compare_results(presto_rows, duckdb_rows, types, is_sorted_query):
+def compare_results(presto_rows, duckdb_rows, types):
     row_count = len(presto_rows)
     assert row_count == len(duckdb_rows)
 
     duckdb_rows = normalize_rows(duckdb_rows, types)
     presto_rows = normalize_rows(presto_rows, types)
 
-    if not is_sorted_query:
```
paul-aiyedun, line 59: This would not correctly compare the results of sorted queries.

paul-aiyedun (changes requested):
Changes overall look good to me. However, I think there is an issue with the result set comparison update.

presto/testing/common/fixtures.py
```diff
@@ -15,11 +15,18 @@
 import pytest
 
 from . import test_utils
-
+from ..integration_tests.common_fixtures import get_scale_factor
```
paul-aiyedun: `common` should not be depending on `integration_tests`. The `get_scale_factor` fixture should probably be moved to `common`.

presto/testing/integration_tests/test_utils.py
```diff
@@ -19,58 +19,87 @@
...
+    # If we have an ORDER BY clause we want to test that the resulting order of those columns
+    # is correct, in addition to overall values being correct.
+    order_indices = get_orderby_indices(query, column_names)
+    if order_indices:
+        # Approx only on ORDER BY columns
+        for col_index in order_indices:
+            if types[col_index].id in FLOATING_POINT_TYPES:
+                for r in range(len(duckdb_rows)):
+                    duckdb_rows[r][col_index] = pytest.approx(duckdb_rows[r][col_index], abs=0.02)
+        # Project both results to ORDER BY columns and compare in original order
```
paul-aiyedun: Consider extracting the columns first and then calling the `approx_floats` function. That should avoid the above for loop.

benchmark_data_tools/generate_data_files.py
```diff
@@ -43,55 +42,61 @@ def generate_partition(table, partition, raw_data_path, scale_factor, num_partit
...
 
-def generate_data_files(benchmark_type, data_dir_path, scale_factor, convert_decimals_to_floats, use_duckdb, num_threads, verbose, max_rows):
-    Path(f"{data_dir_path}").mkdir(parents=True, exist_ok=True)
+def generate_data_files(args):
+    if os.path.exists(args.data_dir_path):
+        shutil.rmtree(args.data_dir_path)
+    Path(f"{args.data_dir_path}").mkdir(parents=True, exist_ok=True)
+
     # tpchgen is much faster, but is exclusive to generating tpch data.  Use duckdb as a fallback.
-    if benchmark_type == "tpch" and not use_duckdb and float(scale_factor) >= 1:
```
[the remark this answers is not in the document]
misiugodfrey: Yes.  We can generate data with scale factors less than 1 so it was only in place before because we had an arbitrary limit on the SFs that were allowed.

misiugodfrey:
> Per this comment: [#74 (comment)](https://github.com/rapidsai/velox-testing/pull/74#discussion_r2430337526), we should be able to run the GPU Q15 test with this PR update?

I think there was a misunderstanding.  Q15 is the only query that this update will not fix, as that does not have to do with the comparison of output value types so much as whether or not there are any rows returned for that query.

presto/testing/integration_tests/test_utils.py
misiugodfrey: If we want to use `approx_floats()` I'll just run it over the unsorted duck_db results.

presto/testing/integration_tests/tpch_test.py
```diff
@@ -16,11 +16,11 @@
...
 from . import test_utils
 from .common_fixtures import presto_cursor, setup_and_teardown
-from ..common.fixtures import tpch_queries
+from ..common.fixtures import tpch_queries, get_scale_factor
 
 BENCHMARK_TYPE = "tpch"
 
 
-@pytest.mark.usefixtures("setup_and_teardown")
+@pytest.mark.usefixtures("setup_and_teardown", "get_scale_factor")
```
paul-aiyedun: I don't think this update is needed, since `tpch_queries` already depends on `get_scale_factor`?

presto/testing/integration_tests/test_utils.py
```diff
@@ -19,60 +19,80 @@
...
+    sorted_duckdb_rows = sorted(duckdb_rows)
+    sorted_presto_rows = sorted(presto_rows)
+    approx_floats(sorted_duckdb_rows, types)
+    assert sorted_presto_rows == sorted_duckdb_rows
+
+    # If we have an ORDER BY clause we want to test that the resulting order of those
+    # columns is correct, in addition to overall values being correct.
+    order_indices = get_orderby_indices(query, column_names)
+    # Only a sorted query should have ORDER BY indicies.
+    assert not (order_indices == get_is_sorted_query(query))
```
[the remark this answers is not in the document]
misiugodfrey: Yes.  Empty Python lists are "falsy", so they will evaluate to false, while a non-empty list evaluates to true.  If this isn't clear I'll change the syntax to something more obvious.

PR #50, opened 2025-09-17
Moved duckdb schema calls outside of multi-threaded portions

misiugodfrey:
The rewrite_parquet portions of the data generation were making concurrent duckdb calls on the same connection, which is not thread-safe.

This patch runs all the duckdb calls sequentially before the multi-threaded portion of the code, storing the results in a dictionary that can be accessed in a thread-safe way.

### DIFF
--- a/benchmark_data_tools/rewrite_parquet.py
+++ b/benchmark_data_tools/rewrite_parquet.py
@@ -6,25 +6,36 @@ import duckdb
 from pathlib import Path
 from concurrent.futures import ThreadPoolExecutor
 
+# Perform a single-threaded pass to obtain the table names and schema.
+# Assumes hive data heirarchy.
+def map_table_schemas(input_dir, verbose):
+    table_to_schema_map = {}
+    for root, subdirs, files in os.walk(input_dir):
+        for table_name in subdirs: # assumes each subdir in root path represents a table.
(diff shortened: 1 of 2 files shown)


benchmark_data_tools/rewrite_parquet.py
```diff
@@ -6,25 +6,36 @@
 from pathlib import Path
 from concurrent.futures import ThreadPoolExecutor
 
+# Perform a single-threaded pass to obtain the table names and schema.
+# Assumes hive data heirarchy.
+def map_table_schemas(input_dir, verbose):
```
paul-aiyedun, line 11: `duckdb_utils.py` is probably a better location for this function. Also, I don't think the directory path dependency is needed (i.e. DuckDB can return the table names).

misiugodfrey:
> Would this have an impact on the runtime of that step? If so, should we instead consider having each thread simply have its own connection to `duckdb` (which I believe should be thread-safe?)

If anything, this should actually make things faster, as previously we were making multiple duckdb calls to get the same schema (since each partition for the same table would make a call to get the schema, despite it being the same as the last partition).  The duckdb call only needs to be made once to get the schema per table, so I don't think there is any benefit to multi-threading that portion of code.  The amount of duckdb calls should be fixed regardless of the SF, so it's not a bottleneck to scaling.

PR #51, opened 2025-09-18
Proposed fix for Presto Native Worker runtime link failures

simoneves:
The latest Presto stack seems to suffer from a problem whereby the final CPU/GPU worker `presto_server` executable fails to find `libboost_iostreams.so`.

This is because the current implementation uses an `ldd` and `awk` trick to determine the set of required runtime DSOs to copy to a separate location, presumably with the intention of creating a minimal deployment.

For some reason, that step now fails to find and copy `libboost_iostreams.so` even though it's in `/usr/local/lib` just like all the others.

This PR removes the copy step completely, as it's not required in this context, since this runtime container is based on the dependencies container and contains all the runtime DSO dependencies in their original locations.

The `libboost_iostreams` DSO is still not found unless `/usr/local/lib` is explicitly added to the search path. This can be done using a custom `ld.conf` file for the executable, or more simply just by setting `LD_LIBRARY_PATH` in the env.

Note that the required Boost libraries appear to be defined by `presto-native-execution/CMakeLists.txt` in the Presto repo AND in the root `CMakeLists.txt` in the Velox repo, but adding `iostreams` to one or both of those does not resolve the original problem.

Discuss!

Note that this PR is chained off #47 which is also required with latest Presto.


presto/docker/native_build.dockerfile
```diff
@@ -13,15 +14,17 @@ ENV NUM_THREADS=${NUM_THREADS}
 
 RUN mkdir /runtime-libraries
 
+# TODO: revert this change once facebook updates the prestissimo image
+RUN dnf install -y -q libnvjitlink-$(echo ${CUDA_VERSION} | tr . -) libnvjitlink-devel-$(echo ${CUDA_VERSION} | tr . -)
```
paul-aiyedun, line 18: Can we please move this to a separate PR with detailed description of the issue being addressed?

simoneves: @paul-aiyedun This bit can be moved into Avi's PR (#47) if you want, but I figured that one was for Velox and this one is for Presto...?

simoneves:
> > For some reason, the ldd call in that step fails to find libboost_iostreams.so even though it's in /usr/local/lib just like all the others, and hence it does not get copied. The failure of the copy step does not kill the worker image build, which completes, but the resulting container exits immediately because the server executable can't find the iostreams DSO, presumably for the same reason that ldd can't find it.
> 
> I believe this is expected. IIRC `/usr/local/lib` is a not a path that resides in `LD_LIBRARY_PATH` in most environments by default. It either has to be added to it, or you need to add a `-L` option to specify where it should find this.

Then why is it only that one that doesn't get found, when it happily finds and copies seven other Boost DSOs that are in the same source location?

mattgara:
> when it happily finds and copies seven other Boost DSOs that are in the same source location?

Could it be that

1) Those other DSOs are explicitly specified with `-L` during linkage?
2) There exist other versions of that library in paths that are included in `LD_LIBRARY_PATH` by default, whereas this library is only found under `/usr/local`?

simoneves:
> 1. ~Those other DSOs are explicitly specified with `-L` during linkage?`~ I guess in this case the directory is the same for all libs so this can't explain it.
> 2. There exist other versions of that library in paths that are included in `LD_LIBRARY_PATH` by default, whereas this library is only found under `/usr/local`?

`ldd` reports the other seven as being in `/usr/local/lib`. It reports `iostreams` as `not found` even though it's also there.

There are Boost library links in the CMake for both Presto and Velox, which do not currently include `iostreams`, so I figured that was why, but I added it in both places and it made no difference. Something else weird is going on.

presto/docker/native_build.dockerfile
```diff
@@ -13,15 +14,17 @@ ENV NUM_THREADS=${NUM_THREADS}
...
+
 RUN --mount=type=bind,source=presto/presto-native-execution,target=/presto_native_staging/presto \
     --mount=type=bind,source=velox,target=/presto_native_staging/presto/velox \
     --mount=type=cache,target=${BUILD_BASE_DIR} \
     make --directory="/presto_native_staging/presto" cmake-and-build BUILD_TYPE=${BUILD_TYPE} BUILD_DIR="" BUILD_BASE_DIR=${BUILD_BASE_DIR} && \
-    ldd ${BUILD_BASE_DIR}/presto_cpp/main/presto_server | awk 'NF == 4 { system("cp " $3 " /runtime-libraries") }' && \
+    ldd ${BUILD_BASE_DIR}/presto_cpp/main/presto_server | grep presto_native | awk 'NF == 4 { print $3 }' | cp -t /runtime-libraries || true && \
```
simoneves, line 24: The `grep` makes it copy ONLY the NV libs (`cudf`, `nvcomp`, `kvikio` etc.) and not all the standard ones, which the executable will find in their original locations.

The `|| true` addition allows for the case of ZERO NV libs, which is the case for a CPU build.

presto/docker/native_build.dockerfile
```diff
@@ -13,15 +14,17 @@ ENV NUM_THREADS=${NUM_THREADS}
...
     cp ${BUILD_BASE_DIR}/presto_cpp/main/presto_server /usr/bin
 
 RUN mkdir /usr/lib64/presto-native-libs && \
-    cp /runtime-libraries/* /usr/lib64/presto-native-libs/ && \
-    echo "/usr/lib64/presto-native-libs" > /etc/ld.so.conf.d/presto_native.conf
+    cp /runtime-libraries/* /usr/lib64/presto-native-libs || true
```
simoneves, line 28: Same as above for the `|| true`.

presto/scripts/build_centos_deps_image_patches.diff
```diff
@@ -0,0 +1,23 @@
...
+                  ../scripts/setup-adapters.sh && \
+                  source ../velox/scripts/setup-centos9.sh && \
+                  source ../velox/scripts/setup-centos-adapters.sh && \
+-                 install_adapters && \
++                 # equivalent of install_adapters() but without problematic Hadoop SDK download
++                 install_adapters_deps_from_dnf && \
++                 install_s3 && \
++                 install_gcs-sdk-cpp && \
++                 install_azure-storage-sdk-cpp && \
++                 # continue as before
```
simoneves, line 15: This omits the Hadoop SDK install, which is a very unreliable download, pending a Velox change to a more reliable mirror

presto/scripts/build_centos_deps_image_patches.diff
```diff
@@ -0,0 +1,23 @@
...
+                  install_clang15 && \
+-                 install_cuda 12.8) && \
++                 install_cuda 12.8 && \
++                 # also install matching libnvjitlink
++                 dnf install -y -q libnvjitlink-12-8 libnvjitlink-devel-12-8) && \
```
simoneves, line 20: This installs a matching `nvjitlink` in the container alongside CUDA, rather than installing it later prior to USING the container to build the worker.

paul-aiyedun (approved):
Looks good to me.

PR #54, opened 2025-09-23
Velox: add debug build option

mattgara:
* Add `RELWITHDEBINFO` and `DEBUG` builds

* Update downstream tasks to support these new builds


velox/scripts/build_velox.sh
```diff
@@ -122,6 +126,23 @@ parse_args() {
           exit 1
         fi
         ;;
+      --build-type)
+        if [[ -n "${2:-}" && ! "${2}" =~ ^- ]]; then
+          case "${2}" in
+            "Release"|"Debug"|"RelWithDebInfo")
```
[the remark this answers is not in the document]
misiugodfrey: Paul's recent PR taught me that lower case can be achieve via `@L` in more recent versions of bash.  You can use:
`BUILD_TYPE=${2@L}`
Which will take the second argument and convert it to lower case before assignment.

misiugodfrey (approved):
LGTM, although I agree Paul's comments are worth looking into before the merging.

velox/scripts/utils.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Shared utility functions for Velox build and test scripts
+
+# Helper function to get BUILD_TYPE from container environment
+get_build_type_from_container() {
```
paul-aiyedun, line 6: Nit: Move this to the test_velox.sh file, since it is only used there?

velox/scripts/test_velox.sh
```diff
@@ -16,7 +16,15 @@
...
-source ./config.sh
+source "config.sh"
+source "utils.sh"
+
+# Get BUILD_TYPE from container environment
+BUILD_TYPE=$(get_build_type_from_container "$COMPOSE_FILE" "$CONTAINER_NAME")
+
+# expected output directory
+EXPECTED_OUTPUT_DIR="/opt/velox-build/${BUILD_TYPE}"
+EXPECTED_OUTPUT_LIB_DIR="${EXPECTED_OUTPUT_DIR}/lib"
```
paul-aiyedun, line 27: Is this variable used?

paul-aiyedun (approved):
I had some code cleanup comments, but changes look good to me.

velox/scripts/utils.sh
mattgara: Removed file and moved to test_velox.sh.

PR #59, opened 2025-09-24
Patches for successful Presto deps container build (as of 9/24)

simoneves:
Without these patches, a Presto dependencies container will not currently build.

* Bump Hadoop from 3.3.0 to 3.3.6 allowing download from main Apache server instead of very slow archive
* Install `nvJitLink` package along with CUDA, required for CUDF 25.10

Note that the patches are to the Velox in the sub-module of the Presto clone, which must always be derived from Velox mainline (at a pinned version) not ours, so must be patched live until these changes are in mainline Velox.

Two other fixes are also required, but they are in the Presto scripts, and not in this PR (at present)

* Arrow re-build removal
* GPerf download mirror change

Those are currently included in the "bleeding edge" Presto from @devavret's fork (@karthikeyann please confirm)


presto/scripts/build_centos_deps_image.sh
```diff
@@ -2,7 +2,10 @@
 
 set -e
 
-PATCH_FILE_PATH=$(readlink -f copy_arrow_patch.patch)
+# patch Velox submodule for deps container build success
```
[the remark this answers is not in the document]
simoneves: The Velox sub-module that needs patching is at `presto/velox/...`

presto/scripts/build_centos_deps_image.sh
```diff
@@ -2,7 +2,10 @@
...
+pushd ../../../presto
+git apply ../velox-testing/presto/scripts/patch_hadoop_and_nvjitlink_092225.diff || true
```
misiugodfrey, line 7: Presumably the `|| true` is here so that we will continue unobstructed if the upstream changes land and this patch no longer applies.  Would we rather that over a failure if the patch does not fit?  The lesser concern is that we forget to ever remove this once the upstream fixes land, but is there any concern - since this diff is two patches - that one lands and the other does not?

misiugodfrey (approved):

presto/scripts/build_centos_deps_image.sh
simoneves: @misiugodfrey the reason was actually so that if you ran it again on an already patched clone it would just no-op, but this isn't a very tidy way of achieving that. We should probably at least check the result of the attempted patch, or check the code BEFORE attempting the patch (although not sure how we'd do the latter).

Certainly it would be better if it threw an actual error if the patch actually failed (for a reason other than it was already done).

Suggestions welcomed, although I am now painfully aware that I have created yet another conflict here with the other development efforts to have something to apply patches. I really just wish we didn't have do.

Unfortunately, even a formal Presto fork wouldn't help here, due to the sub-module thing.

Maybe we should propose an option for Presto to build the deps from a sibling Velox instead of a sub-module, and then it would use one that's completely under our control, and any such patches could just be combo PRs in our fork.

simoneves: We still really need a `rapidsai/prestodb` in order to have control over the Presto side too. My understanding is that we have tried to avoid this, but I'm not sure that's gonna remain possible. We're stuck on the wrong side of that line already now that the best-working Presto is only in Devavret's fork.

presto/patches/patch_arrow_092525.diff
```diff
@@ -0,0 +1,13 @@
+diff --git a/presto-native-execution/scripts/setup-adapters.sh b/presto-native-execution/scripts/setup-adapters.sh
+index 42596d1c05..108ecbbc63 100755
+--- a/presto-native-execution/scripts/setup-adapters.sh
++++ b/presto-native-execution/scripts/setup-adapters.sh
+@@ -48,7 +48,7 @@ function install_arrow_flight {
+   # NOTE: benchmarks are on due to a compilation error with v15.0.0, once updated that can be removed
+   # see https://github.com/apache/arrow/issues/41617
+   EXTRA_ARROW_OPTIONS=" -DARROW_FLIGHT=ON -DARROW_BUILD_BENCHMARKS=ON -DgRPC_SOURCE=BUNDLED -DProtobuf_SOURCE=BUNDLED "
+-  install_arrow
++  #install_arrow
```
simoneves, line 10: The Arrow patch (to Presto)

presto/scripts/build_centos_deps_image.sh
```diff
@@ -2,10 +2,56 @@
...
+DEPS_IMAGE="presto/prestissimo-dependency:centos9"
+
+if [ "${REBUILD_DEPS}" == "1" ]; then
+	echo "Forcing rebuild of Presto dependencies/run-time container image"
+	echo "This does not delete any existing worker container images which must be manually deleted"
+	docker rmi -f ${DEPS_IMAGE} 
+	docker builder prune -f
+elif [ ! -z $(docker images -q ${DEPS_IMAGE}) ]; then
+	echo "Found existing Presto dependencies/run-time container image"
+	exit 0
```
simoneves, line 18: This skips the rest of the script if a valid container already exists, mostly to avoid a likely failed re-patch attempt if you've already run this once.

presto/scripts/start_presto_helper.sh
```diff
@@ -76,7 +76,7 @@ fi
 DOCKER_COMPOSE_FILE_PATH=../docker/docker-compose.$DOCKER_COMPOSE_FILE.yml
 if (( ${#BUILD_TARGET_ARG[@]} )); then
   if [[ ${BUILD_TARGET_ARG[@]} =~ ($CPU_WORKER_SERVICE|$GPU_WORKER_SERVICE) ]]; then
-    ./build_centos_deps_image.sh
+    REBUILD_DEPS=${REBUILD_DEPS} ./build_centos_deps_image.sh
```
simoneves, line 79: Pass the required env into the child script

presto/scripts/build_centos_deps_image.sh
```diff
@@ -2,10 +2,56 @@
...
+	git checkout .
+fi
+git apply ../velox-testing/presto/patches/patch_arrow_092525.diff
+popd
+# in Velox sub-module, change the Hadoop version and mirror, and add libnvjitlink install
+pushd ../../../presto/presto-native-execution/velox
+if [ ${REBUILD_DEPS} == "1" ]; then
+	git checkout .
+fi
+git apply ../../../velox-testing/presto/patches/patch_hadoop_and_nvjitlink_092225.diff
```
simoneves, line 42: Ditto and ditto for the Velox sub-module patches

presto/patches/patch_hadoop_and_nvjitlink_092225.diff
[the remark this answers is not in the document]
simoneves: That won't help in the short term. These patches are for problems with the Velox sub-module in Presto. Unless we can find a way to drop OUR Velox in there for the purposes of a Deps Container build, we're stuck with patching.

I had a poke at swapping out the sub-module the other day, but couldn't get it to behave, but we should pursue that otherwise we'll keep getting blocked like this.

presto/scripts/start_presto_helper.sh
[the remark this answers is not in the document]
simoneves: Maybe, but we still need to handle the third case of not having a container already and automatically building it.

presto/patches/patch_hadoop_and_nvjitlink_092225.diff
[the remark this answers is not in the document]
simoneves: I tried a simple delete-and-replace-with-symlink but that failed because the symlink ended obviously pointed to something outside the implicit container file mapping. We could probably persuade the sub-module to be to our fork and stable branch instead. I will look into this again tomorrow.

simoneves:
@paul-aiyedun I don't know what you mean by "moving to `docker compose build`". Please elaborate.

paul-aiyedun:
> @paul-aiyedun I don't know what you mean by "moving to `docker compose build`". Please elaborate.

This is in response to
>I can't find a way to pass the --no-cache option to the docker compose which does the actual build, as the args that adds are only valid for a docker build not a docker compose. Discuss.

We can use `docker compose build` in the script instead of `docker compose up/down`.

presto/patches/patch_hadoop_and_nvjitlink_092225.diff
```diff
@@ -0,0 +1,38 @@
...
+diff --git a/scripts/setup-common.sh b/scripts/setup-common.sh
+index 594616649..15049326f 100755
+--- a/scripts/setup-common.sh
++++ b/scripts/setup-common.sh
+@@ -393,7 +393,7 @@ function install_azure-storage-sdk-cpp {
+ 
+ function install_hdfs_deps {
+   # Dependencies for Hadoop testing
+-  wget_and_untar https://archive.apache.org/dist/hadoop/common/hadoop-"${HADOOP_VERSION}"/hadoop-"${HADOOP_VERSION}".tar.gz hadoop
++  wget_and_untar https://dlcdn.apache.org/dist/hadoop/common/hadoop-"${HADOOP_VERSION}"/hadoop-"${HADOOP_VERSION}".tar.gz hadoop
```
simoneves, line 22: THIS LINE IS WRONG. CORRECTED IN #70

PR #60, opened 2025-09-24
Created common script argument parsing

misiugodfrey:
A lot of our scripts use almost identical format for parsing.  Notably a case statement with the form:
```
--<option>)
        if [[ -n "${2:-}" ]]; then
          <option>="$2"
          shift 2
        else
          echo "ERROR: --<option> requires true or false argument" >&2
          exit 1
        fi
        ;;
```

I decided to take a quick shot at outlining and refactoring this code to make adding argument parsing more generic.

The process now is that each file will declare it's own arrays of options and flags (options being flags that take additional parameters and flags needing none):
```
declare -A OPTION_MAP=( ["-b"]="--build-target" ["-j"]="--num-threads" )
```
The common parse_args() functions (or custom ones if custom handling is required) will use these maps to parse all options and assign the values to environment variables with the same name.

I'm curious what folks think about this approach.

### DIFF
--- a/scripts/helper_functions.sh
+++ b/scripts/helper_functions.sh
@@ -16,6 +16,21 @@
 
 shopt -s extglob # needed for paramaterized case statement
 
+# --- Print error messages in red ---
+function echo_error() {
+    echo -e "\033[0;31m$*\033[0m" >&2
+}
+
+# --- Print warning messages in yellow ---
+function echo_warning() {
+    echo -e "\033[0;33m$*\033[0m: " >&2
+}
+
+# --- Print warning messages in green ---
+function echo_success() {
+    echo -e "\033[0;32m$*\033[0m: "
+}
+
 function make_options() {
     local -n LOCAL_OPTION_MAP="$1"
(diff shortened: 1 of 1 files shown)


misiugodfrey:
> I think there may be an ideal time to do this where there are not many in-flight PRs with scripts that would directly conflict with this change. Right now, this will probably conflict with a lot of in-flight PRs.

I'm fine with holding off on this until more things have merged, as it's not a semantic change.  That being said, now's a good time for conceptual feedback so I can apply any requested changes before I inevitably need to resolve conflicts with all the other landing PRs.

simoneves:
@misiugodfrey updated thoughts on this one? Are we any closer to a point where it could go in, or have we proliferated individual script implementations even more. I think we might have... :(

simoneves:
Another good utility function to add... something to validate that a script is being run from the correct location (most likely its own directory)... similar to https://github.com/rapidsai/velox-testing/blob/main/presto/scripts/generate_presto_config.sh#L32

PR #64, opened 2025-09-27
build(docker): velox: run CentOS adapters setup during adapters image build

Avinash-Raj:


velox/docker/adapters_build.dockerfile
```diff
@@ -47,6 +47,10 @@ ${BUILD_BASE_DIR}/${BUILD_TYPE}/_deps/nvcomp_proprietary_binary-src/lib64"
 
 WORKDIR /workspace/velox
 
+# Copy and run CentOS adapters setup script
+COPY velox/scripts/ /scripts
```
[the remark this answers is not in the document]
Avinash-Raj: Main `velox/scripts/setup-centos-adapters.sh` script depends on many other scripts which resides on the same directory. So I copied the entire directory.

Avinash-Raj:
I don't think there is a separate PR for curl issue on velox fork.

simoneves (approved):
I'm fine with this if there's no neater way to do it.

I would like to see more discussion on the Curl problem. If it's real, then there needs to be a PR to fix it in our Velox fork. I'm just surprised that this hasn't come up for somebody else in some other context.

velox/docker/adapters_build.dockerfile
```diff
@@ -47,6 +47,43 @@ ${BUILD_BASE_DIR}/${BUILD_TYPE}/_deps/nvcomp_proprietary_binary-src/lib64"
 
 WORKDIR /workspace/velox
 
+# Copy and run CentOS adapters setup script
```
paul-aiyedun, line 50: I don't think this is the right place for dependency updates. These updates should likely be in https://github.com/facebookincubator/velox/blob/6998a1461ded77060b817ff3179d3f806fd0b625/scripts/setup-centos-adapters.sh.

paul-aiyedun (changes requested):

mattgara:
> I would like to see more discussion on the Curl problem. If it's real, then there needs to be a PR to fix it in our Velox fork. I'm just surprised that this hasn't come up for somebody else in some other context.

I think the curl fix probably needs a deeper investigation. The "fix" presented here is a fix I used in my work to unblock testing the cudf 25.10 bump. 

However, the core issue is with the `cmake` system I believe. The reason this issue seems to happen is because:

* `kvikio` needs a recent enough curl with certain symbols
* a recent enough version is built with `cpr`, which is another dependency in the build
* `kvikio` detects this version and finds it satisfactory
* come link time, `kvikio` lets the system `curl` take precedence, the system curl is too old (doesn't have the required symbols) 
* an error ensues, even though the correct version of `curl` (the one bundled with `cpr`) is in the same link command, it tries to link the system curl first and fails

The fix here would be to figure out why `kvikio` uses both the system curl and the bundled curl, and have it use the bundled curl only.

scripts/fetch_docker_image_from_s3.sh
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env bash
```
mattgara: Why is this script readded in this PR? Is this a rebase/merge issue?

presto/testing/integration_tests/analyze_tables.py
```diff
@@ -0,0 +1,89 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.
```
mattgara: Similarly why is this script also readded in this PR?

mattgara (commented):
This PR seems to be lumping in changes unrelated to it (perhaps a bad rebase?) Or perhaps there is a merge with `main` that I've missed?

velox/docker/adapters_build.dockerfile
```diff
@@ -16,6 +16,7 @@ ARG BUILD_TYPE=release
 ENV VELOX_DEPENDENCY_SOURCE=SYSTEM \
     GTest_SOURCE=BUNDLED \
     cudf_SOURCE=BUNDLED \
+    curl_SOURCE=BUNDLED \
```
mattgara: Is this needed for the `curl` fix? Does the build work without this?

scripts/fetch_docker_image_from_s3.sh
Avinash-Raj: merge issue

velox/docker/adapters_build.dockerfile
karthikeyann: This is not required anymore. I added the curl fix to `merged-prs`, and the upstream PR https://github.com/facebookincubator/velox/pull/15183 is open

velox/docker/adapters_deps.dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+# Copy and run CentOS adapters setup script
+COPY velox/scripts/ /scripts
+
+RUN set -euxo pipefail && \
+    if ! dnf list installed cuda-nvcc-$(echo ${CUDA_VERSION} | tr '.' '-') 1>/dev/null || \
+       ! dnf list installed libnvjitlink-devel-$(echo ${CUDA_VERSION} | tr '.' '-') 1>/dev/null; then \
+       bash -c "source /scripts/setup-centos-adapters.sh &&  install_cuda ${CUDA_VERSION}"; \
+    fi && \
+    pip install cmake==3.30.4
```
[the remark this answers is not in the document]
karthikeyann: This is also updated in upstream PR.
This fix is required only until velox docker images are ready, after the upstream PR https://github.com/facebookincubator/velox/pull/15183 is merged.

velox/docker/adapters_build.dockerfile
Avinash-Raj: Created a separate dependency image.

Avinash-Raj:
Merging this PR with respect to the latest [CI job run](https://github.com/rapidsai/velox-testing/actions/runs/18706856398) passes the build step for both Velox CPU and GPU builds.

PR #70, opened 2025-10-03
WIP Experiments building Presto Deps with Velox fork

simoneves:


presto/patches/patch_hadoop_and_nvjitlink_092225.diff
```diff
@@ -1,15 +1,3 @@
...
---- a/scripts/setup-centos-adapters.sh
-+++ b/scripts/setup-centos-adapters.sh
-@@ -58,6 +58,7 @@ function install_cuda {
-     cuda-minimal-build-"$dashed" \
-     cuda-nvrtc-devel-"$dashed" \
-     libcufile-devel-"$dashed" \
-+    libnvjitlink-devel-"$dashed" \
-     numactl-libs
- }
-
```
simoneves, line 12: This is already present in our fork, so this part of the patch file is no longer required.

presto/patches/patch_hadoop_100225.diff
```diff
@@ -1,15 +1,3 @@
...
-+++ b/scripts/setup-centos-adapters.sh
-@@ -58,6 +58,7 @@ function install_cuda {
-     cuda-minimal-build-"$dashed" \
-     cuda-nvrtc-devel-"$dashed" \
-     libcufile-devel-"$dashed" \
-+    libnvjitlink-devel-"$dashed" \
-     numactl-libs
- }
- 
 diff --git a/scripts/setup-common.sh b/scripts/setup-common.sh
```
simoneves, line 1: This has already been added in our fork, so this part of the patch is no longer required.

presto/scripts/build_centos_deps_image.sh
```diff
@@ -8,48 +8,87 @@ set -e
 
 DEPS_IMAGE="presto/prestissimo-dependency:centos9"
 
-if [ "${REBUILD_DEPS}" == "1" ]; then
-	echo "Forcing rebuild of Presto dependencies/run-time container image"
+if [ "${REBUILD_DEPS}" == "true" ]; then
```
simoneves, line 11: Tweak per @paul-aiyedun

presto/scripts/build_centos_deps_image.sh
```diff
@@ -8,48 +8,87 @@ set -e
...
 fi
 
 #
 # apply current patches for Presto deps container build success
-# as of 09/25/25
+# as of 10/2/25
 #
 
-echo "Applying required local patches to Presto repo and contained Velox sub-module (as of 9/25/25)"
+echo "Modifying Presto clone and contained Velox sub-module for deps container build (as of 10/2/25)"
```
simoneves, line 28: The overall premise of the next change block is to override the Velox sub-module in the Presto clone to be from our Velox fork instead. Then we add the remaining still necessary patches.

If those patches can be included in our forks (Presto and Velox) even if just temporarily, then they do not need to be applied here.

presto/scripts/start_presto_helper.sh
```diff
@@ -76,7 +76,7 @@ fi
 DOCKER_COMPOSE_FILE_PATH=../docker/docker-compose.$DOCKER_COMPOSE_FILE.yml
 if (( ${#BUILD_TARGET_ARG[@]} )); then
   if [[ ${BUILD_TARGET_ARG[@]} =~ ($CPU_WORKER_SERVICE|$GPU_WORKER_SERVICE) ]]; then
-    ./build_centos_deps_image.sh
+    REBUILD_DEPS=${REBUILD_DEPS} ./build_centos_deps_image.sh
```
simoneves, line 79: Surprised that I needed to add this, for the value to get into the child script

presto/scripts/build_centos_deps_image.sh
simoneves: Date needs updating...

presto/scripts/build_centos_deps_image.sh
```diff
@@ -2,10 +2,94 @@
...
 popd
+
+#
+# build deps container image
+#
+
+echo "Building Presto dependencies/run-time image..."
+
+pushd ../../../presto/presto-native-execution
+docker compose --progress plain build centos-native-dependency
```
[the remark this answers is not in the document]
simoneves: Also note the removal of the call to `make submodules` here, as if that needed doing, then the above sync/update has already done what is necessary, and doing it again here messes up the sub-module again.

PR #80, opened 2025-10-15
New Presto Deps Container Build Script

simoneves:
This update was moved from #74 per @paul-aiyedun 

UPDATE 10/17

New much simpler version after discussion. Turns out that the Presto deps container build only needs the `scripts` and `CMake` sub-directories of the Velox sub-module. So instead of messing around with switching the sub-module, we just move it aside and create a temporary replacement with just those two sub-directories copied from the sibling Velox clone, whatever that may be. After the build is done, we put the original sub-module back.

Hopefully the deps container will will CONTINUE to ONLY require those parts, but we can deal with that it if changes.

Obviously this relies on the sibling Velox clone being valid. Note that for some future automated process for building the deps image, the sibling Velox clone will still be required in order for this script to work. We can maybe even add an option to NOT do the switch, so that the default sub-module is used, once that becomes valid. That may actually be quite likely as the current incompatibility is mostly due to the CUDA lib requirement change in cuDF 25.10, and that may not change again for a while, so once that is upstreamed, it may well be that the default sub-module (and indeed some as-yet-unpublished default deps image) may be fine to use.

### DIFF
--- a/presto/scripts/build_centos_deps_image.sh
+++ b/presto/scripts/build_centos_deps_image.sh
@@ -56,18 +56,25 @@ parse_args() {
 parse_args "$@"
 
 # verify sibling Presto and Velox clones
-if [[ ! -d "../../../presto" || ! -d "../../../velox" ]]; then
+if [[ ! -d "../../../presto/presto-native-execution" || ! -d "../../../velox" ]]; then
   echo "Error: Sibling Presto and/or Velox clone not found"
   exit 1
 fi
-
-exit 0
 
 # remove any existing image?
 if [[ ! -z $(docker images -q ${IMAGE_NAME}) ]]; then
 	echo "Removing existing Presto dependencies/run-time image..."
 	docker rmi -f ${IMAGE_NAME}
 fi
+
+# restore original Presto Velox on exit
+# on a clean exit, this happens before the automatic popd
+function cleanup {
+  echo "Restoring original Presto Velox..."
+  rm -rf velox
+  mv velox.bak velox
+}
+trap cleanup EXIT
 
 # move to Presto Velox
 pushd ../../../presto/presto-native-execution > /dev/null
@@ -83,13 +90,5 @@ cp -r ../../velox/CMake velox
 echo "Building..."
 docker compose --progress plain build ${NO_CACHE_ARG} centos-native-dependency
 
-# restore Presto Velox
-echo "Restoring original Presto Velox..."
-rm -rf velox
-mv velox.bak velox
-
-# return
-popd > /dev/null
-
-# done
+# done (will cleanup on exit)
 echo "Presto dependencies/run-time container image built!"


paul-aiyedun (changes requested):

presto/scripts/build_centos_deps_image.sh
```diff
@@ -2,10 +2,147 @@
 
 set -e
 
-PATCH_FILE_PATH=$(readlink -f copy_arrow_patch.patch)
+IMAGE_NAME='presto/prestissimo-dependency:centos9'
+VELOX_REPO='rapidsai/velox'
+VELOX_COMMIT='merged-prs'
```
paul-aiyedun: Can we consistently use the checked out Presto/Velox repos instead of hard-coding these repos/pointing to another repo?

simoneves:
I made it use `git submodule` commands to modify the sub-module instead of rewriting the file manually.

I also removed the bits which reset the two clones, as that was really only necessary if patches were going to be applied. We can put that back again if we have to enhance this script to be able to apply general patches, if they are needed again in the future (hopefully not).

presto/scripts/build_centos_deps_image.sh
```diff
@@ -2,10 +2,86 @@
...
+
+WARNING: If an image of the given name already exists, it will be removed prior to the build.
+
+OPTIONS:
+    -h, --help           Show this help message
+    -i, --image-name     Desired Docker Image name (default: presto/prestissimo-dependency:centos9)
+    -n, --no-cache       Do not use Docker build cache (default: use cache)
+
+EOF
+}
```
paul-aiyedun: Are these options needed? Is there a use case for using different image names (would this break the S3 image download workflow)?

presto/scripts/build_centos_deps_image.sh
simoneves: Moot. See below.

simoneves:
Updated after discussion. See full description up top.

presto/scripts/build_centos_deps_image.sh
```diff
@@ -2,10 +2,86 @@
...
+cp -r ../../velox/CMake velox
+
+# now build
+echo "Building..."
+docker compose --progress plain build ${NO_CACHE_ARG} centos-native-dependency
+
+# restore Presto Velox
+echo "Restoring original Presto Velox..."
+rm -rf velox
+mv velox.bak velox
```
paul-aiyedun: Can we do this in a `trap`, so the restoration occurs even when there is an error?

presto/scripts/build_centos_deps_image.sh
simoneves: Just planning ahead for automated builds. We might want to build with a date suffix or something.

PR #86, opened 2025-10-16
Add option to override Presto GPU build to all CUDA architectures

simoneves:
Previous PR #84 defaulted the Presto GPU build to only the native CUDA architecture, optimizing for CI.

This PR reinstates the ability to build for ALL supported CUDA architectures (70 to 120 inclusive), when this is desirable for builds that may be distributed to run on arbitrary hosts.

### DIFF
--- a/presto/scripts/start_presto_helper_parse_args.sh
+++ b/presto/scripts/start_presto_helper_parse_args.sh
@@ -34,7 +34,7 @@ OPTIONS:
     --build-type         Build type for native CPU and GPU image builds. Possible values are "release",
                          "relwithdebinfo", or "debug". Values are case insensitive. The default value
                          is "release".
-    --all-cuda-archs     Build for all supported CUDA architectures (default: false).
+    --all-cuda-archs     Build for all supported CUDA architectures (GPU only) (default: false).
 
 EXAMPLES:
     $SCRIPT_NAME --no-cache


presto/scripts/start_presto_helper.sh
```diff
@@ -79,15 +79,24 @@ fi
 
 # must determine CUDA_ARCHITECTURES here as nvidia-smi is not available in the docker build context
 CUDA_ARCHITECTURES=""
-if [[ "$VARIANT_TYPE" == "gpu" ]]; then
+if [[ "$ALL_CUDA_ARCHS" == "true" ]]; then
+  if [[ "$VARIANT_TYPE" == "gpu" ]]; then
```
mattgara: Can we please clean up the nested `if`s here? 

For example, the `gpu` `if` is checked twice (here and below), we should probably have it as the external `if` here.

simoneves: Not sure how that makes it any simpler, as I still want to reject the use of that option in a non-GPU build.

simoneves: It could be flattened a bit to this...
```
CUDA_ARCHITECTURES=""
if [[ "$VARIANT_TYPE" == "gpu" && "$ALL_CUDA_ARCHS" == "true" ]]; then
  # build for all supported CUDA architectures
  CUDA_ARCHITECTURES="70;75;80;86;89;90;100;120"
  echo "Building GPU with all supported CUDA architectures"
elif [[ "$VARIANT_TYPE" == "gpu" ]]; then
  # check that nvidia-smi is available
  if ! command -v nvidia-smi &> /dev/null; then
    echo "ERROR: nvidia-smi could not be found. Please ensure that the NVIDIA drivers and Docker runtime are properly installed."
    exit 1
  fi
  # build for the native compute capability of the first GPU (assuming all GPUs are the same)
  CUDA_ARCHITECTURES="$(nvidia-smi --query-gpu=compute_cap --format=csv,noheader | head -n 1 | sed 's/\.//g')"
  echo "Building GPU with CUDA_ARCHITECTURES=$CUDA_ARCHITECTURES"
elif [[ "$ALL_CUDA_ARCHS" == "true" ]]; then
  # invalid options combination
  echo "ERROR: --all-cuda-archs specified but VARIANT_TYPE is not 'gpu'."
  exit 1
fi
```
...?

simoneves: I guess the last one is the most preferable, due to having no nesting. I'll go with that.

mattgara (approved):
LGTM, thanks!

PR #87, opened 2025-10-17
Commented properties files

misiugodfrey:
Added comments for presto properties in the templates.


simoneves (commented):
LGTM but I have cherry-picked it into #48 so you can close.

Thanks for doing it! :)

PR #89, opened 2025-10-21
Add libnvjitlink dependency for Velox builds with cuDF >= 25.10

mattgara:
This is a simple PR to unblock people who hit this issue when compiling with cuDF >= 25.10.


mattgara:
This should be handled in #64

PR #90, opened 2025-10-22
Simplify Velox build changes

simoneves:
The changes that landed in #64 are more complex than was required. This PR simplifies them, and brings the mechanism more in line with the one for Presto that landed last week.

Instead of creating an intermediate Docker image, we simply leverage the existing mechanism that is used to generate the upstream base image (ghcr.io/facebookincubator/velox-dev:adapters) to build that image locally using the modified Velox build scripts which now include nvjitlink package install.

This mirrors what we expect Facebook to do themselves once our cuDF 25.10 PR lands in upstream. Once that is done, we will then have the option to revert to using the upstream image.

The only other change required is to remove the --pull option from the Velox Docker build, as that would otherwise FORCE the build to pull the upstream version of that image instead of using our modified local version.

The build and fetch scripts are adapted accordingly, and we will generate new x86 and ARM versions of the modified image to put on S3 for the fetch script to use.

If possible this PR should be reviewed in combination with the #64 changes which already landed, as this one reverts part of that one, although that might be tricky as I just landed #48 in between.

Updated again to resolve a conflict as #52 was also landed in between!

### DIFF
--- a/velox/scripts/build_velox.sh
+++ b/velox/scripts/build_velox.sh
@@ -253,11 +253,13 @@ DOCKER_BUILD_OPTS+=(--build-arg VELOX_ENABLE_BENCHMARKS="${VELOX_ENABLE_BENCHMAR
 DOCKER_BUILD_OPTS+=(--build-arg TREAT_WARNINGS_AS_ERRORS="${TREAT_WARNINGS_AS_ERRORS}")
 DOCKER_BUILD_OPTS+=(--build-arg BUILD_TYPE="${BUILD_TYPE}")
 
+# Create sccache auth directory unconditionally, it is expected by dockerfile
+mkdir -p ../docker/sccache/sccache_auth/
+
 # Add sccache build arguments
 if [[ "$ENABLE_SCCACHE" == true ]]; then
   DOCKER_BUILD_OPTS+=(--build-arg ENABLE_SCCACHE="ON")
   # Copy auth files to build context
-  mkdir -p ../docker/sccache/sccache_auth/
   cp "$SCCACHE_AUTH_DIR/github_token" ../docker/sccache/sccache_auth/
   cp "$SCCACHE_AUTH_DIR/aws_credentials" ../docker/sccache/sccache_auth/
   


velox/scripts/build_centos_deps_image.sh
```diff
@@ -1,8 +1,11 @@
...
 set -e
-container_name="velox-adapters-deps"
-compose_file="../docker/docker-compose.adapters.build.yml"
-docker compose -f "${compose_file}" up "${container_name}"
-docker compose -f "${compose_file}" down "${container_name}"
+
+echo "Building Velox dependencies/run-time container image..."
+
+pushd ../../../velox
+docker compose -f docker-compose.yml --progress plain build adapters-cpp
```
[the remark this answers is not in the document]
simoneves: We also just use `docker compose build` instead of `up`/`down`.

velox/scripts/fetch_centos_deps_image.sh
```diff
@@ -5,7 +5,7 @@ set -e
 source ./config.sh
 source ../../scripts/fetch_docker_image_from_s3.sh
 
-IMAGE_NAME="velox-adapters-deps:centos9"
+IMAGE_NAME="ghcr.io/facebookincubator/velox-dev:adapters"
```
simoneves: We just change the image name to that of the original base image instead of the intermediate image added by the now-reverted Dockerfile changes.

Avinash-Raj: Make sure that this image is available on S3 on the correct path, wondering what would be the image path of s3.

velox/docker/adapters_build.dockerfile
```diff
@@ -16,7 +16,6 @@ ARG BUILD_TYPE=release
 ENV VELOX_DEPENDENCY_SOURCE=SYSTEM \
     GTest_SOURCE=BUNDLED \
     cudf_SOURCE=BUNDLED \
-    simdjson_SOURCE=BUNDLED \
```
mattgara: I'm wondering if this was a necessary change? Does a recent build still work without this?

simoneves: It worked fine for me without it, both on x86 and ARM.

mattgara (approved):
LGTM

velox/docker/adapters_build.dockerfile
karthikeyann: It worked fine for me without it

PR #91, opened 2025-10-22
Presto: Add Upstream Nightly CI

simoneves:
(currently chained off #105)

This PR adds a second nightly job which builds Presto with their default pinned upstream Velox instead of ours.

This is achieved by having the job fetch the pinned version SHA from the Presto sub-module and check out that same SHA as the sibling Velox clone.

This allows the use of our normal build recipe which expects Velox to be in the sibling clone not the sub-module.

For the purposes of true upstream validation, we arguably need to use the default Presto build recipe for this CI but that will require further changes. Similarly, this CI still uses our Presto deps image (currently just additions to support newer cuDF) but again technically not "vanilla" Presto.

Discuss.

This PR also removes the `Velox Testing Commit` option and associated plumbing from the Presto workflow stack. This was intended to allow devs to run a workflow from a branch and specify that same branch for the checkout of `velox-testing` so that any script additions or changed matched the workflow being tested. Turns out there's a pre-defined GHA variable which contains the branch name that the workflow is being run from, so we can use that directly for the `velox-testing` checkout so that they match automatically.

*NOTE*
THIS PR GOT CLOSED AUTOMATICALLY, PRESUMABLY TO ME CLOSING AND DELETING THE #105 BRANCH/PR
Please still review the changes in this PR (the Presto ones... ignore the Velox one which is unrelated)


presto/scripts/checkout_velox_to_match_presto.sh
```diff
@@ -0,0 +1,13 @@
...
+set -e
+
+# get SHA of Presto Velox submodule
+pushd ../../presto/presto-native-execution/velox
+SHA=$(git rev-parse HEAD)
+popd
+
+# checkout sibling Velox to that SHA
+pushd ../../velox
+get checkout ${SHA}
```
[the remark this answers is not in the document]
simoneves: Well, that's embarrassing. I guess I didn't even test that script...

.github/workflows/presto-nightly-upstream.yml
```diff
@@ -1,10 +1,21 @@
 name: Presto Nightly Test (Upstream)
```
simoneves: All-new top-level workflow. The version currently in `main` is just a placeholder to make it appear in the list, so ignore the diff.

.github/workflows/presto-nightly-upstream.yml
```diff
@@ -1,10 +1,21 @@
...
-    steps:
-      - name: dummy
-        run: echo dummy
+  nightly:
+    uses: ./.github/workflows/presto-test.yml
+    with:
+      presto_repository: prestodb/presto
+      presto_commit: master
+      velox_repository: IGNORED
+      velox_commit: IGNORED
```
simoneves: These two are passed in, but deliberately invalid as they should never be used in the composite workflow in upstream mode.

.github/workflows/presto-nightly-upstream.yml
```diff
@@ -1,10 +1,21 @@
...
+      run_java_tests: true
+      run_cpu_tests: true
+      run_gpu_tests: true
+      set_velox_backward_compatible: false
```
simoneves: Deliberately do NOT invoke this option for upstream builds, as they should work without it.

.github/workflows/presto-nightly.yml
```diff
@@ -17,4 +17,5 @@ jobs:
       run_cpu_tests: true
       run_gpu_tests: true
       set_velox_backward_compatible: ${{ vars.SET_PRESTO_VELOX_BACKWARD_COMPATIBLE }}
+      build_upstream: false
```
simoneves: Just add this to the regular nightly job

.github/workflows/presto-test.yml
```diff
@@ -44,18 +39,22 @@ on:
...
+        type: boolean
+        default: false
 
   workflow_call:
     inputs:
       presto_repository: *presto_repository
       presto_commit: *presto_commit
       velox_repository: *velox_repository
       velox_commit: *velox_commit
-      velox_testing_commit: *velox_testing_commit
```
simoneves: Removed per above

.github/workflows/presto-test.yml
```diff
@@ -44,18 +39,22 @@ on:
...
       run_java_tests: *run_java_tests
       run_cpu_tests: *run_cpu_tests
       run_gpu_tests: *run_gpu_tests
       set_velox_backward_compatible: *set_velox_backward_compatible
+      build_upstream: *build_upstream
```
simoneves: Pass down

.github/workflows/presto-test-composite.yml
```diff
@@ -51,18 +50,29 @@ jobs:
...
+          echo "Found Presto pinned Velox SHA: ${PRESTO_PINNED_VELOX_SHA}"
+          echo "PRESTO_PINNED_VELOX_SHA=${PRESTO_PINNED_VELOX_SHA}" >> $GITHUB_OUTPUT
+          popd
       - name: Checkout Velox
         if: ${{ inputs.presto_worker_type != 'java' }}
         uses: actions/checkout@v4
         with:
-          repository: ${{ inputs.velox_repository }}
-          ref: ${{ inputs.velox_commit }}
+          repository: ${{ inputs.build_upstream && 'facebookincubator/velox' || inputs.velox_repository }}
```
simoneves: YAML ternaries are a bit ugly, but this basically uses the upstream Velox repo and the SHA determined above in upstream mode, and the regular passed-in repo and SHA in normal mode.

.github/workflows/presto-nightly.yml
```diff
@@ -16,5 +16,6 @@ jobs:
       run_java_tests: true
       run_cpu_tests: true
       run_gpu_tests: true
-      set_velox_backward_compatible: ${{ vars.SET_PRESTO_VELOX_BACKWARD_COMPATIBLE }}
+      set_velox_backward_compatible: ${{ vars.SET_PRESTO_VELOX_BACKWARD_COMPATIBLE == 'true' }}
```
simoneves: This fixes the breakage introduced by #94 which was not found in my testing of that as I didn't actually run the Nightly workflow, only the Presto Test one directly (blush)

.github/workflows/presto-nightly-upstream.yml
simoneves: Note that the placeholder version is bugged, which is making GHA unhappy in general right now. If we don't want to land the new parts of this one yet, I need to split out a fix for that. Discuss.

PR #94, opened 2025-10-23
Add option to Presto CI to set VELOX_ENABLE_BACKWARD_COMPATIBLE in Velox compile

simoneves:
This is a useful option which allows Presto compiles to work even if there has been a Velox API change and hence a temporary mismatch, as long as the developers have used `#ifdef VELOX_ENABLE_BACKWARD_COMPATIBILITY` correctly (which they usually do).

Right now, there is a mismatch which requires this option. This PR adds a CI job option to conditionally patch the Presto Velox CMake config to add that compile definition, which may allow builds to succeed while the mismatch is resolved.

Test run which proves the principle. This build would otherwise fail due to the current mismatch.

https://github.com/rapidsai/velox-testing/actions/runs/18736659234/job/53444764794

### DIFF
--- a/.github/workflows/presto-test-composite.yml
+++ b/.github/workflows/presto-test-composite.yml
@@ -77,10 +77,11 @@ jobs:
           ./fetch_centos_deps_image.sh
           popd
       - name: Set Velox Backward Compatibility Flag
+        # for Velox builds insert VELOX_ENABLE_BACKWARD_COMPATIBILITY definition?
         if: ${{ inputs.presto_worker_type != 'java' && inputs.set_velox_backward_compatible == 'true' }}
         run: |
-          pushd presto
-          echo 'add_definitions(-DVELOX_ENABLE_BACKWARD_COMPATIBILITY)' >> CMakeLists.txt
+          pushd presto/presto-native-execution
+          sed -i 's/add_subdirectory(presto_cpp)/add_definitions(-DVELOX_ENABLE_BACKWARD_COMPATIBILITY)\nadd_subdirectory(presto_cpp)/' CMakeLists.txt
           popd
       - name: Build Presto
         run: |


.github/workflows/presto-nightly.yml
```diff
@@ -16,4 +16,5 @@ jobs:
       run_java_tests: true
       run_cpu_tests: true
       run_gpu_tests: true
+      set_velox_backward_compatible: ${{ vars.SET_PRESTO_VELOX_BACKWARD_COMPATIBLE }}
```
simoneves: Fetches value from GH Variable (added and defined as `true`)

.github/workflows/presto-test-composite.yml
```diff
@@ -72,6 +76,13 @@ jobs:
           pushd velox-testing/presto/scripts
           ./fetch_centos_deps_image.sh
           popd
+      - name: Set Velox Backward Compatibility Flag
+        # for Velox builds insert VELOX_ENABLE_BACKWARD_COMPATIBILITY definition?
+        if: ${{ inputs.presto_worker_type != 'java' && inputs.set_velox_backward_compatible == 'true' }}
+        run: |
+          pushd presto/presto-native-execution
+          sed -i 's/add_subdirectory(presto_cpp)/add_definitions(-DVELOX_ENABLE_BACKWARD_COMPATIBILITY)\nadd_subdirectory(presto_cpp)/' CMakeLists.txt
```
simoneves: This is a bit hacky, but avoids the need for a patch file. The definition needs to go before the `add_subdirectory` otherwise it has no effect, which is why this isn't just a simple `cat >>`. Also more robust in case the structure of that file changes.

.github/workflows/presto-test.yml
```diff
@@ -64,6 +69,7 @@ jobs:
       velox_repository: ${{ inputs.velox_repository }}
       velox_commit: ${{ inputs.velox_commit }}
       velox_testing_commit: ${{ inputs.velox_testing_commit }}
+      set_velox_backward_compatible: false
```
simoneves: Not relevant to Java builds

.github/workflows/presto-nightly.yml
Avinash-Raj: fallback to `false` if the var isn't defined

Avinash-Raj (approved):
LGTM :-)

PR #95, opened 2025-10-23
Velox: Limit num threads while running gpu tests

Avinash-Raj:


.github/workflows/velox-test.yml
```diff
@@ -93,7 +93,7 @@ jobs:
           echo "   - CUDF support: DISABLED"
           echo "   - Target: CPU-only build"
           cd ${{ github.workspace }}/velox-testing/velox/scripts
-          TREAT_WARNINGS_AS_ERRORS=0 ./build_velox.sh --all-cuda-archs --cpu --log build_velox_cpu.log
```
Avinash-Raj: Removed `--al-cuda-archs` in this PR itself in-order to greedily check the Velox Nightly job, Sorry @simoneves

velox/scripts/config.sh
```diff
@@ -4,4 +4,4 @@
 CONTAINER_NAME="velox-adapters-build"
 COMPOSE_FILE="../docker/docker-compose.adapters.build.yml"
 
-NUM_THREADS=${NUM_THREADS:-$(( ($(nproc) > 2 ? $(nproc) - 2 : 1) ))}
```
Avinash-Raj: Fallback to 3/4th, to avoid CI lost communication with runner node error.

.github/workflows/velox-test.yml
simoneves: One of my PRs takes this out too, so no worries!

velox/scripts/test_velox.sh
```diff
@@ -104,7 +104,8 @@ test_preamble='if [ -f "/opt/miniforge/etc/profile.d/conda.sh" ]; then
 if [[ "$DEVICE_TYPE" == "cpu" ]]; then
   test_cmd="ctest -j ${NUM_THREADS} --label-exclude cuda_driver --output-on-failure --no-tests=error -E \"velox_exec_test|velox_hdfs_file_test|velox_hdfs_insert_test\""
 else
-  test_cmd="ctest -j ${NUM_THREADS} --output-on-failure --no-tests=error -E \"velox_exec_test|velox_hdfs_file_test|velox_hdfs_insert_test|velox_s3\""
+  # Run cuda_driver tests with 1 thread
+  test_cmd="ctest -j 1 -L cuda_driver --output-on-failure --no-tests=error -E \"velox_exec_test|velox_hdfs_file_test|velox_hdfs_insert_test|velox_s3\""
```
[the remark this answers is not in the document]
simoneves: Also maybe move the string of test names to a constant to make it more visible and this line of code easier to parse.

velox/scripts/config.sh
Avinash-Raj: I'm not sure about that.  If we use total_cores - 2 or total_cores, CI server lost communication with the runner machine.

simoneves (approved):
This is fine, if we want this script to be smart about threading when run locally, although I can't help but think that CI should always just be single-threaded, which could be achieved by just putting `--num-threads 1` in the invocation of this script from the `Velox Test` workflow.

PR #101, opened 2025-10-23
Update sccache version and configurations

trxcllnt:
This PR updates the sccache client version and a few configurations to hopefully improve build performance.

1. The latest `sccache@v0.12.0-rapids.1` release defaults to using HTTP/2 connection pooling, which ensures long-lived connections don't unexpectedly disconnect
2. Increase `dist.net.request_timeout` to match the build cluster's max job timeout
3. Update `SCCACHE_DIST_URL` to use docker's `TARGETARCH` var to support dist-compiling both x86 and aarch64
4. Define cache mounts for the sccache preprocessor and toolchains caches across `docker build` invocations
5. If `sccache` is enabled, add `-t=100` to `NVCC_APPEND_FLAGS` so multiple architectures are built in parallel
6. Mount sccache auth configs as docker build secrets
7. Work around gcc bug raising `range-loop-construct` warnings when building remotely

### DIFF
--- a/velox/docker/adapters_build.dockerfile
+++ b/velox/docker/adapters_build.dockerfile
@@ -135,11 +135,6 @@ if [ "$ENABLE_SCCACHE" = "ON" ]; then
   # Add sccache CMake flags
   EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache -DCMAKE_CUDA_COMPILER_LAUNCHER=sccache";
   export NVCC_APPEND_FLAGS="${NVCC_APPEND_FLAGS:+$NVCC_APPEND_FLAGS }-t=100";
-  if ! test -v SCCACHE_NO_DIST_COMPILE; then
-    # Work around gcc bug when dist-compiling preprocessor output
-    export CFLAGS="${CFLAGS:+$CFLAGS }-Wno-error=range-loop-construct";
-    export CXXFLAGS="${CXXFLAGS:+$CXXFLAGS }-Wno-error=range-loop-construct";
-  fi
 fi
 
 if test -n "${MAX_HIGH_MEM_JOBS:-}"; then


velox/docker/adapters_build.dockerfile
```diff
@@ -46,68 +89,77 @@ ${BUILD_BASE_DIR}/${BUILD_TYPE}/_deps/rmm-build:\
...
-    fi
+# Install and configure sccache if enabled
+if [ "$ENABLE_SCCACHE" = "ON" ]; then
+  # Run sccache setup script
+  bash /sccache_setup.sh;
+  # Add sccache CMake flags
+  EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache -DCMAKE_CUDA_COMPILER_LAUNCHER=sccache";
+  export NVCC_APPEND_FLAGS="${NVCC_APPEND_FLAGS:+$NVCC_APPEND_FLAGS }-t=100";
+  if ! test -v SCCACHE_NO_DIST_COMPILE; then
+    # Work around gcc bug when dist-compiling preprocessor output
```
mattgara: Good idea with the workaround :+1:

velox/docker/docker-compose.adapters.build.yml
```diff
@@ -5,7 +5,17 @@ services:
...
       dockerfile: velox-testing/velox/docker/adapters_build.dockerfile
+      ulimits: {nofile: {soft: 500000, hard: 500000}}
+      secrets:
+        - github_token
+        - aws_credentials
     runtime: ${DOCKER_RUNTIME:-nvidia}
     volumes:
       - ../../../velox:/workspace/velox
     working_dir: /workspace/velox
+
```
mattgara: TIL the use of `secrets` are so much cleaner than copying these values around.

mattgara (approved):
Thank you @trxcllnt, LGTM

velox/docker/adapters_build.dockerfile
trxcllnt: I've seen this at least once before (with clang), so I'll make `sccache` augment the user's compile flags when performing a distributed compilation. I'll likely have a default list that can be overridden via envvar like `SCCACHE_DIST_WARNINGS_NOT_ERRORS=range-loop-construct;reserved-user-defined-literal`

velox/docker/docker-compose.adapters.build.yml
simoneves: Interesting. Not being a YAML person until I was obliged to deal with GHA workflow files, I guess I assumed that `secrets` was specific to that. Is it actually a generic scoped dictionary?

simoneves: Is there a non-secret version, like GHA `vars`?

trxcllnt: The docker compose specification on secrets is [here](https://docs.docker.com/compose/how-tos/use-secrets/#build-secrets), and docker build secrets docs are [here](https://docs.docker.com/build/building/secrets/#using-build-secrets). The composefile spec is yaml, but the compose plugin is essentially syntactic sugar for orchestrating multiple containers, so most of its features map down to underlying `docker-cli` flags.

That's why the compose plugin parses additional syntax like `${SCCACHE_AUTH_DIR:-${HOME}/.sccache-auth}` -- this is a regular string to a yaml parser, and compose further interprets it in a shell-like way.

trxcllnt: This is another reason I'm not a huge fan of compiling code via `docker build`, and nearly always opt to compiling code via `docker run`. There are just many more features available when running a container vs. building one. If I need to make a container with the contents of a build, I will break it out into two steps:

```shell
docker run \
    -v $PWD:/src \
    -v $PWD/out:/out \
    -v $PWD/secrets:/run/secrets:ro \
    my-devtools sh -c '
FOO=$(cat /run/secrets/foo) cmake -S /src -B /out;
cmake --build /out;
'

docker build -f Dockerfile -t my-build-img ./out
```

simoneves: Interesting. Thank you.

PR #103, opened 2025-10-24
Presto: Reinstate a query optimizer flag for GPU only (ALSO SEE PR #111)

simoneves:
TPCH Q21 on GPU requires this flag to be on for best performance, but CPU requires it to be off.

As there is no way to pass CPU-or-GPU state through `pbench genconfig`, this required diverging the CPU and GPU Coordinator config files, which are otherwise identical.

### DIFF
--- a/presto/docker/docker-compose.native-cpu.yml
+++ b/presto/docker/docker-compose.native-cpu.yml
@@ -4,7 +4,7 @@ services:
       file: docker-compose.common.yml
       service: presto-base-coordinator
     volumes:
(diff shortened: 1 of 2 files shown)


presto/docker/config/template/etc_coordinator/config_native_cpu.properties
```diff
@@ -0,0 +1,112 @@
+# Run this node as the cluster coordinator; it schedules and manages queries.
```
simoneves: This file is identical to the GPU version other than the one flag

PR #106, opened 2025-10-24
Fixup `build_velox.sh`

mattgara:
PR #101 introduces some minor issues that this PR addresses. This PR is based on #101 and is intended to merge after #101 merges. This also does some minor fixes unrelated to #101.

Fixes/updates:

* Fix issue with non `sccache` build workflow not working due to secrets related to `sccache` being required. (Related to #101)

* Make ninja update optional (enabled by default). (Related to #101)
    
* Update nsys install to use correct repo and key for centos9.

* Keeping in line with other scripts do not force change the current workdir.

* Minor updates to text that is printed to user to fix outdated info.

### DIFF
--- a/velox/docker/docker-compose.adapters.build.yml
+++ b/velox/docker/docker-compose.adapters.build.yml
@@ -5,14 +5,7 @@ services:
     build:
       context: ../../..
       dockerfile: velox-testing/velox/docker/adapters_build.dockerfile
-      ulimits: {nofile: {soft: 500000, hard: 500000}}
     runtime: ${DOCKER_RUNTIME:-nvidia}
     volumes:
       - ../../../velox:/workspace/velox
     working_dir: /workspace/velox
-
-secrets:
-  github_token:
-    file: "${SCCACHE_AUTH_DIR:-${HOME}/.sccache-auth}/github_token"
-  aws_credentials:
-    file: "${SCCACHE_AUTH_DIR:-${HOME}/.sccache-auth}/aws_credentials"


velox/scripts/build_velox.sh
```diff
@@ -56,7 +50,9 @@ Options:
   -j|--num-threads            NUM Number of threads to use for building (default: 3/4 of CPU cores).
   --benchmarks true|false     Enable benchmarks and nsys profiling tools (default: true).
   --sccache                   Enable sccache distributed compilation caching (requires auth files in ~/.sccache-auth/).
+  --sccache-version           Install a specific version of rapidsai/sccache, e.g. "0.12.0-rapids.1" (default: latest)
```
simoneves: Why does this need to be an option? In what situations would you want to install something different from `latest`?

trxcllnt: Main reason is in case a release breaks `velox-testing` and you want to roll back to an earlier version. I don't anticipate that'll be common, but I like providing the option. Defaulting to installing the latest version makes it easier to ensure all users have the latest fixes, like the recent fixes to HTTP/2 connection pooling.

Feel free to remove the option if you have enough confidence in me to not break things :wink:, or you could hide it behind an envvar like `SCCACHE_VERSION=<older ver> build-velox.sh` instead of a flag.

velox/scripts/build_velox.sh
```diff
@@ -56,7 +50,9 @@ Options:
...
   --sccache-enable-dist       Enable distributed compilation (WARNING: may cause compilation differences like additional warnings that could lead to build failures).
+  --update-ninja true|false   Update ninja build tool during build (default: true).
```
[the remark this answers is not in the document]
trxcllnt: This one I'm less certain about. The ninja in the image is extremely old, I can't think of any reason you wouldn't want to use one of the more recent releases.

velox/scripts/build_velox.sh
simoneves: This one makes sense. No preference on implementation, as both would require GHA workflow plumbing.

velox/scripts/build_velox.sh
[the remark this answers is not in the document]
simoneves: I imagine that if we need a newer `ninja` for Velox, we'll also need it when we get `sccache` into Presto builds too.

PR #112, opened 2025-10-28
Updated velox benchmark workflow w.r.t recent addition of velox deps image pull from S3

Avinash-Raj:
This PR comprises of,

 - Updated existing Velox Benchmarking Workflow in-order to support the recent addition of "loading centos dependency image from S3" for building Velox.
 - Updated the `velox/benchmarks/tpch.sh` script to no longer pass the `VELOX_CUDF` CLI arguments 
 `--velox_cudf_enabled=true --velox_cudf_memory_resource=async` to the cudf benchmark executable, as these two arguments are no longer supported
[
Here is the successful benchmark run using the code from this PR.](https://github.com/rapidsai/velox-testing/actions/runs/18866676909) . You could also download the benchmark results artifact from the above job link.


velox/benchmarks/tpch.sh
```diff
@@ -202,7 +202,7 @@ run_tpch_single_benchmark() {
       cudf_pass_read_limit=0
       BENCHMARK_EXECUTABLE="$(get_tpch_benchmark_executable_path "$device_type")"
       CUDF_FLAGS="--cudf_chunk_read_limit=${cudf_chunk_read_limit} --cudf_pass_read_limit=${cudf_pass_read_limit}"
-      VELOX_CUDF_FLAGS="--velox_cudf_enabled=true --velox_cudf_memory_resource=async"
+      VELOX_CUDF_FLAGS=""
```
mattgara: * I think we need some way to be able to control the memory resource. This needs to be passed as either CLI option (for older branches) or as a configuration parameter (for newer branched.)

* I'm not clear if the `velox_cudf_enabled` is necessary (worth checking), otherwise if not specified we might just be testing CPU runs. Can you please confirm?

* Can we support both the case where these CLI are supported but also when these are configuration parameters? A natural way to tell which way to pass this might be by checking the `--help` menu of the benchmark binary.

Avinash-Raj: As for my understanding there are two executables are generated one for gpu and other for cpu benchmarking. Gpu benchmark executable is supposed to work for cudf since no need to pass cudf enabled flag. And for the other flag, we need guidance. @karthikeyann @devavret May I know your thoughts on this?

devavret: cudfEnabed is defaulted to true in newer branch. If you're testing CPU execution with a separate executable, you don't need to set it. If you're using the same executable, you need to set it to false.

Avinash-Raj: @mattgara We could pass `--velox_cudf_memory_resource=async` cli argument once this [PR](https://github.com/rapidsai/velox/pull/87) gets bmerged with the `merged-prs` branch. Thanks for the PR @devavret.

.github/workflows/velox-benchmark.yml
```diff
@@ -40,17 +40,40 @@ jobs:
...
 
-      - name: Setup Velox Testing and Velox
+      - name: Setup Velox
         uses: ./velox-testing/.github/actions/velox-setup
         with:
           repository: ${{ inputs.repository }}
           velox_commit: ${{ inputs.velox_commit }}
 
-      - name: "Build Velox GPU with Benchmarks (SHA: ${{ env.VELOX_COMMIT_SHA }})"
+      - name: Download and Load Velox Dependencies Container Image
```
paul-aiyedun: We seem to have some duplication between this file and velox-test.yml. Can we factor out common steps into a shared file/action?

velox/benchmarks/tpch.sh
mattgara: Yes I think that's fine, can you file a github issue for this to track it (so we don't forget to re-add later?) Thanks!

PR #115, opened 2025-10-30
Presto: Fix Java memory config

simoneves:
The recent PR #109 broke Java mode because of a typo in the config file mapping.

Also, there are still occasional failures due to the JVM heap size being set too large
compared with the remainder of the memory options, which are hard-wired in Java
mode, so we hard-wire that too instead of it being RAM-size dependent. We do this
in the generate script as a post-modification to avoid having to fork the config file.

### DIFF
--- a/presto/docker/docker-compose.java.yml
+++ b/presto/docker/docker-compose.java.yml
@@ -5,7 +5,7 @@ services:
       service: presto-base-coordinator
     volumes:
       - ./config/generated/java/etc_common:/opt/presto-server/etc
-      - ./config/generated/java/etc_coordinator/config_java.properties:/opt/presto-server/etc/config.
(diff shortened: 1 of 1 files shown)


presto/docker/docker-compose.java.yml
```diff
@@ -5,7 +5,7 @@ services:
       service: presto-base-coordinator
     volumes:
       - ./config/generated/java/etc_common:/opt/presto-server/etc
-      - ./config/generated/java/etc_coordinator/config_java.properties:/opt/presto-server/etc/config.
+      - ./config/generated/java/etc_coordinator/config_java.properties:/opt/presto-server/etc/config.properties
```
simoneves: Not sure how (1) I broke this, and (2) didn't find it before landing #109

simoneves:
Closing as replaced bt #117

PR #119, opened 2025-11-02
Presto: Add Nightly (Upstream) CI (alternative implementation 2)

simoneves:
OK, well, this appears to work, but only by collapsing the body of `Presto Test` in as additional jobs in the same sequence with their own `runs-on` tags, rather than as a single job which calls another workflow which spawns multiple jobs.

Again, maybe I got the intermediate version wrong.


simoneves:
Closing as replaced by #120 because I'm stupid

PR #120, opened 2025-11-02
Presto: Add Nightly (Upstream) CI (alternative to #110)

simoneves:
This works too.

My concern remains the expense of the additional job submission, but that does only take about 20 seconds to run.

This reverts some of the changes to Presto Test and Presto Test Composite done in #110, leaving only the change to remove the `velox_testing_commit` parameter, which is superfluous, because the name of the branch from which the workflow was run can be fetched from a GH built-in variable.

There appears to be a new problem with upstream builds, however, in that they now suffer from the SHARED/STATIC library mismatch problem.

### DIFF
--- a/.github/workflows/presto-nightly-upstream.yml
+++ b/.github/workflows/presto-nightly-upstream.yml
@@ -31,39 +31,16 @@ jobs:
         id: set_variable_step
         run: |
           echo "PRESTO_PINNED_VELOX_SHA=${{ steps.get-presto-pinned-velox-version.outputs.PRESTO_PINNED_VELOX_SHA }}" >> $GITHUB_OUTPUT
-  java:
+  nightly:
     needs: get-presto-pinned-velox
-    uses: ./.github/workflows/presto-test-composite.yml
-    with:      
-      presto_worker_type: 'java'
-      node_label: 'linux-amd64-cpu8'
-      presto_repository: prestodb/presto
-      presto_commit: master
-      velox_repository: facebookincubator/velox
-      velox_commit: ${{ needs.get-presto-pinned-velox.outputs.PRESTO_PINNED_VELOX_SHA }}
-      set_velox_backward_compatible: false
-    secrets: inherit
-  native-cpu:
-    needs: get-presto-pinned-velox
-    uses: ./.github/workflows/presto-test-composite.yml
+    uses: ./.github/workflows/presto-test.yml
     with:
-      presto_worker_type: 'native_cpu'
-      node_label: 'linux-amd64-cpu16'
       presto_repository: prestodb/presto
       presto_commit: master
       velox_repository: facebookincubator/velox
       velox_commit: ${{ needs.get-presto-pinned-velox.outputs.PRESTO_PINNED_VELOX_SHA }}
-      set_velox_backward_compatible: false
-    secrets: inherit
-  native-gpu:
-    needs: get-presto-pinned-velox
-    uses: ./.github/workflows/presto-test-composite.yml
-    with:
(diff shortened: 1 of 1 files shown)


.github/workflows/velox-nightly-staging.yml
```diff
@@ -1,4 +1,4 @@
-name: Velox Nightly Test
+name: Velox Nightly Test (Staging)
```
simoneves: Rename

.github/workflows/presto-test-composite.yml
```diff
@@ -62,7 +57,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: rapidsai/velox-testing
-          ref: ${{ inputs.velox_testing_commit }}
+          ref: ${{ github.ref_name }} # automatically match the workflow branch
```
simoneves: Use `github.ref_name` to automatically match the `velox-testing` branch to that of the workflow being run (`main` or dev).

.github/workflows/presto-test.yml
```diff
@@ -23,11 +23,6 @@ on:
         type: string
         required: false
         default: 'main'
-      velox_testing_commit: &velox_testing_commit
-        description: 'Velox Testing Commit SHA or Branch'
-        type: string
-        required: false
-        default: 'main'
       run_java_tests: &run_java_tests
```
simoneves: As above

.github/workflows/presto-nightly-upstream.yml
```diff
@@ -1,16 +1,44 @@
...
   nightly:
+    needs: get-presto-pinned-velox
     uses: ./.github/workflows/presto-test.yml
     with:
       presto_repository: prestodb/presto
       presto_commit: master
-      velox_repository: IGNORED
-      velox_commit: IGNORED
+      velox_repository: facebookincubator/velox
+      velox_commit: ${{ needs.get-presto-pinned-velox.outputs.PRESTO_PINNED_VELOX_SHA }}
```
simoneves: Get the SHA from the previous job output

.github/workflows/presto-nightly-upstream.yml
```diff
@@ -1,16 +1,44 @@
 name: Presto Nightly Test (Upstream)
 
 on:
+  schedule:
+    - cron: '0 4 * * *'  # daily at 04:00 UTC
   workflow_dispatch:
 
 jobs:
+  get-presto-pinned-velox:
```
simoneves: A whole pre-job which also checks out Presto and fetches the SHA of the pinned Velox sub-module.

.github/workflows/presto-nightly-upstream.yml
```diff
@@ -1,16 +1,44 @@
...
+    runs-on: linux-amd64-cpu8 # wasteful but no reliable smaller runner available
+    outputs:
+      presto_pinned_velox_sha: ${{ steps.set_variable_step.outputs.PRESTO_PINNED_VELOX_SHA }}
+    steps:
+      - name: Checkout Presto
+        uses: actions/checkout@v4
```
simoneves: This can likely be a shallow clone since this isn't the one we're gonna build.

simoneves: Can probably add the options `fetch-depth: 1` to not pull all history, and `submodules: true` to avoid the need for the `make submodules`.

The same would surely apply to ALL checkouts we do for CI, as we don't need history or anything but the specified branch.

These repos aren't very big, though. In my testing of a manual clone with the command-line options the above tags invoke, it only saved a few seconds on each checkout.

PR #127, opened 2025-11-14
Added GHA job for uploading velox deps image to S3

Avinash-Raj:
This PR adds a GitHub Actions workflow for building and uploading the latest Velox dependency image to S3. Any previously uploaded image is automatically overwritten.


.github/workflows/velox-deps-upload.yml
```diff
@@ -0,0 +1,79 @@
...
+        required: false
+        default: 'main'
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build-and-upload-deps:
+    runs-on: linux-amd64-cpu4
```
simoneves: Presumably we need an ARM version of this too? Are there any cheap CPU-only ARM GHA runners?

.github/workflows/velox-deps-upload.yml
```diff
@@ -0,0 +1,79 @@
...
+        working-directory: ${{ github.workspace }}/velox-testing/velox/scripts
+        run: ./build_centos_deps_image.sh
+
+      - name: Upload Velox Dependencies Container Image to S3
+        env:
+          AWS_ARN_STRING: ${{ secrets.AWS_ARN_STRING }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          S3_BUCKET_NAME: ${{ vars.S3_BUCKET_NAME }}
+          S3_BUCKET_REGION: ${{ vars.S3_BUCKET_REGION }}
```
simoneves: IIUC, these creds are read-only? Do we need a different set of creds for this process?

simoneves (commented):
LGTM but need ARM version and also verify creds to write to bucket.

We might also consider a versioning system, or at least a history system, whereby we rename the existing file as `old/filename_DATE` before pushing the new one.

mattgara (commented):
Could you please elaborate on how we *choose* which version of the deps images to build? 

That to me seems like a critical point, I don't think we want to be building every single commit (since many of them will have no deps updates.)

.github/workflows/velox-deps-upload.yml
Avinash-Raj: Yes, you were correct. We need different creds (with read/write access).

Avinash-Raj:
[Successful CI Run](https://github.com/rapidsai/velox-testing/actions/runs/21536202172)

Avinash-Raj:
@simoneves @paul-aiyedun Updated the workflow as per our recent recent discussion of overriding the existing images on S3.

simoneves (approved):
LGTM

PR #128, opened 2025-11-14
Add Cluster Tag

johallar:
Adds cluster tag config to our velox-testing start scripts for java, cpu, gpu. 

Sister PR to https://github.com/prestodb/presto/pull/26485, which will display this tag in the UI so we can easily identify which type of cluster we have created.


simoneves:
> The PR that uses this flag and UI changes to display it linked in the PR description 😄

Ugh. Sorry. My brain was fuzz yesterday.

PR #134, opened 2025-11-24
Misiug/multi node docker containers

misiugodfrey:
Support in docker containers to run multiple workers (each with a separate pinned gpu) on a single machine.  You can now run multiple workers (up to 4) controlled through the NUM_WORKERS env variable.

I'm not aware of a way to clean up the duplication in the docker services (AFAIK you can't programatically change the gpu parameters).  So this set up declares 4 gpu services and then picks them based on the number of workers you want to run.


presto/docker/config/template/etc_worker/config_native.properties
```diff
@@ -1,7 +1,7 @@
 # This node is a worker; it executes tasks but doesn't coordinate queries.
 coordinator=false
 # Worker REST/HTTP port for internal and admin endpoints.
-http-server.http.port=8080
+http-server.http.port=8081
```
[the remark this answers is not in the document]
paul-aiyedun: Can you please expand on why this is required now? This should be running as a separate service on the docker network.

presto/docker/docker-compose.native-gpu.yml
```diff
@@ -28,3 +28,44 @@ services:
       - ./config/generated/gpu/etc_common:/opt/presto-server/etc
       - ./config/generated/gpu/etc_worker/node.properties:/opt/presto-server/etc/node.properties
       - ./config/generated/gpu/etc_worker/config_native.properties:/opt/presto-server/etc/config.properties
+
+# These workers are available to run on one node with a single GPU pinned to each.
+  presto-native-worker-gpu-0:
```
paul-aiyedun: We seem to be hardcoding and deploying a fixed number of workers?

presto/scripts/generate_presto_config.sh
```diff
@@ -97,9 +97,9 @@ EOF
     # optimizer.default-filter-factor-enabled=true
     COORD_CONFIG="${CONFIG_DIR}/etc_coordinator/config_native.properties"
     sed -i 's/\#optimizer/optimizer/g' ${COORD_CONFIG}
-    
+
     # Adds a cluster tag for gpu variant
-    WORKER_CONFIG="${CONFIG_DIR}/etc_coordinator/config_native.properties"
+    WORKER_CONFIG="${CONFIG_DIR}/etc_worker/config_native.properties"
```
misiugodfrey: I don't think this was correct before; as we were referring to the coordinator's config as worker_config.

presto/scripts/stop_presto.sh
```diff
@@ -2,4 +2,7 @@
 
 set -e
 
-docker compose -f ../docker/docker-compose.java.yml -f ../docker/docker-compose.native-cpu.yml -f ../docker/docker-compose.native-gpu.yml down
+OVERRIDE=""
+[ -f ../docker/docker-compose.workers.override.yml ] && OVERRIDE="-f ../docker/docker-compose.workers.override.yml"
```
misiugodfrey: If we generated a multi-workers file, then use that to stop containers or we may leave them dangling.

presto/scripts/generate_presto_config.sh
simoneves: Are you sure? The configs that this `sed` is intended to un-comment are absolutely in the coordinator version of this file, not the worker!

PR #135, opened 2025-11-25
Add GPU benchmark stats guard

patdevinwilson:
Add a guard in presto/scripts/run_benchmark.sh that detects when the presto-native-gpu stack is active and verifies all Hive .prestoSchema stats files exist for the chosen TPCH schema before running GPU benchmarks. If any stats file is missing, the script instructs the user to run presto-native-cpu and ANALYZE tables first.

### DIFF
--- a/presto/scripts/run_benchmark.sh
+++ b/presto/scripts/run_benchmark.sh
@@ -44,8 +44,8 @@ verify_schema_stats() {
   if [[ $missing -ne 0 ]]; then
     cat <<EOF
 
-Column statistics are required before running GPU benchmarks.
-Run presto-native-cpu and ANALYZE all tables in schema '$schema', then retry.
+Column statistics are required before running TPCH benchmarks.
+Run ./analyze_tables.sh --schema-name '$schema' (and optional connection args) before retrying.
 
 EOF
     exit 1
@@ -203,13 +203,6 @@ if [[ -z ${SCHEMA_NAME} ]]; then
   exit 1
 fi
 
-IS_GPU_CLUSTER=0
-if command -v docker >/dev/null 2>&1; then
-  if docker ps --format '{{.Names}}' 2>/dev/null | grep -q 'presto-native-worker-gpu'; then
-    IS_GPU_CLUSTER=1
-  fi
-fi
-
 PYTEST_ARGS=("--schema-name ${SCHEMA_NAME}")
 
 if [[ -n ${QUERIES} ]]; then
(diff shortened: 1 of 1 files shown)


presto/scripts/run_benchmark.sh
```diff
@@ -167,6 +203,13 @@ if [[ -z ${SCHEMA_NAME} ]]; then
   exit 1
 fi
 
+IS_GPU_CLUSTER=0
+if command -v docker >/dev/null 2>&1; then
+  if docker ps --format '{{.Names}}' 2>/dev/null | grep -q 'presto-native-worker-gpu'; then
```
misiugodfrey: Is there any reason to only require ANALYZE to have been run for gpu clusters?  If we want an appropriate comparison, should we not check this for all clusters regardless of type?

If that's the case, then we can simplify here by removing all the gpu cluster checking and just always require ANALYZE to have been run.

mattgara:
This PR is a great initiative.  `velox-testing` could always use more guardrails and intuitive checks. 

> Please review: #136
> 
> I’ve reverted the previous stats-checking logic and repurposed the changes so the PR now ensures we automatically run ANALYZE right after tables are created by the setup scripts. This targets your earlier suggestion directly.

I think even if we implement an (optional or mandatory) call to analyze tables when setting up the benchmarks, it would still make sense to independently validate that the state we expect actually exists persistently (what this PR does/did originally.) 

It probably makes sense to do both :)

presto/scripts/run_benchmark.sh
```diff
@@ -16,6 +16,42 @@
 
 set -e
 
+TPCH_TABLES=(customer lineitem nation orders part partsupp region supplier)
+HIVE_METASTORE_DIR="$(readlink -f ../docker/.hive_metastore)"
```
paul-aiyedun: Should we be using something like the `SHOW STATS` command instead of going through Hive Metastore files?

patdevinwilson: > Should we be using something like the `SHOW STATS` command instead of going through Hive Metastore files?

Doing that up front for all TPCH tables adds noticeable overhead before every benchmark run. The command only works per table (and sometimes per column), so we’d need to run a Presto query for every table/column we care about. Some Presto versions return empty stats rows even when analysis hasn’t been done, so we’d still need to cross-check other metadata to be sure. That can get flaky when users customize their catalogs. Inspecting the .prestoSchema files in the Hive metastore gives us a fast, deterministic check if the JSON file exists and its columnStatistics block is non-empty, we know ANALYZE has been run at least once for that table. It’s essentially the ground truth that SHOW STATS would read anyway.

PR #136, opened 2025-11-25
Run ANALYZE after setting up benchmark tables

patdevinwilson:
I’ve reverted the previous stats-checking logic and repurposed the changes so the PR now ensures we automatically run ANALYZE right after tables are created by the setup scripts. This targets your earlier suggestion directly:
setup_benchmark_tables.sh now runs ./analyze_tables.sh --schema-name ... after creating the tables, so column stats are always produced.
setup_benchmark_data_and_tables.sh already calls the tables script, so it inherits the same behavior (no duplicate logic needed).
Both scripts print clear messaging before and after ANALYZE to show progress.

### DIFF
--- a/presto/scripts/setup_benchmark_data_and_tables.sh
+++ b/presto/scripts/setup_benchmark_data_and_tables.sh
@@ -22,8 +22,8 @@ the value set for the --data-dir-name argument."
 
 SCRIPT_EXAMPLE_ARGS="-b tpch -s my_tpch_sf100 -d sf100 -f 100 -c"
 
-SCRIPT_EXTRA_OPTIONS_DESCRIPTION="-f, --scale-factor                  The scale factor of the generated dataset.
-    -c, --convert-decimals-to-floats    Convert all decimal columns to float column type."
+RUN_ANALYZE=true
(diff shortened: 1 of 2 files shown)


presto/scripts/setup_benchmark_data_and_tables.sh
```diff
@@ -61,3 +61,7 @@ DATA_GEN_SCRIPT_PATH=$(readlink -f ../../benchmark_data_tools/generate_data_file
 $CONVERT_DECIMALS_TO_FLOATS_ARG
 
 ./setup_benchmark_tables.sh -b $BENCHMARK_TYPE -s $SCHEMA_NAME -d $DATA_DIR_NAME
+
+echo "Running ANALYZE TABLES for schema '$SCHEMA_NAME'..."
```
mattgara: It probably makes sense to make this optional. Otherwise, for the foreseeable future this would place a CPU worker requirement on this script (due to the requirement on the analyze tables script.)

That is add a command line option to this script that (may be enabled by default) will toggle this behaviour on/off.

misiugodfrey (changes requested):

misiugodfrey:
This functionality is currently in main, so I'm closing this PR.

PR #137, opened 2025-11-26
Temporary fix for Presto builds pending upstream presto-function-server fixes

simoneves:
See https://github.com/prestodb/presto/issues/26705

A new Presto module (and associated JAR file) was added so our Java build script needs to deploy it.

This works with current `master` although it would not have worked a few days ago when I originally discovered this problem. They have landed a partial fix since then, with more fixes (for parts which don't affect us) still in PR.

However, another Velox update happened before that fix, so `rapidsai/velox` is now incompatible. However, I apparently missed the memo that we're once again suspending `rapidsai/velox` in favor of working in upstream (or personal forks, at least) as allegedly everything that was in `IBM-techpreview` has now been upstreamed (TO BE CONFIRMED!)


paul-aiyedun:
Based on comments in https://github.com/prestodb/presto/issues/26705, this appears to be the long-term fix?

simoneves:
> Based on comments in [prestodb/presto#26705](https://github.com/prestodb/presto/issues/26705), this appears to be the long-term fix?

It does seem likely, and certainly works for now. We can revisit it if the dust settles differently on their side.

PR #144, opened 2025-12-12
Remove "Say Hello" GHA Workflow

simoneves:
This was intended as a template for testing workflow syntax. I think it has served its purpose and can be removed.


Avinash-Raj (approved):
LGTM!

PR #145, opened 2025-12-15
Fix duckdb vs presto differences in sql for tpcds

karthikeyann:
- Fix duckdb vs presto differences in sql for tpcds
Strings at some places need to be _casted_ as `date`, and interval days.
-  add scale-factor to benchmarks
-  print duration with integration test 

Only Q70, Q86 fail due to differences in capability difference of orderby and aggregations processed separately in presto.


presto/testing/common/conftest.py
```diff
@@ -22,13 +22,16 @@ def pytest_generate_tests(metafunc):
     if TPCDS_FIXTURE_NAME in metafunc.fixturenames:
         TPCDS_NUM_QUERIES = 99
         TPCDS_DISABLED_QUERIES = [
+            70, # PrestoUserError(type=USER_ERROR, name=SYNTAX_ERROR, message="Invalid reference to output of SELECT clause from grouping() expression in ORDER BY", query_id=20250815_182928_01495_uy5t2)
+            86, # PrestoUserError(type=USER_ERROR, name=SYNTAX_ERROR, message="Invalid reference to output of SELECT clause from grouping() expression in ORDER BY", query_id=20250815_182935_01511_uy5t2)
+        ]
+        # Following queries are fixed.
+        [
```
simoneves: Confused by the syntax here. Is the remainder of the original array now just an orphaned immediately-discarded stack declaration which has no effect? Would it be clearer to just comment them out?

PR #154, opened 2025-12-22
Fix for Presto build failures since airbase update

simoneves:
Presto builds (all variants) have been broken since this commit to upstream Presto...

https://github.com/prestodb/presto/commit/44200d7872b25f155b82852014f2cb5fc4504072

I reported it to IBM, but they did not respond. Probably because it turns out it's due to a staleness in our build process, which is a little embarrassing.

@quasiben came up with this fix, which despite a big red warning about how Node 18 is deprecated, seems to fix it, although it's not clear to me yet why. Perhaps an even newer version of Node would also work? Presumably the regular Presto build process avoids this somehow. More investigation needed. @cmatzenbach and @johallar please comment.

### DIFF
--- a/presto/scripts/build_presto_java_package.sh
+++ b/presto/scripts/build_presto_java_package.sh
@@ -32,7 +32,7 @@ docker run --rm \
     git config --global --add safe.directory /presto &&
     curl -fsSL https://deb.nodesource.com/setup_22.x | bash - &&
     apt-get install -y nodejs &&
-    npm install -g yarn &&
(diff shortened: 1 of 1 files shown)


johallar:
@simoneves The [pom.xml](https://github.com/prestodb/presto/blob/master/presto-ui/pom.xml#L60C1-L70C33) for `presto-ui` specifies 
```
<nodeVersion>v22.15.1</nodeVersion>
<yarnVersion>v1.22.22</yarnVersion>
```
So I think we should target v22.15.1 for node

PR #158, opened 2026-01-09
Re-enable Java CI and fix Java mode properly

simoneves:
The problem was nothing to do with `airbase` but some Parquet-specific config properties which Java mode apparently now barfs on, thus making the Java Worker not start.

This PR reverts the previous one, re-enabling Java CI builds, and adds logic to the config generation to take the above into account.

### DIFF
--- a/presto/scripts/generate_presto_config.sh
+++ b/presto/scripts/generate_presto_config.sh
@@ -113,7 +113,8 @@ EOF
   # for Java variant, disable some Parquet properties which are now rejected
   if [[ "${VARIANT_TYPE}" == "java" ]]; then
     HIVE_CONFIG="${CONFIG_DIR}/etc_worker/catalog/hive.properties"
-    sed -i 's/parquet/#parquet/g' ${HIVE_CONFIG}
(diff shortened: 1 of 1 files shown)


bdice (approved):

presto/scripts/generate_presto_config.sh
```diff
@@ -110,6 +110,12 @@ EOF
     echo "cluster-tag=native-cpu" >> ${WORKER_CONFIG}
   fi
 
+  # for Java variant, disable some Parquet properties which are now rejected
+  if [[ "${VARIANT_TYPE}" == "java" ]]; then
+    HIVE_CONFIG="${CONFIG_DIR}/etc_worker/catalog/hive.properties"
+    sed -i 's/parquet/#parquet/g' ${HIVE_CONFIG}
```
misiugodfrey: This probably won't be our preferred method in the future, but I'm ok with it for now.  That being said, we may want to be a little more specific as `*parquet*` could match a lot of potential options in the future that we don't want to be accidentally disabling.

PR #159, opened 2026-01-09
Temporary tweaks to Velox CI unit tests

simoneves:
* Raise ulimit -n (open files) to 4096 for CPU mode, avoiding failure in velox_table_evolution_fuzzer_test

Proper fix unclear.
Perhaps the code can be modified to not open so many files.
Or we just leave the ulimit raised, as is apparent done in a few other CI situations.

* Disable velox_cudf_s3_read_test in GPU mode, as this currently fails with a CUDA error on shutdown

Proper fix is https://github.com/rapidsai/rmm/pull/2202
Re-enable test once Velox is using an RMM with that in.


mattgara (commented):
LGTM

PR #160, opened 2026-01-12
[HERC-261] Default to pulling not fetching/building Velox deps container

simoneves:
In the early stages of this project, we decided to cache the two upstream deps container images...

  Velox: ghcr.io/facebookincubator/velox-dev:adapters
  Presto: presto/prestissimo-dependency:centos9

...mostly because the latter was not directly fetch-able from a public repository, but needed to be built locally.

They have not iterated very often (we have only had to update our version twice so far) but the Velox one has now bumped three times in a month, which breaks our CI.

Belatedly, I realized that since the Velox one IS reliably pull-able from their public repo, and we have had additional confirmation (from Christian Zentgraf) that the public repo is automatically and reliably updated, we should change our Velox CI jobs to just pull this image directly instead of using the local cache.

We should still retain the OPTION of using the local cache, to handle a local Velox fork with (say) a future cuDF update.

This PR adds an option to the core `velox-test` workflow to implement this.

### DIFF
--- a/.github/workflows/velox-benchmark-sanity-test.yml
+++ b/.github/workflows/velox-benchmark-sanity-test.yml
@@ -53,7 +53,6 @@ jobs:
 
       - name: Pull Velox Dependencies
         if: ${{ inputs.use_upstream_deps_image }}
-        name: Pull Velox Dependencies
         run: |
           docker pull ghcr.io/facebookincubator/velox-dev:adapters
 


mattgara (approved):
LGTM

simoneves:
Thanks for the approval, @paul-aiyedun , but unfortunately this PR appears insufficient. The compiles still fail the same way. So the problem must be deeper. I am resistant to thrashing back to this right now, though.

simoneves:
Turns out this WAS due to a breakage in actual upstream Velox, and NOT because of the deps image being out of date. A fix was pushed yesterday, and last night's Velox CI built OK.

This PR should still be discussed, but I think we either want to always pull upstream image or just keep the S3 cache, not both, because if we're USUALLY pulling the upstream image, the S3 cache will just go out of date without us realizing...

PR #166, opened 2026-01-16
Maximize row group size for rewritten parquet

lga-zurich:
This creates parquet files that result in better performance as the parquet decoding is more efficient not having to process as many row groups individually.


lga-zurich:
@paul-aiyedun ok, I see, if the parquet files are generated with a sufficiently large `parquet-row-group-bytes` then there is no need to rewrite them with larger rowgroups. Thanks for the update!

PR #177, opened 2026-01-19
S3 support

lga-zurich:
Allow to run benchmarks and data generation when data is on S3.


lga-zurich:
> @lga-zurich Thanks for adding this update. Would it be possible to extend the integration tests (i.e. the `run_integ_test.sh` workflow) to cover this use case?

I would love to, but I need to find the bandwidth for that.

PR #181, opened 2026-01-21
Add lukewarm times and sum to benchmark report

paul-aiyedun:
* Update benchmark report to distinguish between hot and lukewarm times.

* Add sum row to benchmark text report.

* Add iterations count to benchmark report.

* Show only lukewarm times for single iteration runs.


simoneves (approved):
TIL: more Python string formatting syntax, e.g. f"{'SUM':^10}"

misiugodfrey (approved):

PR #189, opened 2026-01-24
Automatically drop caches when running benchmarks

misiugodfrey:
Altered run_benchmark.sh to automatically drop caches when running (can be turned off via option).

Also added a script to run benchmarks with multiple configurations - useful for comparing benchmark settings quickly.


presto/scripts/run_multiple_benchmarks.sh
```diff
@@ -0,0 +1,81 @@
...
+        -w|--num-workers)
+            if [[ -n $2 ]]; then
+                IFS=',' read -ra WORKERS_ARRAY <<< "$2"
+                shift 2
+            else
+                echo "Error: --num-workers requires a value"
+                exit 1
+            fi
+            ;;
+        -s|--schema)
```
paul-aiyedun: `--schemas`

presto/scripts/run_multiple_benchmarks.sh
```diff
@@ -0,0 +1,81 @@
...
+        *)
+            echo "Error: Unknown argument $1"
+            print_help
+            exit 1
+            ;;
+    esac
+  done
+}
+
+export PRESTO_DATA_DIR=/raid/ocs_benchmark_data/tpch/experimental
```
paul-aiyedun: This should probably be set by a parameter if not set already.

presto/scripts/run_multiple_benchmarks.sh
```diff
@@ -0,0 +1,81 @@
...
+parse_args "$@"
+for schema in "${SCHEMA_ARRAY[@]}"; do
+    for kvikio in "${KVIKIO_ARRAY[@]}"; do
+        for drivers in "${DRIVERS_ARRAY[@]}"; do
+            for workers in "${WORKERS_ARRAY[@]}"; do
+                for memory in "${MEMORY_ARRAY[@]}"; do
+                    echo "Running combo: num_workers = $workers, kvikio_threads = $kvikio, num_drivers = $drivers, schema = $schema, memory = $memory"
+                    ./start_native_gpu_presto.sh -w $workers --kvikio-threads $kvikio --num-drivers $drivers --memory-percent $memory
+                    ./run_benchmark.sh -b tpch -s ${schema} --tag "${schema}_${workers}w_${drivers}d_${kvikio}k_${memory}m_dropcache"
```
paul-aiyedun: Consider `${schema}_${workers}wk_${drivers}dr_${kvikio}kv_${memory}mp`. The `dropcache` suffix should no longer be needed after extending the benchmark script to drop caches by default.

paul-aiyedun (commented):
Changes overall look good to me. However, I think cache dropping should be integrated into the benchmarking test suite.

karthikeyann (changes requested):

presto/scripts/generate_presto_config.sh
```diff
@@ -54,7 +54,7 @@ function duplicate_worker_configs() {
         ${coord_config}/config_native.properties
     sed -i "s+single-node-execution-enabled.*+single-node-execution-enabled=false+g" \
 	${worker_config}/config_native.properties
-  # make cudf.exchange=true if we are running multiple workers
+    # make cudf.exchange=true if we are running multiple workers
     sed -i "s+cudf.exchange=false+cudf.exchange=true+g" ${worker_config}/config_native.properties
   fi
   echo "join-distribution-type=PARTITIONED" >> ${coord_config}/config_native.properties
```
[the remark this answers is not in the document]
misiugodfrey: I've changed this so that `join-distribution-type=PARTITIONED` is in the base template.  If this will have no effect on single-node execution, then that is the right place for it.

PR #190, opened 2026-01-27
Use paths relative to the script and/or repository root

bdice:
This makes it possible to use the scripts from anywhere, rather than only specific directories.

### DIFF
--- a/presto/pbench/run_tpch.sh
+++ b/presto/pbench/run_tpch.sh
@@ -1,10 +1,12 @@
-
 #!/bin/bash
 
-BASE_DIR="$(dirname $(realpath $0))/../.."
+# Compute the directory where this script resides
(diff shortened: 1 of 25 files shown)


presto/pbench/run_tpch.sh
```diff
@@ -1,10 +1,12 @@
-
 #!/bin/bash
 
-BASE_DIR="$(dirname $(realpath $0))/../.."
+# Compute the directory where this script resides
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
paul-aiyedun: Is there a way to avoid repeating this line in all the scripts?

bdice: Nope, this is pretty much the best practice we have.

PR #195, opened 2026-01-28
Add username tag to Docker container name

shrshi:
Avoid conflicts due to multiple Presto coordinator and worker containers, and resolves the following error on a shared lab machine
```
Error response from daemon: Conflict. The container name "/presto-coordinator" is already in use by container "4ac48748c75b473172af9e359f9421200d70ec25d8cf466c1d4fd5f352acc982". You have to remove (or rename) that container to be able to reuse that name.
```

### DIFF
--- a/presto/docker/docker-compose.common.yml
+++ b/presto/docker/docker-compose.common.yml
@@ -14,7 +14,7 @@ services:
   presto-base-coordinator:
     extends:
       service: presto-base-java
-    container_name: presto-coordinator
+    container_name: presto-coordinator-${IMAGE_TAG:-latest}
     image: presto-coordinator:${IMAGE_TAG:-latest}
     ports:
       - 8080:8080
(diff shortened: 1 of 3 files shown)


presto/docker/docker-compose.common.yml
```diff
@@ -14,7 +14,7 @@ services:
   presto-base-coordinator:
     extends:
       service: presto-base-java
-    container_name: presto-coordinator
+    container_name: presto-coordinator-${IMAGE_TAG:-latest}
```
paul-aiyedun: I think we want to keep the default name as-is when `$IMAGE_TAG` is not set. Have you considered using the `--project-name`/`-p` argument for this (https://docs.docker.com/compose/how-tos/project-name/) i.e. update https://github.com/rapidsai/velox-testing/blob/f05e8523fbfe63e999e482959b804be08005a048/presto/scripts/start_presto_helper.sh#L167 and https://github.com/rapidsai/velox-testing/blob/f05e8523fbfe63e999e482959b804be08005a048/presto/scripts/stop_presto.sh#L10-L14 to conditionally pass in `-p $IMAGE_TAG` to the `docker compose` command, if `$IMAGE_TAG` is set?

misiugodfrey:
As far as I understand, this functionality has been added to main; I think we can close this as a duplicate PR.

PR #197, opened 2026-01-29
Fix pbench path

bdice:
Fixes a broken path.


shrshi (approved):
Thanks for the quick fix!

PR #200, opened 2026-01-31
Fix first time build target for multi-gpu presto

karthikeyann:
When multi-gpu presto is launched for first time, the docker image target is not available because it's named by gpu id. 
but target image is used instead of service name. This PR fixes it.


paul-aiyedun (approved):

karthikeyann:
issue fixed by different logic. closing the PR.

PR #209, opened 2026-02-04
introduce workflow to publish CI images to GHCR

jameslamb:
Contributes to https://github.com/rapidsai/github-infrastructure/issues/15 (private issue, sorry).

#127 introduced a workflow for using S3 to store private container images used in CI. This proposes an alternative, using GitHub's container registry (GHCR).

### Benefits of this approach

* 0 management of external secrets
* 0 new cloud cost
  - according to https://docs.github.com/en/billing/concepts/product-billing/github-packages#free-use-of-github-packages, container registry usage on GitHub Actions is free and unlimited
* can pull with `docker pull`, `docker run`, etc. ... no need to `docker load` / `docker save`

### How do we keep these images private?

The first build from this workflow created this package: https://github.com/rapidsai/velox-testing/pkgs/container/velox-testing-images

I manually changed it to `private`, following https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility.

<details><summary>screenshot (click me)</summary>

image

ref: https://github.com/rapidsai/velox-testing/settings

</details>

So if you're logged into GitHub and have an account in the `rapidsai` org, you'll see it from the homepage:

<details><summary>screenshot (click me)</summary>

image

</details>

But if not logged in, the package won't be visible.

<details><summary>screenshot (click me)</summary>

image

</details>

All future pushes will be private. By default, only this repo will be able to pull it in GitHub Actions jobs:

And you cannot `docker pull` it if unauthenticated.

```console
$ IMAGE_URI="ghcr.io/rapidsai/velox-testing-images:presto-deps-cu13-amd64"
$ docker logout ghcr.io
$ docker run --rm -it ${IMAGE_URI} /bin/bash -c 'echo ${CUDA_ARCHITECTURES}'
Unable to find image 'ghcr.io/rapidsai/velox-testing-images:presto-deps-cu13-amd64' locally
docker: Error response from daemon: error from registry: unauthorized
unauthorized
```

But you can pull it if authorized.

```console
$ gh auth refresh --scopes 'read:packages'
$ echo "$(gh auth token)" | docker login -u jameslamb --password-stdin ghcr.io
Login Succeeded

$ docker run --rm -it ${IMAGE_URI} /bin/bash -c 'echo ${CUDA_ARCHITECTURES}'
Unable to find image 'ghcr.io/rapidsai/velox-testing-images:presto-deps-cu13-amd64' locally
presto-deps-cu13-amd64: Pulling from rapidsai/velox-testing-images
44915bc1102a: Pull complete 
5dfbb821b924: Pull complete 
2c1ce468d9f3: Pull complete 
ffd87e050c02: Pull complete 
008112ead84b: Pull complete 
7c7eb956eccf: Download complete 
Digest: sha256:84227c09cf0bce90864fe2e37b4a3060ac9716816a10a428151c936e3fd17b2e
Status: Downloaded newer image for ghcr.io/rapidsai/velox-testing-images:presto-deps-cu13-amd64
"100;120"
```

<details><summary>screenshot (click me)</summary>

image

ref: https://github.com/orgs/rapidsai/packages/container/velox-testing-images/settings

</details>

### How I tested this

Tested the visibility and auth using the steps above.

Tested that the images could be pulled onto NVIDIA's self-hosted runners without any new configuration with a small test run:

https://github.com/rapidsai/velox-testing/actions/runs/21685556153/job/62531665903?pr=209

### DIFF
--- a/.github/workflows/ci-images.yml
+++ b/.github/workflows/ci-images.yml
@@ -1,8 +1,6 @@
 name: ci-images
 
 on:
-  # TODO: remove 'push' trigger before merging to main
-  push:
   schedule:
     - cron: "0 5 * * *"  # 5am UTC / 1am EST
   workflow_dispatch:


.github/workflows/ci-images.yml
```diff
@@ -0,0 +1,796 @@
...
+      matrix:
+        include:
+          - arch: amd64
+            cuda_version: '12.9'
+          - arch: arm64
+            cuda_version: '12.9'
+          - arch: amd64
+            cuda_version: '13.1'
+          - arch: arm64
+            cuda_version: '13.1'
```
jameslamb:
```suggestion
          - arch: amd64
            cuda_version: &latest_cu12 '12.9'
          - arch: arm64
            cuda_version: *latest_cu12
          - arch: amd64
            cuda_version: &latest_cu13 '13.1'
          - arch: arm64
            cuda_version: *latest_cu13
```

GitHub Actions now supports YAML anchors in workflow files (https://github.blog/changelog/2025-09-18-actions-yaml-anchors-and-non-public-workflow-templates/).

I think there's significant opportunity to use them to reduce duplication in this file, especially for CUDA versions and the list of supported GPU architectures.

Do what you want with that suggestion, consider this non-blocking.

.github/workflows/ci-images.yml
```diff
@@ -0,0 +1,796 @@
...
+      - name: Ensure sccache auth dir exists
+        run: |
+          if [[ "${{ steps.sccache.outputs.enabled }}" != "true" ]]; then
+            echo "::error::sccache setup failed. sccache is required for ci-images builds."
+            exit 1
+          fi
+
+      - name: Create .dockerignore
+        run: |
+          cat > .dockerignore <<'IGNORE'
```
jameslamb: Any reason not to just check this into source control in the repo?

That way, it could be used when testing builds locally too.

jameslamb (commented):
*(this looks like self-reviewing but really @bdice has carried this forward a LOT since the initial POC I put up)*

Overall looks good to me! I left some notes on a few areas for improvement but nothing major.

And I check https://github.com/rapidsai/velox-testing/pkgs/container/velox-testing-images, things seem to be working ok there and configured with the right permissions.

misiugodfrey (commented):
This looks great.

I agree with some of the comments that it may be nice to use anchors more aggressively to de-duplicate some code, but I don't think that should hold this back.

I do have a comment about whether we want the default for velox to be upstream or presto's pinned branch.  I'm in favour of the latter, but if we have already decided on upstream then so be it.

bdice:
All the observed failures are expected at this point, to the best of my knowledge. We have some known failures in the Presto-pinned Velox GPU tests.

I think we have a solid base to build from here, and this will allow us to move towards replacing our other CI workflows. I haven't yet tried to mirror the full Presto test suite from `presto-test-composite.yml`. I think that is best done in a follow-up PR. We can start revamping various parts of our infrastructure to use these images and refine them as we go.

Once this is approved, I will remove the temporary `push` triggers and other temporary/testing bits, and then merge.

Avinash-Raj:
@bdice This PR looks good to me, but it would be better if we address the above comments (avoid code duplication, enable sccache S3 preprocessor cache for Presto) before merging. Thanks.

presto/docker/native_build.dockerfile
```diff
@@ -38,11 +38,13 @@ ENV CC=/opt/rh/gcc-toolset-14/root/bin/gcc \
...
     SCCACHE_S3_NO_CREDENTIALS=false \
+    SCCACHE_S3_USE_SSL=true \
+    SCCACHE_DIRECT=true \
     SCCACHE_IDLE_TIMEOUT=0 \
     SCCACHE_DIST_AUTH_TYPE=token \
     SCCACHE_DIST_REQUEST_TIMEOUT=7140 \
     SCCACHE_DIST_SCHEDULER_URL="https://${TARGETARCH}.linux.sccache.rapids.nvidia.com" \
-    SCCACHE_DIST_MAX_RETRIES=4 \
+    SCCACHE_DIST_MAX_RETRIES=10 \
     SCCACHE_DIST_FALLBACK_TO_LOCAL_COMPILE=true
```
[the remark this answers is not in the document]
bdice: Fixed in a7815f5.

bdice:
@Avinash-Raj @misiugodfrey @simoneves Any further thoughts? The CI failures showing here are expected (known errors). I'd like to get an approval so this can be merged, so that we can begin to iterate.

Before merging, I'll need to disable a few CI triggers -- I want to leave them in place for right now in case we need to test again after a round of review. **Please don't merge until I've removed those triggers.**

PR #229, opened 2026-02-10
CI: Re-enable velox_cudf_s3_read_test

simoneves:
This test was disabled because of an RMM issue, since fixed, but it also did not run properly unless with `ctest` due to incorrect path resolution to a source file.


simoneves:
Landing now, as the fixed test IS now in upstream Velox `main`.

PR #239, opened 2026-02-16
Add benchmark configuration output to support posting to DB

misiugodfrey:
This PR consolidates benchmark result reporting, unifies engine/GPU detection across Docker and SLURM environments, and refactors shared test infrastructure into a common module.  This is for the purpose of providing an interface to automatically post benchmark results to the benchmarking DB via the post_results.py script.

Run context collection (run_context.py):
Automatically gathers engine type, GPU model, scale factor, worker count, and num_drivers at benchmark time and writes them into the "context" section of benchmark_result.json, which can then be parsed as part of post_results.py.  E.g.:

```
  "context": {
    "iterations_count": 5,
    "schema_name": "sf1",
    "scale_factor": 1,
    "n_workers": 1,
    "kind": "single-node",
    "engine": "java",
    "gpu_count": 0,
    "gpu_name": "NA",
    "num_drivers": 32,
    "execution_number": 1,
    "timestamp": "2026-03-04T23:51:21Z",
    "benchmark": "tpch"
  },
```

Unified engine and GPU detection:
It was necessary to gather information about a running cluster to put into the benchmark context.  To that end some changes were necessary to provide the appropriate information at runtime (rather than trying to predict context from things like image names) as well as to unify how we present this information across the docker and slurm scripts.
Engine detection via cluster-tag: All variants (GPU, CPU, Java) now consistently set cluster-tag in the coordinator config. run_context.py queries the Presto /v1/cluster API to determine the engine type, replacing the previous approach of parsing Docker image names or SLURM logs.
GPU model detection via worker logs: Docker worker containers now run nvidia-smi -L at startup and write output to worker_logs/worker_<id>.log, matching the existing SLURM behavior. run_context.py reads these logs uniformly regardless of the deployment environment.
launch_presto_servers.sh rewritten to capture nvidia-smi output and redirect presto_server stdout/stderr to per-worker log files. Mounted as a read-only volume so changes don't require image rebuilds.
worker_logs volume mount added to all worker services (GPU, CPU, Java) in Docker Compose files.


presto/scripts/run_benchmark.sh
```diff
@@ -34,13 +34,22 @@ OPTIONS:
     -m, --metrics           Collect detailed metrics from Presto REST API after each query.
                             Metrics are stored in query-specific directories.
 
+ENVIRONMENT:
```
paul-aiyedun: Can we use command line arguments instead of environment variables?

presto/scripts/run_benchmark.sh
```diff
@@ -34,13 +34,22 @@ OPTIONS:
...
+    PRESTO_BENCHMARK_DEBUG   Set to 1 to print debug logs for worker/engine detection
+                             (e.g. node URIs, reachability, metrics, Docker containers).
+                             Use when engine is misdetected or the run fails.
+    Docker                  In Docker setups, engine is inferred from running worker
```
paul-aiyedun: Can you please expand on this?

presto/testing/performance_benchmarks/conftest.py
```diff
@@ -109,6 +111,28 @@ def write_section(terminalreporter, text_report, content, **kwargs):
     text_report.append(f" {content} ".center(120, sep))
 
 
+def _build_run_config(session):
```
paul-aiyedun: Nit: Remove underscore prefix.

presto/testing/performance_benchmarks/run_context.py
```diff
@@ -0,0 +1,295 @@
...
+            return None
+        first_line = result.stdout.strip().split("\n")[0]
+        return _parse_gpu_name_from_text(first_line)
+    except (FileNotFoundError, subprocess.TimeoutExpired, Exception):
+        return None
+
+
+def get_worker_image() -> str | None:
+    """Return worker image name from env (set by cluster/container setup)."""
+    return os.environ.get("WORKER_IMAGE")
```
paul-aiyedun: Where is this set?

presto/testing/performance_benchmarks/run_context.py
```diff
@@ -0,0 +1,295 @@
...
+    if match:
+        return match.group(1).strip()
+    if line.startswith("GPU "):
+        return line.split(":", 1)[-1].strip()
+    return None
+
+
+def get_gpu_name_from_slurm_logs() -> str | None:
+    """
+    When running under SLURM, workers run nvidia-smi -L and write to LOGS/worker_<id>.log.
```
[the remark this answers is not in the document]
misiugodfrey: The current approach get the gpu details by running `nvidia-smi` in both contexts, it's just that in the  docker context we run it on the host, whereas in the slurm context, the login-nodes don't have `nvidia-smi` installed, so you need to run it on the worker node (i.e. in the worker's container).

We could potentially try to consolidate this, but it's more an issue of how the slurm clusters' encapsulate things that required extra steps in that context.

presto/testing/performance_benchmarks/run_context.py
misiugodfrey: This is only set in the slurm scripts.  I'm removing this section since it doesn't make much sense to specify a WORKER_IMAGE name since we are detecting the engine/variant through other means.

presto/testing/performance_benchmarks/run_context.py
```diff
@@ -0,0 +1,295 @@
...
+    except (FileNotFoundError, subprocess.TimeoutExpired, Exception):
+        return None
+
+
+def get_worker_image() -> str | None:
+    """Return worker image name from env (set by cluster/container setup)."""
+    return os.environ.get("WORKER_IMAGE")
+
+
+def _current_username() -> str:
```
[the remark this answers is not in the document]
misiugodfrey: I've changed this to determine the engine/variant through the presto API, so this is all removed now.

presto/docker/docker-compose/template/docker-compose.native-gpu.yml.jinja
```diff
@@ -49,6 +49,8 @@ x-presto-native-worker-gpu: &gpu_worker_base
     - ../../config/generated/gpu/etc_worker/node.properties:/opt/presto-server/etc/node.properties
     - ../../config/generated/gpu/etc_worker/config_native.properties:/opt/presto-server/etc/config.properties
     - ../../config/generated/gpu/etc_worker/catalog/hive.properties:/opt/presto-server/etc/catalog/hive.properties
+    - ../../worker_logs:/opt/presto-server/logs
+    - ../../launch_presto_servers.sh:/opt/launch_presto_servers.sh:ro
```
misiugodfrey: Mounting this file was necessary so that older images can still benefit from the new version.

presto/docker/docker-compose/template/docker-compose.native-gpu.yml.jinja
```diff
@@ -82,9 +85,12 @@ services:
       KVIKIO_NTHREADS: {{ kvikio_threads }}
       CUDA_VISIBLE_DEVICES: {{ gpu_id }}
     volumes:
+      - ../../config/generated/gpu/etc_common:/opt/presto-server/etc
```
paul-aiyedun: Why is the update needed?

presto/testing/performance_benchmarks/run_context.py
```diff
@@ -0,0 +1,240 @@
...
+        return None
+    finally:
+        if conn is not None:
+            try:
+                conn.close()
+            except Exception:
+                pass
+
+
+def _parse_gpu_name_from_text(line: str) -> str | None:
```
paul-aiyedun: Can we avoid this by directly printing the GPU name i.e. `nvidia-smi --query-gpu=name --format=csv,noheader`?

presto/testing/performance_benchmarks/run_context.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+_ENGINE_TO_VARIANT = {
+    "presto-velox-gpu": "gpu",
+    "presto-velox-cpu": "cpu",
+    "presto-java": "java",
+}
+
+
+def _get_num_drivers(engine: str) -> int | None:
```
paul-aiyedun: Can we also get this from the logs?

presto/testing/performance_benchmarks/run_context.py
```diff
@@ -0,0 +1,240 @@
...
+    sf = _get_scale_factor_from_schema(hostname, port, user, schema_name)
+    if sf is not None:
+        ctx["scale_factor"] = int(sf) if isinstance(sf, float) and sf == int(sf) else sf
+
+    n_workers = _get_node_count(hostname, port)
+    engine = _get_engine(hostname, port)
+
+    if n_workers is not None:
+        ctx["n_workers"] = n_workers
+        ctx["kind"] = "single-node" if n_workers == 1 else f"{n_workers}-node"
```
paul-aiyedun: Out of curiosity, what is the purpose of the `kind` field?

PR #126, opened 2025-11-13
Scripts to run presto in a Slurm cluster

misiugodfrey:
Creating a set of scripts to automate running presto (gpu-native) within a Slurm cluster.  Currently should support single-node-single-gpu and single-node-multi-gpu (multiple workers on the same node with different GPU assignments).  It's currently being used to run TPCH-SF1k

TODOs:
Refactor this to use more of the velox-testing scripts internally (such as run_benchmarks.sh).
Create a common config setup

PR #240, opened 2026-02-16
Misiug/slurmprofiles

misiugodfrey:
Set up a system to record profiles in a slurm cluster environment.

When the profiling option is used for `run_benchmark.sh` (-p|--profile), each worker container will run nsys start, wrapping the presto executable.  When `run_benchmark.sh` is finished running, it will send a signal to each worker by writing to a file based on that worker_id.  Each worker will  periodically check it's respective file and will run nsys stop when the signal is received.

This means that we currently profile the entire query suite, rather than individual queries - and all profiles are only for one worker.  If we want to profile an individual query, it is best to run the `run_benchmark.sh` with only that query.

This PR is also based on the as-of-yet unmerged slurmscripts PR: https://github.com/rapidsai/velox-testing/pull/202 so it will need to wait for that to merge first.


karthikeyann:
@misiugodfrey This PR is currently in draft mode. Is it superseded by any of the other PRs? https://github.com/rapidsai/velox-testing/pull/382 ?

There are also a few old draft PRs: #126, #206, #273, #391, #392, #393, and #394. Could you please review these PRs, close any that are no longer required, and refresh/update the ones that are still needed?

PR #243, opened 2026-02-18
Add sccache support for Presto native builds

Avinash-Raj:
Add sccache support for Presto native builds 

Integrate sccache build caching into the Presto native build pipeline,
mirroring the existing Velox sccache integration. This enables remote S3
cache sharing across machines and significantly reduces rebuild times.

**Changes:**
- Add `--sccache`, `--sccache-version`, and `--sccache-enable-dist` CLI flags
  to start_presto_helper_parse_args.sh
- Update `start_presto_helper.sh` with sccache auth validation and
  conditional compose file / build arg selection
- Modify `native_build.dockerfile` to conditionally install sccache,
  configure S3 caching, and set CMake compiler launchers
- Add `docker-compose.native-cpu.sccache.yml` overlay for CPU builds
  with secrets and ulimits
- Update GPU Jinja template with conditional sccache secrets/ulimits
  for multi-worker support

Just run the startup script with the `--sccache` option, like this: `./start_native_gpu_presto.sh --sccache`. This builds and starts Presto with GPU support, using sccache to cache C++ compilation results in a shared S3 bucket. On cache hits, precompiled objects are downloaded instead of being recompiled, significantly speeding up builds across machines and repeated runs.


mattgara (commented):
Testing this I see a drop from `900s` -> `100s`, looks good! 👍 

Also, upon a second repeated build: 
```
#9 143.0 Cache hits rate                     100.00 %
#9 143.0 Cache hits rate (C/C++)             100.00 %
#9 143.0 Cache hits rate (CUDA)              100.00 %
```

Overall LGTM, can you address the comment I made about warnings? Thanks!

PR #247, opened 2026-02-23
Add option to select custom queries for tpch and tpcds benchmarks

devavret:
This PR adds an option to provide a path to a custom json file containing TPC-H and TPC-DS queries. 
Also adds queries_duckdb.json which is a file containing queries converted to SQL from duckdb plans.
Adds queries.best.json which is a clone of queries.json with SQL for Q17 and Q21 replaced with duckdb plan generated queries. This is because Q17 and Q21 showed notable improvement in performance while other queries had the same performance with standard SQL vs duckDB converted SQL.


presto/testing/common/queries/tpch/queries_best.json
paul-aiyedun: I believe adding multiple query files in the codebase could be a source of confusion, as it is no longer clear what the authoritative source of queries is. Also, can the benchmarks still be considered standard if we are running queries that are different from those in the TPC-H/TPC-DS specification? Ideally, query rewrites should be done by the query engine we are publishing benchmarks for and not as a preprocessing step.

karthikeyann: The original 22 queries are part of the official TPC-H specification.
May be consider renaming the query file name. (`queries_rapidsmpf.json` or better).
We are testing new SQL queries for TPCH that give us insights to better plans. So this SQL should be saved.

mattgara (approved):
LGTM.

> The source polars plans are found here 
https://github.com/pola-rs/polars-benchmark/tree/8975df5acb720c520dfdeb888d3668ba151f41df/queries/polars
The file queries_polars.json contains the translated queries.

Out of curiosity, does rewriting SQL translate to a different plan always being used? Is this typically verified manually?

PR #251, opened 2026-02-24
Add support for Spark Gluten

paul-aiyedun:
* Testing and benchmarking infrastructure for Spark with Gluten.

* Integration tests for TPC-H and TPC-DS comparing results against DuckDB or a pre-existing reference result.

* Performance benchmarks with configurable iterations, statistical aggregation, and automatic cache dropping.

* Refactored common testing utilities to `common/testing/` (test data, pytest fixtures, integration test utilities, benchmark infrastructure).

* Presto tests updated to use shared utilities while maintaining backward compatibility.

* Removed GPU requirement from cache dropping utility.

* Add custom dataset name support.

* Updated README with Spark Gluten quick start, testing, and benchmarking guides.

* Scripts: `download_gluten.sh`, `run_integ_test.sh`, `run_benchmark.sh` with comprehensive CLI options.


common/testing/performance_benchmarks/cache_utils.py
```diff
@@ -16,8 +16,6 @@ def drop_cache():
         "run",
         "--rm",
         "--privileged",
-        "--gpus",
```
mattgara: Just checking, was the removal of `gpus all` here intended? I don't believe it will do any harm, so likely not a big deal either way.

PR #265, opened 2026-03-11
Add explicit page cache control (hot, cold) to presto-velox benchmarks

kingcrimsontianyu:
This PR allows presto-velox testing to drop the page cache on the file scope without requiring sudo privilege. This PR also supports cold cache status where cache dropping is performed before each iteration of each query.

### DIFF
--- a/presto/testing/performance_benchmarks/common_fixtures.py
+++ b/presto/testing/performance_benchmarks/common_fixtures.py
@@ -1,7 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
 # SPDX-License-Identifier: Apache-2.0
 
-import os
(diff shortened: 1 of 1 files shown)


common/testing/performance_benchmarks/cache_utils.py
```diff
@@ -19,11 +109,66 @@ def drop_cache():
...
+        for file in expanded_path.rglob("*"):
+            if file.is_file():
+                _drop_file_cache(file)
+        return
+
+    raise ValueError(f"Path is neither a regular file nor a directory: {expanded_path}")
+
+
+def cache_setup_per_iteration(cache_mode, data_dir):
+    if cache_mode == "cold":
```
paul-aiyedun: Do we need to do more than one iteration for cold run?

PR #274, opened 2026-03-14
Update CI images workflow to build and push Presto coordinator

Avinash-Raj:


.github/workflows/test.yml
```diff
@@ -0,0 +1,391 @@
...
+    name: integ-test (${{ matrix.test_type }}, presto-gpu-cuda${{ matrix.cuda_version }}, ${{ matrix.arch }})
+    needs: [resolve-inputs]
+    permissions:
+      contents: read
+      packages: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: amd64
```
bdice: Can we have ARM coverage here?

.github/workflows/test.yml
```diff
@@ -0,0 +1,391 @@
...
+        run: ./generate_presto_config.sh
+
+      - name: Patch coordinator configs for container execution
+        run: |
+          CONFIG_DIR="${GITHUB_WORKSPACE}/presto/docker/config/generated/gpu"
+
+          sed -i 's/-Xmx[0-9]*G/-Xmx8G/;s/-Xms[0-9]*G/-Xms8G/' \
+            "${CONFIG_DIR}/etc_common/jvm.config"
+
+          sed -i \
```
bdice: Is it possible for us to commit all these options as a file and use that? I would like to avoid extensive use of sed to edit file contents, and would like to keep the weight of the bash commands in this workflow file to a minimum. Eventually it would be best to have it all in shell scripts that we call, rather than inserting inline scripts into the YAML.

.github/workflows/test.yml
Avinash-Raj: Only for integration tests, or for all the GPU tests?

PR #275, opened 2026-03-14
Add validation to run_benchmark.sh

misiugodfrey:
## Summary

Adds automatic result validation to the Presto benchmark pipeline and consolidates the comparison logic into a shared module used by both the benchmark and integration test paths.

### What's added

- **`run_benchmark.sh` validates after every run.** Compares the benchmark's output parquet files against a directory of reference parquet files and writes `validation_results.json` next to `benchmark_result.json`.
- **`common/testing/validate_results.py`** — standalone CLI that performs the comparison (also invokable directly).
- **`common/testing/result_comparison.py`** — the shared comparison pipeline. Reused by `presto/testing/integration_tests/` so that both paths use identical semantics.
- **`benchmark_reporting_tools/post_results.py`** now reads `validation_results.json` alongside `benchmark_result.json` when uploading to the database.

### Reference data resolution (in order of precedence)

1. `--reference-results-dir <path>` — explicit; missing directory is a fatal error.
2. `PRESTO_EXPECTED_RESULTS_DIR` env var — implicit; missing directory is a warning, validation skipped.
3. Otherwise, validation is skipped and overall status is `not-validated`.

### Comparison pipeline (`result_comparison.py`)

For each query (positional, column-label-agnostic):

1. **Column count** check.
2. **Type normalization** — coerce `actual`'s dtypes and value types to match `expected`. Handles the cross-engine date mismatch (Presto returns DATE as `str`; DuckDB as `datetime.date`/`datetime64`) and pandas 3.0 pyarrow-backed string dtype quirks.
3. **Row count** check.
4. **ORDER BY parse** — extract column indices and ASC/DESC directions from the SQL via sqlglot.
5. **Per-frame ORDER BY validation** — verify each engine respected ORDER BY: each column is monotonic in its direction within tie groups of the prior ORDER BY columns, using strict equality (values come from one engine, bit-identical to themselves).
6. **LIMIT tie-boundary handling** — when LIMIT is present, peel the tolerance-tied tail (in engine order) from both frames. Compare the front fully (after canonical sort); compare only the tail's ORDER BY values, since engines may select different rows from a tied bucket at the cutoff.
7. **Canonical sort + compare** — sort by `[non-float columns, float columns]` so engine ordering does not affect the result, then compare positionally with float tolerance (`rel_tol=1e-5`, `abs_tol=1e-8`) and `pd.isna`-based null handling.

Validation status priority: `failed > not-validated > xfail > passed`.

### Other changes

- Drop `-f/--scale-factor` from `run_benchmark.sh` (auto-detected from the catalog).
- Fix Q11 SF_FRACTION precision (`.10f` → `.12f`) so the HAVING threshold agrees with DuckDB's reference image at SF≥1000.
- Move shared queries JSON files and integration test helpers under `common/testing/`.
- Validation output is prefixed `[VALIDATION]`; passing queries are silent, only XFAIL/FAIL details and the summary line are printed.
- When `-q` is used with `run_benchmark.sh`, only those queries are validated.

## Test plan

- [x] `python -m pytest common/testing/result_comparison_test.py` — 16 unit tests pass.
- [x] End-to-end TPC-H at SF=1 against the SF=1 expected directory — all 22 queries validate as passed.
- [x] SF=3000 precision stress (Q11 row-count + tie-boundary) — passes with the `.12f` SF_FRACTION fix and tolerance-aware tie detection.
- [x] Run with no reference dir → overall status `not-validated`.
- [x] Run with explicit `--reference-results-dir /nonexistent` → fatal error before benchmark starts.
- [x] Run with `PRESTO_EXPECTED_RESULTS_DIR=/nonexistent` → warning, validation skipped.


benchmark_reporting_tools/post_results.py
```diff
@@ -616,6 +644,18 @@ async def _process_benchmark_dir(
             print("  Warning: no config directory found. Use --config-dir to specify one.", file=sys.stderr)
         engine_config = None
 
+    validation_results_path = benchmark_dir / "validation_results.json"
+    if validation_results_path.exists():
+        print("  Loading validation results...", file=sys.stderr)
+        try:
+            validation_results = json.loads(validation_results_path.read_text())
+        except (json.JSONDecodeError, FileNotFoundError) as e:
+            print(f"  Warning: could not load validation results: {e}", file=sys.stderr)
```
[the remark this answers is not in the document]
misiugodfrey: I'm now popagating the exception and exiting if the json is malformed.

TomAugspurger (approved):
Looks good. Just one note about a recent change to `assert_tpch_result_equal` in cudf-polars.

paul-aiyedun (changes requested):

mattgara (approved):
LGTM. Awesome been waiting for a Polars verification utility, will really help with validation of larger runs!

benchmark_reporting_tools/validate_results.py
```diff
@@ -0,0 +1,496 @@
...
+# /// script
+# requires-python = ">=3.9"
+# dependencies = [
+#     "polars",
+# ]
+# ///
+"""
+Validate TPC-H query results against expected parquet files.
+
+Validation logic is ported from cudf_polars's assert_tpch_result_equal
```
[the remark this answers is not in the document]
paul-aiyedun: The validation logic should be semantically the same in both projects, so what is the reason for the change? Using the same module for validation across projects would be ideal, but I don't think code duplication is the best way to get there.

misiugodfrey:
I've updated this PR to deduplicate the code between the Polars validation and what already existed in integration tests.

Current output looks like:
Passing case:
```
[VALIDATION] Running validation: /velox-testing/presto/scripts/benchmark_output/query_results vs /Data/sf1_expected
[VALIDATION] Benchmark:  tpch
[VALIDATION] Results: 22 passed, 0 failed, 0 expected-failure, 0 skipped
[VALIDATION] Results written to /velox-testing/presto/scripts/benchmark_output/validation_results.json
```
Case with failures:
```
[VALIDATION] Running validation: /velox-testing/presto/scripts/benchmark_output/query_results vs /Data/sf1_expected_bad
[VALIDATION] Benchmark:  tpch
[VALIDATION] Q1  : FAIL     AssertionError: Found 4 mismatch(es):
  Row 0, col 'sum_base_price': 56586554400.729935 vs 0.0 (diff=5.66e+10, tol=5.66e+05)
  Row 1, col 'sum_base_price': 1487504710.380001 vs 0.0 (diff=1.49e+09, tol=1.49e+04)
  Row 2, col 'sum_base_price': 111701729697.73993 vs 0.0 (diff=1.12e+11, tol=1.12e+06)
  Row 3, col 'sum_base_price': 56568041380.89992 vs 0.0 (diff=5.66e+10, tol=5.66e+05)
[VALIDATION] Q14 : FAIL     AssertionError: Found 1 mismatch(es):
  Row 0, col 'promo_revenue': 16.38077862639554 vs 32.76155725279108 (diff=1.64e+01, tol=3.28e-04)
[VALIDATION] Q7  : FAIL     AssertionError: Row count mismatch: 4 (actual) vs 5 (expected)
[VALIDATION] Results: 19 passed, 3 failed, 0 expected-failure, 0 skipped
[VALIDATION] Results written to /velox-testing/presto/scripts/benchmark_output/validation_results.json
```

TomAugspurger:
@misiugodfrey can you  walk me through what those two outputs are saying? IIUC, it's that with the previous code there's some tests somewhere (where?) running something and validating the results, and all the tests passed validation. But with the new validation code, some of those tests are failing with validation errors? Based on the traceback, it looks like the values are different, so it's a good thing there's a validation error there (but the tests or implementation will need to be updated?).

Or am I misreading things?

misiugodfrey:
> @misiugodfrey can you walk me through what those two outputs are saying? 

@TomAugspurger I should clarify, the two cases I posted above were for the sake of showing the new output format.  The "failing" case was a set of parquet files I intentionally changed to fail to test that path.  So far everything seems to validate correctly (tested up to 3k).

presto/scripts/run_benchmark.sh
```diff
@@ -292,3 +307,33 @@ echo "Using PRESTO_IMAGE_TAG: $PRESTO_IMAGE_TAG"
...
+fi
+
+RESULTS_DIR="${ACTUAL_OUTPUT_DIR}/query_results"
+VALIDATE_SCRIPT="${SCRIPT_DIR}/../../benchmark_reporting_tools/validate_results.py"
+
+# Resolve reference results directory.
+# PRESTO_EXPECTED_RESULTS_DIR env var is the implicit fallback (warning if missing).
+# Explicit --reference-results-dir was already validated before the benchmark ran.
+if [[ -n ${PRESTO_EXPECTED_RESULTS_DIR} && ! -d ${PRESTO_EXPECTED_RESULTS_DIR} ]]; then
+  echo "[Validation] Warning: PRESTO_EXPECTED_RESULTS_DIR not found: ${PRESTO_EXPECTED_RESULTS_DIR}; validation skipped."
```
paul-aiyedun, line 318: Unless I am missing something, we still seem to proceed with validation in this case. Should we exit after this line?

presto/scripts/run_benchmark.sh
```diff
@@ -292,3 +307,33 @@ echo "Using PRESTO_IMAGE_TAG: $PRESTO_IMAGE_TAG"
 
 BENCHMARK_TEST_DIR=${TEST_DIR}/performance_benchmarks
 pytest -q -s ${BENCHMARK_TEST_DIR}/${BENCHMARK_TYPE}_test.py ${PYTEST_ARGS[*]}
+
+# Compute the actual output directory (mirrors pytest's --output-dir / --tag logic).
+ACTUAL_OUTPUT_DIR="${OUTPUT_DIR:-$(pwd)/benchmark_output}"
+if [[ -n ${TAG} ]]; then
+  ACTUAL_OUTPUT_DIR="${ACTUAL_OUTPUT_DIR}/${TAG}"
+fi
```
paul-aiyedun: Consider moving this logic into the `validate_results.py` script and reusing the same function that sets this (https://github.com/rapidsai/velox-testing/blob/0f2e363f09e9be86519f4839a2cbc4089e3bb513/common/testing/performance_benchmarks/conftest.py#L184).

presto/scripts/run_benchmark.sh
```diff
@@ -292,3 +307,33 @@ echo "Using PRESTO_IMAGE_TAG: $PRESTO_IMAGE_TAG"
...
+  VALIDATE_ARGS+=(--reference-results-dir "${PRESTO_EXPECTED_RESULTS_DIR}")
+fi
+if [[ -n ${QUERIES} ]]; then
+  VALIDATE_ARGS+=(--queries "${QUERIES}")
+fi
+
+VALIDATE_REQUIREMENTS="${SCRIPT_DIR}/../../benchmark_reporting_tools/requirements.txt"
+echo "[Validation] Running validation: ${RESULTS_DIR} vs ${PRESTO_EXPECTED_RESULTS_DIR:-<not set>}"
+pip install -q -r "${VALIDATE_REQUIREMENTS}"
+python "${VALIDATE_SCRIPT}" "${VALIDATE_ARGS[@]}"
```
paul-aiyedun: Did you consider using `run_py_script.sh`?

benchmark_reporting_tools/post_results.py
```diff
@@ -396,10 +397,22 @@ def _query_sort_key(name: str):
 
     query_names = sorted(raw_times.keys(), key=_query_sort_key)
 
+    per_query_validation = (validation_results or {}).get("queries", {})
+
+    def _get_validation_result(query_name):
+        # Look up validation result for this query (keys are lowercase e.g. "q1")
+        vkey = "q" + query_name.lstrip("Q").lower()
+        vdata = per_query_validation.get(vkey)
```
paul-aiyedun, line 405: What happens if `per_query_validation` is an empty dictionary (from line 400)?

benchmark_reporting_tools/post_results.py
```diff
@@ -680,6 +707,15 @@ async def _process_benchmark_dir(
         print(f"  Identifier hash: {payload['query_engine']['identifier_hash']}", file=sys.stderr)
         print(f"  Node count: {payload['node_count']}", file=sys.stderr)
         print(f"  Query logs: {len(payload['query_logs'])}", file=sys.stderr)
+        print(f"  Validation status: {payload['validation_status']}", file=sys.stderr)
+        xfail_queries = [
```
paul-aiyedun, line 711: What does the `x` prefix mean here?

benchmark_reporting_tools/validate_results.py
```diff
@@ -0,0 +1,226 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# /// script
+# requires-python = ">=3.9"
+# dependencies = [
+#     "pandas",
+#     "pyarrow",
+#     "sqlglot",
+# ]
+# ///
```
paul-aiyedun: Nit: Remove this comment block? I don't think we need to keep track of the dependencies in the python file.

common/testing/validate_results.py
paul-aiyedun, line 1: This file should probably be in `common/testing`.

common/testing/validate_results.py
```diff
@@ -0,0 +1,226 @@
...
+
+    Returns a dict with keys:
+      overall_status: "passed" | "failed" | "expected-failure" | "not-validated"
+      queries: { "q1": {"status": ..., "message": ...}, ... }
+    """
+    query_results: dict[str, dict] = {}
+    passed = failed = not_validated = expected_failures = 0
+
+    if query_numbers is not None:
+        result_files = sorted(f for q in query_numbers for f in [results_dir / f"q{q}.parquet"] if f.exists())
```
paul-aiyedun, line 53: Should there be a failure if the file does not exist?

benchmark_reporting_tools/validate_results.py
```diff
@@ -0,0 +1,226 @@
...
+        sys.exit(1)
+
+    # Load canonical SQL queries for the benchmark type
+    queries = get_queries(args.benchmark_type)
+
+    query_numbers = [int(q.strip()) for q in args.queries.split(",")] if args.queries else None
+
+    print(f"[Validation] Benchmark:  {args.benchmark_type}")
+    if query_numbers is not None:
+        print(f"[Validation] Queries:    {query_numbers}")
```
paul-aiyedun: Nit: I don't think we need to print these.

common/testing/validate_results.py
```diff
@@ -0,0 +1,226 @@
...
+    return parser.parse_args()
+
+
+def _write_not_validated(results_dir: Path, reason: str) -> None:
+    """Write a not-validated sentinel JSON and print the reason."""
+    print(f"[Validation] {reason}")
+    results = {"overall_status": "not-validated", "queries": {}}
+    output_path = results_dir.parent / "validation_results.json"
+    output_path.write_text(json.dumps(results, indent=2))
+    print(f"[Validation] Results written to {output_path}")
```
paul-aiyedun, line 188: Why do we write to `validation_results.json` in this case?

common/testing/validate_results.py
```diff
@@ -0,0 +1,226 @@
...
+        queries:        Dict mapping query IDs (e.g. "Q1") to SQL strings.
+        query_numbers:  Optional list of query numbers to validate.  When provided,
+                        only those queries are checked.
+
+    Returns a dict with keys:
+      overall_status: "passed" | "failed" | "expected-failure" | "not-validated"
+      queries: { "q1": {"status": ..., "message": ...}, ... }
+    """
+    query_results: dict[str, dict] = {}
+    passed = failed = not_validated = expected_failures = 0
```
paul-aiyedun, line 50: Why do we have expected failures in this case? Also, can we mark the benchmark run as not validated if any of the checks set the `not_validated` variable?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,489 @@
...
+DuckDB reference) and benchmark validation (comparing result parquet files
+against expected parquet files).
+
+Comparison behaviour
+--------------------
+- Column names are validated; Presto _colN anonymous aggregate columns are
+  renamed to match the expected column names positionally.
+- Decimal columns are cast to float64; narrow integers (int8/int16/int32) are
+  widened to int64; string columns are cast to temporal when the expected
+  column is temporal (Presto may write dates as strings).
```
paul-aiyedun: Why are these casts needed?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,489 @@
...
+- Floating-point values are compared with rel_tol=1e-5, abs_tol=1e-8.
```
paul-aiyedun: What is the reason for doing both relative and absolute comparisons?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,489 @@
...
+- ORDER BY is extracted dynamically from the SQL using sqlglot:
+    - Validates that both frames are sorted by the ORDER BY columns.
+    - Tie-breaking is resolved by sorting on all non-float columns.
+- For queries with ORDER BY + LIMIT, rows at the limit boundary are compared
```
paul-aiyedun: Can we compare the other columns in this case and log a warning if they don't match? With this approach, it is possible to have an edge case where wrong results are not detected.

common/testing/result_comparison.py
```diff
@@ -0,0 +1,489 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Shared result comparison logic for TPC-H/TPC-DS query validation.
+
+Used by both integration tests (comparing live query engine results against a
+DuckDB reference) and benchmark validation (comparing result parquet files
+against expected parquet files).
+
+Comparison behaviour
```
paul-aiyedun: Nit: Is this a repetition of the comments in `compare_result_frames`?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,489 @@
...
+    # 1 & 2. Column reconciliation and validation
+    actual = _reconcile_col_names(actual, expected)
+    if list(actual.columns) != list(expected.columns):
+        extra = set(actual.columns) - set(expected.columns)
+        missing = set(expected.columns) - set(actual.columns)
+        raise AssertionError(
+            f"Column name mismatch — extra: {extra}, missing: {missing}\n"
+            f"  actual:   {list(actual.columns)}\n"
+            f"  expected: {list(expected.columns)}"
+        )
```
paul-aiyedun, line 415: Why are we concerned about column names for this validation?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,489 @@
...
+    expected = expected.copy()
+
+    for col in actual.columns:
+        if col not in expected.columns:
+            continue
+
+        # Decimal → float64
+        if _is_decimal_like(actual[col]):
+            actual[col] = pd.to_numeric(actual[col], errors="coerce")
+        if _is_decimal_like(expected[col]):
```
paul-aiyedun: Instead of repeated if statements, can we have this function do normalization for one dataframe at a time (similar to the `normalize_rows` function that existed before)?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,489 @@
...
+                    mismatches.append(f"Row {row_idx}, col '{col}': {fv1} vs {fv2} (diff={diff:.2e}, tol={tol:.2e})")
+            elif v1 != v2:
+                # Safety net: compare as Timestamps when one side is a date
+                # string (e.g. Presto '1995-03-05') and the other is a
+                # Timestamp object that slipped through _normalize_dtypes.
+                try:
+                    if pd.Timestamp(v1) == pd.Timestamp(v2):
+                        continue
+                except Exception:
+                    pass
```
paul-aiyedun, line 326: Why is the column type not checked here?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,489 @@
...
+        raise AssertionError(f"Row count mismatch: {len(actual)} (actual) vs {len(expected)} (expected)")
+
+    # 5. Parse ORDER BY / LIMIT
+    sort_by, nulls_last = get_orderby_info(query_sql, list(actual.columns))
+    limit = get_limit(query_sql)
+
+    if not sort_by:
+        # No ORDER BY (or unparsable) — sort both sides and compare
+        _assert_frames_equal(_sort_for_comparison(actual), _sort_for_comparison(expected))
+        return
```
paul-aiyedun, line 433: Is this check sufficient for queries with limits?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,489 @@
...
+# SQL parsing helpers
+# ---------------------------------------------------------------------------
+
+
+def get_orderby_info(query_sql: str, column_names: list[str]) -> tuple[list[tuple[str, bool]], bool]:
+    """
+    Extract ORDER BY information from SQL using sqlglot.
+
+    Returns:
+        (sort_by, nulls_last) where sort_by is [(col_name, descending), ...]
```
paul-aiyedun, line 42: Why are we concerned about `nulls_last`?

paul-aiyedun (commented):
Changes overall make sense to me. However, I had a number of questions.

presto/scripts/run_benchmark.sh
misiugodfrey: We proceeded to the next step where the script would output more details about how the directory does not exist and that validation was skipped.  I think you are right though that an early exit is a better idea here, since the extra output is effectively redundant since we are already stating we are skipping due to a missing directory.  I've changed this to an early exit.

presto/scripts/run_benchmark.sh
misiugodfrey: I've refactored this to use the same logic.

presto/scripts/run_benchmark.sh
misiugodfrey: Switching to use the script.

benchmark_reporting_tools/post_results.py
misiugodfrey: If `per_query_validation` is empty then `per_query_validation.get(vkey)` should return `None` for each query and the status will be returned as "not-validated".

benchmark_reporting_tools/post_results.py
misiugodfrey: The "x" prefix was short for "expected".  As in this is an "expected failure".  This convention was set in the Benchmarking DB where one of the possible validation states is "XFAIL" for this particular case.

benchmark_reporting_tools/validate_results.py
misiugodfrey: Removed.

common/testing/validate_results.py
misiugodfrey: Moved.

common/testing/validate_results.py
misiugodfrey: I could see that if you specified an expected directory, then you probably expect it to contain validation for every query.  I'll change this to fail on a missing file.

common/testing/validate_results.py
misiugodfrey: That's because `post_results.py` reads `validation_restuls.json` to tell what level of validation was done on a benchmark.  I think it's most clear if we make sure to write the `not-validated` case as well to be explicit what the state should be.

common/testing/validate_results.py
misiugodfrey: The expected failure is currently only for Q15 of tpch, which has an expected failure state because of the way floats are evaluated in the query.  Once that's fixed we can remove any expected failures.

Right now the overall status of the run is marked as `not_validated` if any of the queries are not validated.

common/testing/result_comparison.py
misiugodfrey: We could log a warning in this case... but I would point out that having non-matching non-sort columns here would be valid behavior.  For larger scale factors it is not uncommon to have multiple ties in the sort columns, enough that it crosses the boundary of a LIMIT clause.

For instance, you could have 15 ties for the sort columns in a query with LIMIT 10, where all the non-sort columns have unique values.  This means any subset of 10 of those 15 unique values would be a valid result for the non-sort columns.  We could issue a warning if we don't match the expected case, but it's still a perfectly valid result - so I'm not sure a warning is informative (it would too often cause alarm for valid cases, or be ignored if there really was a problem).

If we are concerned about potential incorrectness for these tie cases, then I think we should revise our validation system to include _all_ the the potentially correct outcomes.  One potential solution to that would be alter the sql we run to generate the "Expected" results to extend the LIMIT value so that we never end on a tie (it would include all tied values).  In the above case we would extend the LIMIT to 15 so that we always report 15/15 possible rows, and then when we run the real query the validation code could look to make sure our 10 results are valid subsets of the 15 expected.

If we did go in this direction, I think that's still future work.  It think we should submit this as is and then alter it in conjunction with the expected value generation if that's a direction we want to pursue.

common/testing/result_comparison.py
misiugodfrey: Absolute tolerance is permissive for large values (two large values differing by 1.0 would pass), whereas relative tolerance would be too strict for values near zero (1e-10 vs 2e-10 differ by %100 relatively, but are both effectively zero).  This logic was ported in from the Polars validation, which used both for this reason (it's the same logic as python's math.isclose() uses)

common/testing/result_comparison.py
misiugodfrey: Sure.

common/testing/result_comparison.py
misiugodfrey: This is because the Polars validation did this.  If we are unconcerned with the column names I could remove this, but I would rather we keep as close to them as we can.

common/testing/result_comparison.py
misiugodfrey: I'll strip the module docstring down and let the function docstring contain the details.

common/testing/result_comparison.py
misiugodfrey: I'll add a check.

common/testing/result_comparison.py
misiugodfrey: I'm not sure there' s more that we can do in this case, as a LIMIT without a sort_by means any limit-sized subset of the data is valid.

common/testing/result_comparison.py
misiugodfrey: It's a DuckDB thing.  DuckDB defaults to ASC → NULLS LAST, DESC → NULLS FIRST, and we need to track this to know where we should expect nulls in ORDER BY columns.

misiugodfrey (commented):
Addressed recent feedback.

common/testing/requirements.txt
paul-aiyedun, line 1: `common/testing` is not a Python project, so I don't think `requirements.txt` should be here. The dependencies should probably be managed by the project that uses the shared modules/files.

common/testing/result_comparison.py
paul-aiyedun: Perhaps, @TomAugspurger can speak to why this is done for Polars, but I believe the TPC-H specification states that column names are optional
>2.1.3.4 (a) Columns appear in the order specified by the SELECT list of either the functional query definition or an
approved variant. Column headings are optional.

Also, it is possible to have queries without clear column names. For instance, Q18 projects `sum(l_quantity)` without an alias and so, the column name would be query engine defined.

presto/scripts/run_benchmark.sh
```diff
@@ -292,3 +307,32 @@ echo "Using PRESTO_IMAGE_TAG: $PRESTO_IMAGE_TAG"
...
+  fi
+  if [[ -n ${PRESTO_EXPECTED_RESULTS_DIR} ]]; then
+    VALIDATE_ARGS+=(--reference-results-dir "${PRESTO_EXPECTED_RESULTS_DIR}")
+  fi
+  if [[ -n ${QUERIES} ]]; then
+    VALIDATE_ARGS+=(--queries "${QUERIES}")
+  fi
+
+  ACTUAL_OUTPUT_DIR="${OUTPUT_DIR:-$(pwd)/benchmark_output}"
+  [[ -n ${TAG} ]] && ACTUAL_OUTPUT_DIR="${ACTUAL_OUTPUT_DIR}/${TAG}"
```
paul-aiyedun, line 332: We are not using the same python function per https://github.com/rapidsai/velox-testing/pull/275#discussion_r3075065753.

common/testing/result_comparison.py
paul-aiyedun: Why do we only get it for the first sorted column?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,500 @@
...
+    sort_cols = [col for col, _ in sort_by]
+    descending = [d for _, d in sort_by]
+
+    # 6. Validate sort order
+    _validate_sort_order(actual, sort_cols, descending, nulls_last, "actual")
+    _validate_sort_order(expected, sort_cols, descending, nulls_last, "expected")
+
+    if limit is None:
+        # ORDER BY, no LIMIT — sort by non-float cols for tie-breaking
+        _assert_frames_equal(_sort_for_comparison(actual), _sort_for_comparison(expected))
```
paul-aiyedun, line 444: Why are we sorting the results in this case?

common/testing/result_comparison.py
TomAugspurger: Having the column names is helpful for debugging, and simplifies the rest of the validation logic (which can now assume column names match) and error reporting (it's easier to say and read "the data type of column `'<name>'` doesn't match" rather than something about a positional index or `left_name=... right_name=...`)

> the column name would be query engine defined

We made minor changes to our polars expressions to match (e.g. the `.alias("sum(l_quantity)")` to match duckdb [here](https://github.com/rapidsai/cudf/blob/0665e7dd7d0e27a580d4a8626ea467776b74ac4c/python/cudf_polars/cudf_polars/experimental/benchmarks/pdsh.py#L839)). Nothing too onerous.

common/testing/requirements.txt
misiugodfrey: Done.

common/testing/result_comparison.py
misiugodfrey: If the SQL has no LIMIT clause, then we can sort the actual and expected output to get a unique ordering that both should agree on for comparison.  If we don't sort the non-order-by columns, then we could have different orders between the expected and actual when there are ties in the order-by columns.  Sorting on everything guarantees a unique order but can't be used if there is a LIMIT.

misiugodfrey:
I've removed the nulls_last/first part because it should not be relevant to our benchmarks (sorted colums are non-null) and it was adding unnecessary complexity.

presto/testing/common/fixtures.py
```diff
@@ -15,7 +15,7 @@ def tpch_queries(request, presto_cursor):
     # The "fraction" portion of Q11 is a value that depends on scale factor
     # (it should be 0.0001 / scale_factor), whereas our query is currently hard-coded as 0.0001.
     value_ratio = 0.0001 / float(test_utils.get_scale_factor(request, presto_cursor))
-    queries["Q11"] = queries["Q11"].format(SF_FRACTION=f"{value_ratio:.12f}")
+    queries["Q11"] = queries["Q11"].format(SF_FRACTION=f"{value_ratio:.10f}")
```
misiugodfrey, line 18: This needed to change to 10f because that is what the duckdb expected results generator is using, and using a different precision was leading to discrepancies between the two engines.  Once we switch to decimal support, this should no longer matter.

common/testing/result_comparison.py
```diff
@@ -0,0 +1,480 @@
...
+# ---------------------------------------------------------------------------
+# Column name reconciliation
+# ---------------------------------------------------------------------------
+
+
+def _reconcile_col_names(actual: pd.DataFrame, expected: pd.DataFrame) -> pd.DataFrame:
+    """
+    Rename actual columns to match expected column names positionally.
+
+    TPC-H does not require column headings (spec §2.1.3.4), and anonymous
```
paul-aiyedun, line 107: I think we can simplify things/avoid confusion by removing this column name reconciliation requirement. If we want to display the column name in error messages, we can do the name lookup at that point.

common/testing/result_comparison.py
```diff
@@ -0,0 +1,480 @@
...
+        sample = series.dropna()
+        if len(sample) > 0:
+            val = sample.iloc[0]
+            return isinstance(val, (datetime.date, pd.Timestamp))
+    return False
+
+
+def _normalize_df(df: pd.DataFrame, ref: pd.DataFrame) -> pd.DataFrame:
+    """
+    Normalize df's column dtypes for comparison, using ref only for dtype detection
```
paul-aiyedun, line 167: Why is the normalization of a dataframe relative to another dataframe?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,480 @@
...
+    limit = get_limit(query_sql)
+
+    if not sort_cols:
+        # No ORDER BY (or unparsable) — sort both sides and compare
+        _assert_frames_equal(_sort_for_comparison(actual), _sort_for_comparison(expected))
+        return
+
+    # 6. Validate sort order
+    _validate_sort_order(actual, sort_cols, descending, "actual")
+    _validate_sort_order(expected, sort_cols, descending, "expected")
```
paul-aiyedun, line 421: Why is this validation needed?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,480 @@
...
+    ascending = [not d for d in descending]
+    boundary = actual.sort_values(by=sort_cols, ascending=ascending, na_position="last").iloc[-1][sort_cols]
+
+    actual_non_tie_mask = _build_non_tie_mask(actual, sort_cols, descending, boundary)
+    expected_non_tie_mask = _build_non_tie_mask(expected, sort_cols, descending, boundary)
+
+    actual_non_ties = actual[actual_non_tie_mask].reset_index(drop=True)
+    expected_non_ties = expected[expected_non_tie_mask].reset_index(drop=True)
+    actual_ties = actual[~actual_non_tie_mask].reset_index(drop=True)
+    expected_ties = expected[~expected_non_tie_mask].reset_index(drop=True)
```
paul-aiyedun, line 438: Can you please expand on the boundary determination logic? My thinking is that we would only need to pull the first sorted column and determine (from the bottom up) the index where ties begin. This would imply a simpler logic than this, but I may be missing something.

common/testing/result_comparison.py
```diff
@@ -0,0 +1,480 @@
...
+
+    # Non-tie rows: full comparison
```
paul-aiyedun, line 440: In the worst case scenario, the entire result set could have tie rows, in which case, no comparison occurs. I think it makes sense to report the number and/or percentage of rows that are discarded in the comparison.

common/testing/result_comparison.py
```diff
@@ -0,0 +1,480 @@
...
+
+    # 4. Row count
+    if len(actual) != len(expected):
+        raise AssertionError(f"Row count mismatch: {len(actual)} (actual) vs {len(expected)} (expected)")
+
+    # 5. Parse ORDER BY / LIMIT
+    sort_cols, descending = get_orderby_info(query_sql, list(actual.columns))
+    limit = get_limit(query_sql)
+
+    if not sort_cols:
```
paul-aiyedun, line 414: From this point onward, consider this flow, which may be easier to follow (avoids multiple returns):
```
actual_limit_tie_boundary = None
expected_limit_tie_boundary = None

if limit:
  actual_limit_tie_boundary = _get_limit_tie_boundary(actual, sort_cols)
  expected_limit_tie_boundary = _get_limit_tie_boundary(expected, sort_cols)

if actual_limit_tie_boundary != expected_limit_tie_boundary:
  raise AssertionError(...)

actual_first_n_sorted = _sort_first_n_for_comparison(actual, actual_limit_tie_boundary)
expected_first_n_sorted = _sort_first_n_for_comparison(expected, expected_limit_tie_boundary)

_assert_frames_equal(actual_first_n_sorted, expected_first_n_sorted)

if sort_cols:
  _assert_sort_cols_equal(actual, expected, sort_cols)
```

`_get_limit_tie_boundary`, `_sort_first_n_for_comparison`, and `_assert_sort_cols_equal` are new functions to be implemented (I can expand on this if needed).

paul-aiyedun (commented):
I had a few more questions about the result set comparison logic. I think there are opportunities for simplification.

common/testing/result_comparison.py
misiugodfrey: Ok.  I'll go ahead and remove the name reconciliation.  We'll just go by column indexes.

common/testing/result_comparison.py
misiugodfrey: This is to handle the case where one engine returns an object dtype and the other returns datetime64.  This happens a few times in tpch, where Presto and Spark both return date columns as object dtypes whereas DuckDB returns datetime64[us].

We use the ref to know if the other string column needs converting.  We could split this logic up, as it's a little different than the other types we check for, but I do think something like this is necessary to catch this case.

common/testing/result_comparison.py
misiugodfrey: This was in place just to add clearer error messages.  It's not a big difference though, so I'm fine with removing it.

common/testing/result_comparison.py
misiugodfrey: It's essentially the same as what you mentioned, but it's tracking a mask of all the rows for what is tied and what is not.  I think you're right though that an index can track all the information since the ties would be contiguous, so a mask is unnecessary.  I'll just go with the index solution.

common/testing/result_comparison.py
misiugodfrey: Sure.  Added a warning with %boundary ties.

common/testing/result_comparison.py
misiugodfrey: I think my other refactoring has addressed most of this comment.

common/testing/result_comparison.py
```diff
@@ -0,0 +1,405 @@
...
+        )
+
+    # 2. Capture expected column names (used for ORDER BY name resolution), then
+    # reset both frames to positional integer indices for all subsequent operations.
+    expected_col_names = list(expected.columns)
+    n = len(expected_col_names)
+    actual = actual.copy()
+    expected = expected.copy()
+    actual.columns = range(n)
+    expected.columns = range(n)
```
paul-aiyedun: Why do we need to do this?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,405 @@
...
+
+    # 4. Row count
+    if len(actual) != len(expected):
+        raise AssertionError(f"Row count mismatch: {len(actual)} (actual) vs {len(expected)} (expected)")
+
+    # 5. Parse ORDER BY / LIMIT
+    sort_col_indices, descending = get_orderby_info(query_sql, expected_col_names)
+    limit = get_limit(query_sql)
+
+    if not sort_col_indices or limit is None:
```
paul-aiyedun: I don't think the `or limit is None` part is needed?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,405 @@
...
+
+    # Non-tie rows: full comparison
+    _assert_frames_equal(
+        _sort_for_comparison(actual_sorted.iloc[:actual_tie_start]),
+        _sort_for_comparison(expected_sorted.iloc[:expected_tie_start]),
+    )
+
+    # Tie rows: compare only sort columns (non-sort columns may legitimately differ)
+    _assert_frames_equal(
+        _sort_for_comparison(actual_sorted.iloc[actual_tie_start:][sort_col_indices]),
```
paul-aiyedun: I don't think we should be sorting the sort columns before comparison?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,405 @@
...
+        return
+
+    # 6. ORDER BY + LIMIT: sort both frames, then split at the tie boundary
+    ascending = [not d for d in descending]
+    actual_sorted = actual.sort_values(by=sort_col_indices, ascending=ascending, na_position="last").reset_index(
+        drop=True
+    )
+    expected_sorted = expected.sort_values(by=sort_col_indices, ascending=ascending, na_position="last").reset_index(
+        drop=True
+    )
```
paul-aiyedun: Why do we sort the dataframe here?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,405 @@
...
+
+
+def _find_tie_start(df: pd.DataFrame, sort_col_indices: list[int], boundary: pd.Series) -> int:
+    """
+    Scan upward from the last row and return the index where the contiguous
+    block of boundary-tie rows begins. Rows at or above this index are non-ties.
+    """
+    for i in range(len(df) - 1, -1, -1):
+        row = df.iloc[i]
+        for col_idx in sort_col_indices:
```
paul-aiyedun: I think this only needs to check the first sorted column?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,405 @@
...
+            v, b = row[col_idx], boundary[col_idx]
+            if _is_null(b):
+                if not _is_null(v):
+                    return i + 1
+            elif _is_float_col(df[col_idx]):
+                fval = float(b)
+                if abs(float(v) - fval) > ABS_TOL + REL_TOL * abs(fval):
+                    return i + 1
```
paul-aiyedun: Do we still need to do checks in this way after the dataframe has been normalized?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,405 @@
...
+
+def _find_tie_start(df: pd.DataFrame, sort_col_indices: list[int], boundary: pd.Series) -> int:
+    """
+    Scan upward from the last row and return the index where the contiguous
+    block of boundary-tie rows begins. Rows at or above this index are non-ties.
+    """
+    for i in range(len(df) - 1, -1, -1):
+        row = df.iloc[i]
+        for col_idx in sort_col_indices:
+            v, b = row[col_idx], boundary[col_idx]
```
paul-aiyedun: Can you please expand on why we need to compare with boundary vs consecutive rows of the first sorted column?

common/testing/result_comparison.py
paul-aiyedun: >We use the ref to know if the other string column needs converting.

Can you please expand on this? How does a string column come into play?

common/testing/result_comparison.py
misiugodfrey: This handles the theoretical case where we could have an ORDER BY with no LIMIT and all sorted rows have the same value.  In this case all rows would be considered ties and the non-sort columns are skipped entirely.  Effectively we would treat all rows as boundary ties, but we don't have to ignore them in the same way because we know we are looking at 100% of the result data (they look like boundary ties, but we don't have to consider there are some  ties that have been pruned by limit).

With "or limit is None" we fallback and sort-and-compare will run instead - which will compare more traditionally.

I don't think we are likely to hit anything like this in practice... but it still seems more correct to leave it in.

common/testing/result_comparison.py
misiugodfrey: Good point.  Removed.

common/testing/result_comparison.py
```diff
@@ -0,0 +1,320 @@
...
+
+ValidationStatus = Literal["passed", "failed", "expected-failure", "not-validated"]
+
+
+# ---------------------------------------------------------------------------
+# SQL parsing helpers
+# ---------------------------------------------------------------------------
+
+
+def get_orderby_info(query_sql: str, expected_col_names: list[str]) -> list[int]:
```
paul-aiyedun: Nit: `get_orderby_col_indices`

common/testing/result_comparison.py
```diff
@@ -0,0 +1,320 @@
...
+            except (ValueError, TypeError):
+                pass
+
+        if isinstance(key, sqlglot.exp.Column):
+            name = key.name
+            if name in expected_col_names:
+                sort_col_indices.append(expected_col_names.index(name))
+                continue
+
+        # Complex expression — skip ORDER BY validation entirely.
```
paul-aiyedun: Should we log a warning here?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,320 @@
...
+    limit_node = next((e for e in expr.find_all(sqlglot.exp.Limit)), None)
+    if limit_node is None:
+        return None
+    limit_expr = limit_node.args.get("expression")
+    if limit_expr is None:
+        return None
+    try:
+        return int(limit_expr.this)
+    except (ValueError, TypeError, AttributeError):
+        return None
```
paul-aiyedun, line 100: Log a warning in this case?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+def _is_float_col(series: pd.Series) -> bool:
+    return pd.api.types.is_float_dtype(series)
+
+
+def _sort_by_non_float(df: pd.DataFrame) -> pd.DataFrame:
+    """
+    Sort df by all non-float columns for deterministic position-by-position
+    comparison. Float columns are excluded because tiny cross-engine precision
```
paul-aiyedun: Do cross-engine precision differences apply when comparing rows from the same dataframe?

common/testing/test_result_comparison.py
paul-aiyedun, line 1: Was this file added intentionally?

paul-aiyedun (commented):
Changes overall look good to me. I had some minor comments, but please verify that the non-float sorting logic is correct.

common/testing/result_comparison.py
misiugodfrey: There's no issue across the same dataframe.  The issue is if we sort dataframes that came from different engines and then compare them.  I've clarified this in the comments.

common/testing/test_result_comparison.py
misiugodfrey: I've been using this for some quick testing in this PR.  Did we have interest in adding tests like this, or should I remove it?

presto/testing/common/fixtures.py
```diff
@@ -14,6 +14,11 @@ def tpch_queries(request, presto_cursor):
 
     # The "fraction" portion of Q11 is a value that depends on scale factor
     # (it should be 0.0001 / scale_factor), whereas our query is currently hard-coded as 0.0001.
+    # Format with 12 decimal places to match the DuckDB reference image's
+    # substitution — both sides emit the same literal so the HAVING threshold
+    # agrees on the same value. (Note: ".10f" at SF=3000 gave only 3 significant
+    # figures, shifting the threshold by ~0.1% and producing row-count
+    # mismatches against the DuckDB reference.)
```
paul-aiyedun, line 21: The .10f note here seems arbitrary.

common/testing/result_comparison_test.py
```diff
@@ -0,0 +1,200 @@
...
+    assert result.iloc[0, 0] == 1.0
+    assert pd.isna(result.iloc[1, 0])
+    assert result.iloc[2, 0] == 2.0
+
+
+def test_canonicalize_handles_zero():
+    df = pd.DataFrame({0: [0.0, 1e-20, -0.0]})
+    result = _canonicalize_floats(df)
+    assert result.iloc[0, 0] == 0.0
+    # 1e-20 is far smaller than any reasonable quantum; canonicalizes to 0 effectively
```
paul-aiyedun: There is this comment but no corresponding `result.iloc[1, 0]` assertion.

common/testing/result_comparison_test.py
```diff
@@ -0,0 +1,200 @@
...
+# ---------------------------------------------------------------------------
+# _canonicalize_floats
+# ---------------------------------------------------------------------------
+
+
+def test_canonicalize_collapses_engine_precision():
+    # Two values differing at ~1e-12 relative — well below quantum (1e-8).
+    df = pd.DataFrame({0: [100.000000000001, 100.000000000002]})
+    result = _canonicalize_floats(df)
+    assert result.iloc[0, 0] == result.iloc[1, 0]
```
paul-aiyedun: Should there be a before canonicalization assertion?

Same comment applies to other canonicalization tests.

common/testing/result_comparison_test.py
```diff
@@ -0,0 +1,200 @@
...
+
+def test_canonicalize_handles_negatives():
+    df = pd.DataFrame({0: [-100.000000000001, -100.000000000002]})
+    result = _canonicalize_floats(df)
+    assert result.iloc[0, 0] == result.iloc[1, 0]
+
+
+def test_canonicalize_scales_with_magnitude():
+    # Big magnitude: precision is relative, so 1e-3 wiggle on 1e7 gets flattened
+    # (quantum at 1e7 is 1e7*1e-8 = 0.1).
```
paul-aiyedun: Can you please expand on this and why the assertion is different from `test_canonicalize_preserves_distinct_values`?

common/testing/result_comparison_test.py
paul-aiyedun, line 1: Should this also be testing the `validate_query_result` entry point?

common/testing/result_comparison.py
```diff
@@ -0,0 +1,461 @@
...
+
+    # 3. Row count
+    if len(actual) != len(expected):
+        raise AssertionError(f"Row count mismatch: {len(actual)} (actual) vs {len(expected)} (expected)")
+
+    # 4. Parse ORDER BY / LIMIT
+    sort_col_indices, descending = get_orderby_col_indices(query_sql, expected_col_names)
+    limit = get_limit(query_sql)
+
+    # 5. Verify each frame is correctly sorted per the SQL ORDER BY (before
```
paul-aiyedun: Why do we need to do this?

https://github.com/rapidsai/velox-testing/pull/275#discussion_r3113696361

common/testing/result_comparison.py
```diff
@@ -0,0 +1,461 @@
...
+    # canonicalization, so we see what the engines actually produced).
+    _verify_sort(actual, sort_col_indices, descending)
+    _verify_sort(expected, sort_col_indices, descending)
+
+    # 6. Canonicalize floats — engine precision noise becomes bit-identical.
+    actual = _canonicalize_floats(actual)
+    expected = _canonicalize_floats(expected)
+
+    # 7. Canonical sort: ORDER BY columns first (with directions), then all
```
paul-aiyedun: It seems like we have reverted back to sorting all columns?

https://github.com/rapidsai/velox-testing/pull/275#discussion_r3125914276

common/testing/result_comparison.py
```diff
@@ -0,0 +1,461 @@
...
+    Both sides null → equal. One side null → mismatch.
+
+    Raises AssertionError describing up to MAX_MISMATCHES differences.
+    """
+    if len(df1) != len(df2):
+        raise AssertionError(f"Row count mismatch: {len(df1)} vs {len(df2)}")
+
+    mismatches: list[str] = []
+
+    for col_idx in range(df1.shape[1]):
```
paul-aiyedun, line 315: The current logic seems more complex than the previous iteration. What is the reason for the structural change?

Previous comparison logic:
https://github.com/rapidsai/velox-testing/blob/41fdfa4a48171401c1fc44717c99d3151aa2a9ff/common/testing/result_comparison.py#L168-L206

common/testing/result_comparison.py
```diff
@@ -0,0 +1,461 @@
...
+# ---------------------------------------------------------------------------
+# Frame comparison helpers
+# ---------------------------------------------------------------------------
+
+
+def _is_float_col(series: pd.Series) -> bool:
+    return pd.api.types.is_float_dtype(series)
+
+
+def _canonicalize_floats(df: pd.DataFrame) -> pd.DataFrame:
```
paul-aiyedun: Nit: Consider `_normalize_floats` i.e. consistently use `normalize`.

paul-aiyedun (commented):
The float canonicalization makes sense to me. Although, the diff seems more complex than expected.

common/testing/result_comparison.py
misiugodfrey: The reason for the change in complexity was to increase performance via vectorization.   Performing  these comparisions as column-based vector comparisons speeds things up by about 400x, which is necessary to scale up to SF30k and beyond.

This does mean we need array math that numpy supports instead of the scalar _is_null stuff, but I've tried to make this more readable by separating it into smaller discrete functions.

paul-aiyedun (approved):

PR #277, opened 2026-03-17
CI workflow refactoring

Avinash-Raj:
**Refactor CI Images workflow:**

- Splitted up `build.yml` and `test.yml` into `velox-build.yml`, `presto-build.yml`, `velox-test.yml`, `presto-test.yml`
- Renamed the existing `presto-test.yml`, `velox-test.yml` into `presto-standalone-test.yml`, `velox-standalone-test.yml` (which could be removed later)

### DIFF
--- a/.github/workflows/presto-test.yml
+++ b/.github/workflows/presto-test.yml
@@ -161,7 +161,11 @@ jobs:
           NUM_WORKERS: 1
           OVERWRITE_CONFIG: 'true'
         working-directory: presto/scripts
-        run: ./generate_presto_config.sh --ci
+        run: ./generate_presto_config.sh
+
+      - name: Patch configs for CI
+        working-directory: presto/scripts
+        run: ./patch_ci_configs.sh ../docker/config/generated/gpu
 
       - name: Create Docker network
         run: docker network create "${DOCKER_NETWORK}"


bdice (approved):

.github/workflows/presto-standalone-test.yml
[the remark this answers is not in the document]
Avinash-Raj: Exactly!

.github/actions/resolve-inputs/action.yml
```diff
@@ -0,0 +1,79 @@
...
+            echo "Parsed presto image: presto=${BASH_REMATCH[1]}, velox=${BASH_REMATCH[2]}, date=${BASH_REMATCH[3]}"
+
+          # Velox format: velox-{SHA}-...-{DATE}[-arch]
+          elif [[ "$TAG" =~ ^velox-([a-f0-9]+)-.*-([0-9]{8})(-[a-z0-9]+)?$ ]]; then
+            echo "velox_short_sha=${BASH_REMATCH[1]}" >> "$GITHUB_OUTPUT"
+            echo "presto_short_sha=" >> "$GITHUB_OUTPUT"
+            echo "date=${BASH_REMATCH[2]}" >> "$GITHUB_OUTPUT"
+            echo "Parsed velox image: velox=${BASH_REMATCH[1]}, date=${BASH_REMATCH[2]}"
+
+          else
```
misiugodfrey: Since we are making this a generic action, did we also want to support the presto-coordinator case?  IIUC it uses a different tag format `presto-coordinator-{SHA}-{DATE}`.

Avinash-Raj: The `resolve-inputs` action used by the `velox-test` and `presto-test` workflows accepts either a Velox or Presto image. If a Presto image URI is passed, it automatically forms the coordinator image URI (using the `presto-sha` and `build-date` extracted from the Presto image URI) which running the presto integration tests.

misiugodfrey (approved):
It's not introduced in this PR, so I think it's fine for these refactors to go through; however, I think we should have a conversation after about whether we want our CI to be using a different build/run process than our scripts.

Specifically, I think we are using/building different entrypoints, and I don't want to enable a situation where our CI passes/fails and our scripts do the opposite unless we have good reason.

Issue #245, opened 2026-02-18
Add username tag by default for the Presto native dependency image


paul-aiyedun:
Update the default Presto native dependency image name from `presto/prestissimo-dependency:centos9` to `presto/prestissimo-dependency:{username}` in order to avoid image overrides when using shared machines.

PR #294, opened 2026-03-25
Tag deps image with user name

misiugodfrey:
Addresses this issue: closes https://github.com/rapidsai/velox-testing/issues/245

Tag the dependency image with the username to avoid collisions between multiple users on the same machine.  This is similar to how we currently tag the worker/coordinator images.


paul-aiyedun (commented):
Changes overall look good to me. However, I had a couple of questions about places where we still seem to be using the old image name.

PR #296, opened 2026-03-30
Update tpc-h data generation

TomAugspurger:
This updates the data generation code in velox-testing to use the strategy / options from https://github.com/TomAugspurger/tpchgen-rs/tree/tom/sync-upstream-clean/scripts.

These rely on a fork of tpcgen-rs adds new features to control data generation. Users will need to ensure that they're in an enviroment with the appropriate tpchgen-cli on their path.

### DIFF
--- a/.github/workflows/velox-benchmark.yml
+++ b/.github/workflows/velox-benchmark.yml
@@ -95,10 +95,10 @@ jobs:
           source .venv/bin/activate
           pip install -q -r benchmark_data_tools/requirements.txt
           python benchmark_data_tools/generate_data_files.py \
-            --benchmark-type tpch \
-            --data-dir-path "${TPCH_DATA_DIR}" \
-            --scale-factor 0.01 \
-            --convert-decimals-to-floats
+            tpch \
+            --output "${TPCH_DATA_DIR}" \
+            --scale 0.01 \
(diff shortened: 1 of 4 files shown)


TomAugspurger (commented):
I'd appreciate help on integrating this the rest of the way into velox-testing.

I've updated one call site that used `benchmark_data_tools/generate_data_files.py` to use the new subcommands. But I haven't attempted to handle the Python environment changes, and I haven't updated any of the other spots.

And I'd appreciate feedback on any other uses of `benchmark_data_tools/generate_data_files.py`. Are they using that `requirements.txt`? Are there places where updating that `requirements.txt` to be "correct" and point to the tpchgen-rs fork is harmful? For example, if you're installing from `requirements.txt` in a spot that runs data generation only if the data aren't present, then building tpcgen-rs from source (and possibly failing) would be unfortunate.

paul-aiyedun (changes requested):
Can we please narrow the scope of this PR to just the use of the tpchgen-cli fork (without changing the `generate_data_files.py` interface)? We can add support for different column encodings in a subsequent PR.

TomAugspurger:
This was superseded by https://github.com/rapidsai/velox-testing/pull/303 and others.

PR #306, opened 2026-04-09
Remove data generation Parquet rewrite step

paul-aiyedun:
* Replace the two-step data generation and rewrite flow with direct tpchgen-cli decimal conversion.

* Delete rewrite_parquet.py and remove the intermediate temporary directory and orig_test_data_dir_path from the test fixture and all test usages.

* Remove data_content_match_test.py and the test_row_group_size_match test case.

* Add decimal_conversion_test.py to verify that all decimal columns become floats when convert_decimals_to_floats is set.


bdice (approved):

misiugodfrey (approved):
Looks good.  I think we can go ahead, although the real test will be once we can get full benchmark testing on the remaining `delta-byte-length-array` option gets added.

PR #309, opened 2026-04-10
fix(ci): Re-enable velox_table_evolution_fuzzer_test in CPU builds

simoneves:
Once https://github.com/facebookincubator/velox/pull/17106 lands, this test can once more be safely run in CI.

### DIFF
--- a/velox/scripts/test_velox.sh
+++ b/velox/scripts/test_velox.sh
@@ -98,9 +98,7 @@ TEST_PREAMBLE='if [ -f "/opt/miniforge/etc/profile.d/conda.sh" ]; then
   export CLASSPATH=$(/usr/local/hadoop/bin/hdfs classpath --glob)'
 
 if [[ "$DEVICE_TYPE" == "cpu" ]]; then
-  # disable velox_table_evolution_fuzzer_test pending resolution of too-many-open-files problem
-  # seves 1/9/26
-  SKIP_TESTS="velox_exec_test|velox_hdfs_file_test|velox_hdfs_insert_test|velox_table_evolution_fuzzer_test"
+  SKIP_TESTS="velox_exec_test|velox_hdfs_file_test|velox_hdfs_insert_test"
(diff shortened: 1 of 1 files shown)


mattgara (approved):
LGTM, is there a passing example run with this enabled?

simoneves:
We can test this part officially once the fix lands.

Avinash-Raj:
But this  script `test_velox.sh` is no longer being called from a [`velox-test.yml` CI workflow.](https://github.com/rapidsai/velox-testing/blob/main/.github/workflows/velox-test.yml#L129)

simoneves:
> But this  script `test_velox.sh` is no longer being called from a [`velox-test.yml` CI workflow.](https://github.com/rapidsai/velox-testing/blob/main/.github/workflows/velox-test.yml#L129)

Oops! OK... what runs tests now then? Does it need an equivalent change. Would look but on my phone and too lazy to fire up the computer...

bdice:
The CPU test command is here: https://github.com/rapidsai/velox-testing/blob/0f2e363f09e9be86519f4839a2cbc4089e3bb513/.github/workflows/velox-test.yml#L87-L88

If we are supposed to use a common/shared test script, that's fine with me as long as we can call it through Docker with a custom image like the current GitHub Actions setup.

simoneves:
> If we are supposed to use a common/shared test script, that's fine with me as long as we can call it through Docker with a custom image like the current GitHub Actions setup.

If the new YAML implementation does the job, I don't see a need to change it. Presumably this is how we were reminded that the `table_evolution_fuzzer` test was problematic, as the old script didn't run it. Also presumably there are no issues with the other three tests the old script didn't run either.

If that's now the definitive method, should we remove the `test_velox.sh` script completely, or leave it for posterity?

Avinash-Raj:
We could remove the `test_velox.sh` script completely and make sure that the `ctest` command should find/include the `table_evolution_fuzzer` test on tests discovery phase.

simoneves:
> We could remove the `test_velox.sh` script completely and make sure that the `ctest` command should find/include the `table_evolution_fuzzer` test on tests discovery phase.

As long as there's still a definitive way to run the same tests from the command-line that CI will do, and they stay in sync, that's fine with me.

PR #310, opened 2026-04-15
Removed docker warnings for unset variables

misiugodfrey:
Added default value substitution to our start/stop scripts to prevent a series of docker warnings (often duplicates) about unset variables.  I don't  think these warnings were useful, and they help clean up the output from the scripts.

### DIFF
--- a/presto/docker/docker-compose/template/docker-compose.native-gpu.yml.jinja
+++ b/presto/docker/docker-compose/template/docker-compose.native-gpu.yml.jinja
@@ -2,7 +2,7 @@ x-presto-native-worker-gpu: &gpu_worker_base
   extends:
     file: ../../docker-compose.common.yml
     service: presto-base-native-worker
-  image: presto-native-worker-gpu:${PRESTO_IMAGE_TAG}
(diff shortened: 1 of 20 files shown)


presto/docker/docker-compose/template/docker-compose.native-gpu.yml.jinja
```diff
@@ -2,7 +2,7 @@ x-presto-native-worker-gpu: &gpu_worker_base
   extends:
     file: ../../docker-compose.common.yml
     service: presto-base-native-worker
-  image: presto-native-worker-gpu:${PRESTO_IMAGE_TAG}
+  image: presto-native-worker-gpu:${PRESTO_IMAGE_TAG:-}
```
paul-aiyedun, line 5: I believe we expect variables like this to be set. Would it make sense to instead do something like `${PRESTO_IMAGE_TAG:?error}`?

misiugodfrey: Good suggestion.  Done.

paul-aiyedun (approved):

PR #312, opened 2026-04-17
Fix CI image deletion

misiugodfrey:
Our CI generates images for arm and x86 architectures and then merges them into a single manifest; it is then supposed to delete the pointers to the x86/arm specific images so that we don't pollute our repository with redundant images (because we now have multi-arch images in place).

However, the approach we were using to delete images was not working correctly, as it was failing with insufficient  permissions, and therefore not deleting the images.

This PR use the gh api to delete the images instead of curl, so that we are able to delete the images with the appropriate permissions.  It also moves the image deletion to after the build is complete, so that we don't end up deleting the deps image when they are still needed to build the worker images.

### DIFF
--- a/.github/workflows/presto-build.yml
+++ b/.github/workflows/presto-build.yml
@@ -378,16 +378,6 @@ jobs:
                 "${REGISTRY}/${IMAGE_NAME}:presto-deps-${PRESTO_SHORT_SHA}-velox-${VELOX_SHORT_SHA}-cuda${cuda}-${BUILD_DATE}-arm64"
             done
           fi
-      - name: Delete intermediate arch-specific tags
-        continue-on-error: true
-        uses: ./.github/actions/delete-arch-tags
-        with:
-          registry: ${{ env.REGISTRY }}
-          image-name: ${{ env.IMAGE_NAME }}
-          base-tags: |
-            presto-deps-${{ inputs.presto_short_sha }}-velox-${{ inputs.velox_short_sha }}-cuda12.9-${{ inputs.date }}
-            presto-deps-${{ inputs.presto_short_sha }}-velox-${{ inputs.velox_short_sha }}-cuda13.1-${{ inputs.date }}
-
   merge-presto-build:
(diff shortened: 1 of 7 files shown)


.github/workflows/presto-build.yml
```diff
@@ -378,16 +378,6 @@ jobs:
                 "${REGISTRY}/${IMAGE_NAME}:presto-deps-${PRESTO_SHORT_SHA}-velox-${VELOX_SHORT_SHA}-cuda${cuda}-${BUILD_DATE}-arm64"
             done
           fi
-      - name: Delete intermediate arch-specific tags
```
paul-aiyedun, line 381: >It also moves the image deletion to after the build is complete, so that we don't end up deleting the deps image when they are still needed to build the worker images.

Do the worker images use the intermediate image tags?

.github/workflows/velox-build.yml
```diff
@@ -292,16 +292,6 @@ jobs:
                 "${REGISTRY}/${IMAGE_NAME}:velox-deps-${VELOX_SHORT_SHA}-cuda${cuda}-${BUILD_DATE}-arm64"
             done
           fi
-      - name: Delete intermediate arch-specific tags
```
paul-aiyedun, line 295: Same question as above.

.github/workflows/presto-build.yml
misiugodfrey: When building the worker images they use the intermediate tags for referencing the deps images (so building the arm-based gpu-worker image will use `presto-deps-*-arm64`.  I left the build process alone so they still reference whichever images they want and only untag the unnecessary images once we are finished building all the images.

misiugodfrey:
Had to fix some pre-check failures that were outside the scope of the original PR.

paul-aiyedun (approved):
Looks good to me. Thanks for addressing the unrelated PR check failures.

@bdice Is there a way of making these checks required for all PRs? We seem to have PRs that allow merging without requiring the PR checks to be run.

PR #273, opened 2026-03-13
New Cluster Scripts

misiugodfrey:
Cluster Benchmarking Infrastructure

  This PR adds multi-node TPC-H benchmarking support in a new ccluster, including CPU and GPU variants, result validation, automated result posting, and sweep tooling.

  New scripts

  - run-sweep.sh — Automates running launch-run.sh + post_results.py across multiple node/scale-factor combinations
  - run-presto-benchmarks.sh — Orchestrates the full benchmark lifecycle (setup, coordinator, workers, queries, results collection)
  - pull_ghcr_image.sh / enroot-decompress.sh — Pull container images from ghcr.io and save as .sqsh files, with transparent gzip/zstd decompression support
  - launch-gen-data.sh / gen-tpch-data.slurm — TPC-H data generation jobs
  - launch-analyze-tables.sh / run-analyze-tables.sh / run-analyze-tables.slurm — Hive table analysis jobs

  Benchmark execution (functions.sh, launch-run.sh, run-presto-benchmarks.slurm)

  - NUMA-aware worker placement with optional --no-numa flag for older images
  - CPU benchmark mode (--cpu) — disables cuDF, one worker per node, no GPU allocation
  - Conditional CUDA_VISIBLE_DEVICES and NVIDIA_VISIBLE_DEVICES for GPU vs CPU runs
  - Writable coord_data and per-worker worker_data_N directories to avoid EROFS on read-only squashfs
  - Miniforge bind-mount so Python shebangs resolve inside the coordinator container
  - inject_benchmark_metadata — injects run context (image digest, engine, node/GPU counts, timestamp) into benchmark_result.json on exit
  - collect_results — copies configs and logs into result_dir for archival
  - Configurable nodelist, image names, and output path via CLI flags
  - Stale result prevention: result_dir and OUTPUT_DIR are fully removed (rm -rf) before each run so cancelled jobs cannot post old data

  Config generation (generate_presto_config.sh, templates)

  - Per-worker config directories (etc_worker_N/) are now generated for both GPU and CPU variants
  - CPU variant explicitly sets cudf.enabled=false
  - Worker config template additions: async-data-cache-enabled=false, cudf.jit_expression_enabled=false, cudf.intra_node_exchange=true, cudf.concat_optimization_enabled, cudf.batch_size_min_threshold

  Result reporting (post_results.py, validate_results.py)

  - Metadata is now read directly from benchmark_result.json context (no separate benchmark.json required)
[...]

PR #313, opened 2026-04-20
Updated Slurm Scripts

misiugodfrey:
Summary

Rebuilds the presto/slurm/presto-nvl72/ launcher layer on current main (replacing the stale misiug/SpaceMicePOC-based branch), ports the cluster-specific fixes that had accumulated there, and adds a cluster-wide shared Hive-metastore layer so analyze runs are a one-time cost per (image, scale-factor) rather than a per-clone tax.

With the defaults in this PR, on this cluster a fresh clone can run:

./launch-run.sh -n 2 -s 3000 -i 1

and go straight to queries against the pre-analyzed metastore — no data gen, no ANALYZE TABLE, no env setup. Snapshots for SF 1000 / 3000 / 10000 / 30000 are already published at /scratch/prestouser/shared_hive_metadata/HIVE-METASTORE-20260419-no-delta/ against /scratch/prestouser/tpch-rs-float-no-delta/.

What landed

Twelve commits across three themes:
  
1. Branch rebase onto main — presto/slurm/presto-nvl72/ ported from misiug/NewClusterScripts; non-slurm changes that upstream had landed (custom-queries plumbing, --skip-analyze-check, benchmark config output) were dropped in favor of the upstream versions. Validation wiring is deliberately deferred to rapidsai/velox-testing#275.
  
2. Cluster/config fixes
  - config: extend coordinator query.max-execution-time to 60m — large-SF Q21 was tripping the 10m default.
  - generate_presto_config: fix CPU multi-worker config generation — CPU variant now sets cudf.enabled=false and runs the worker-config-propagation loop (previously gated on GPU-only), so CPU multi-worker runs actually stand up more than one worker.
  - py_env: pin ncurses<6.5 — works around a conda-forge aarch64 symlink bug that kept conda create from succeeding on nvl72 nodes.
  - slurm: set worker LD_LIBRARY_PATH inside container instead of via pyxis — the current pyxis build double-substitutes --container-env=KEY=VAL, dropping the libcudf path and segfaulting the worker; moved the export into the in-container bash block.
  - slurm: stage static jq for coord, drop validate_results from run_queries — current upstream coord image lacks jq and can't install via yum (read-only squashfs overlay). We now drop a static jq into $VT_ROOT/.cache/bin/ once and prepend it to PATH inside the coord step.
  - slurm: zero gpu_count/gpu_name in injected metadata when cudf is disabled — CPU runs were mis-reporting gpu_count=N/NUM_WORKERS and gpu_name="NVIDIA GB200".
  - post_results: capture node_count, image_digest, engine repo/branch metadata — BenchmarkMetadata grew node_count and image_digest; post_results.py accepts --velox-branch/--velox-repo/--presto-branch/--presto-repo; --identifier-hash is now optional (falls back to the captured image digest). Payload's node_count comes from metadata instead of being hardcoded to 1.
  
3. Shared Hive metastore
  - slurm: add shareable Hive metastore snapshot layer — populate_hive_metastore_from_shared + publish_hive_metastore_to_shared helpers in functions.sh; atomic publish via .staging-<job_id> dir + mv -T. Keyed by HIVE_METASTORE_SHARED_ROOT + HIVE_METASTORE_VERSION + scale_factor. launch-analyze-tables.sh auto-publishes on success;
  launch-run.sh populates on demand when the local snapshot is missing. Retires the orphaned copy_hive_metastore{,_for_analyze} / HIVE_METASTORE_SOURCE scaffold that pointed at a /mnt/data/... path no longer mounted on this cluster.
  - slurm: forward HIVE_METASTORE_{VERSION,SHARED_ROOT} through launchers + slurm: export shared-metastore vars from slurm scripts to child bash — sbatch --export=ALL wasn't reliably propagating these vars through the .slurm → bash run-*.sh chain; fixed by adding them explicitly to each launcher's --export list and re-exporting them after
  source defaults.env.
  - slurm: adopt the published cluster-wide metastore as the default — defaults.env now defaults DATA=/scratch/prestouser/tpch-rs-float-no-delta, HIVE_METASTORE_SHARED_ROOT=/scratch/prestouser/shared_hive_metadata, HIVE_METASTORE_VERSION=HIVE-METASTORE-20260419-no-delta. Unset the version to disable sharing; bump the tag for image/format changes.  

How the shared metastore works
  
Paths in the file-based .prestoSchema records are container-relative (file:/var/lib/presto/data/hive/data/user_data/scale-<SF>/<table>), so a single post-ANALYZE TABLE snapshot is portable across users as long as everyone bind-mounts their $DATA at /var/lib/presto/data/hive/data/user_data/. Stats live inline in the same .prestoSchema JSON files, so a published snapshot is self-contained.
  
Layout:
  /scratch/prestouser/shared_hive_metadata/
  └── HIVE-METASTORE-20260419-no-delta/
      ├── tpchsf1000/
      ├── tpchsf3000/
      ├── tpchsf10000/
      └── tpchsf30000/
  
launch-run.sh populates .hive_metastore/tpchsf<SF>/ only when the local copy is absent, so local analyze output is never clobbered. Mismatched version or empty slot → fail-fast with a message pointing at launch-analyze-tables.sh


kingcrimsontianyu:
Awesome! I will merge or rebase https://github.com/rapidsai/velox-testing/pull/299 based on this work. These can be separate PRs to reduce the turnaround time.

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -59,46 +65,70 @@ function validate_environment_preconditions {
...
+
     # Coordinator runs as a background process, whereas we want to wait for cli
     # so that the job will finish when the cli is done (terminating background
     # processes like the coordinator and workers).
     if [ "${type}" == "coord" ]; then
         srun -w $COORD --ntasks=1 --overlap \
 --container-image=${coord_image} \
+--container-remap-root \
 --export=ALL,JAVA_HOME=/usr/lib/jvm/jre-17-openjdk \
 --container-env=JAVA_HOME=/usr/lib/jvm/jre-17-openjdk \
```
kingcrimsontianyu, line 105: In my experience env vars will have surprising incorrect values when imported this way, and I think this is because pyxis does not actually support the syntax `--container-env=key=value`. The [official doc](https://github.com/nvidia/pyxis) only lists `--container-env=NAME[,NAME...]`, which means it only imports to the container existing host env vars. To assign new values, we may use:
- `--export=ALL,key=value`. The tricky part is that the value itself cannot container comma (e.g. `--queries=9,10,11` will cause trouble).
- Just define the env vars in the command string following `bash -c`.

After all, the `JAVE_HOME` variable is explicitly set in the `bash -lc` region below, so here its reference can be removed.

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -59,46 +65,70 @@ function validate_environment_preconditions {
...
+    fi
+
     # Coordinator runs as a background process, whereas we want to wait for cli
     # so that the job will finish when the cli is done (terminating background
     # processes like the coordinator and workers).
     if [ "${type}" == "coord" ]; then
         srun -w $COORD --ntasks=1 --overlap \
 --container-image=${coord_image} \
+--container-remap-root \
 --export=ALL,JAVA_HOME=/usr/lib/jvm/jre-17-openjdk \
```
kingcrimsontianyu, line 104: The `JAVE_HOME` variable is explicitly set in the `bash -lc` region below, so here its assignment can be removed.

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -59,46 +65,70 @@ function validate_environment_preconditions {
...
 --container-env=JAVA_HOME=/usr/lib/jvm/jre-17-openjdk \
 --container-env=PATH=/usr/lib/jvm/jre-17-openjdk/bin:$PATH \
```
kingcrimsontianyu, line 106: This line can also be removed.

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -59,46 +65,70 @@ function validate_environment_preconditions {
...
--- bash -lc "unset JAVA_HOME; export JAVA_HOME=/usr/lib/jvm/jre-17-openjdk; export PATH=/usr/lib/jvm/jre-17-openjdk/bin:\$PATH; ${script}" >> ${LOGS_DIR}/${log_file} 2>&1 &
+${VT_ROOT}/.hive_metastore:/var/lib/presto/data/hive/metastore${extra_mounts} \
+-- bash -lc "unset JAVA_HOME; export JAVA_HOME=/usr/lib/jvm/jre-17-openjdk; export PATH=/usr/lib/jvm/jre-17-openjdk/bin:\$PATH; ${script}" >> ${LOGS}/${log_file} 2>&1 &
     else
         srun -w $COORD --ntasks=1 --overlap \
+--container-remap-root \
 --container-image=${coord_image} \
 --export=ALL,JAVA_HOME=/usr/lib/jvm/jre-17-openjdk \
 --container-env=JAVA_HOME=/usr/lib/jvm/jre-17-openjdk \
 --container-env=PATH=/usr/lib/jvm/jre-17-openjdk/bin:$PATH \
```
kingcrimsontianyu, line 122: Same as above. `JAVE_HOME` can be removed from `export`, and the two `container-env` lines can be removed.

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -174,45 +209,160 @@ function run_worker {
...
 ${worker_hive}:/opt/presto-server/etc/catalog/hive.properties,\
 ${worker_data}:/var/lib/presto/data,\
 ${DATA}:/var/lib/presto/data/hive/data/user_data,\
-${VT_ROOT}/.hive_metastore:/var/lib/presto/data/hive/metastore \
---container-env=LD_LIBRARY_PATH="$CUDF_LIB:$LD_LIBRARY_PATH" \
+${VT_ROOT}/.hive_metastore:/var/lib/presto/data/hive/metastore,\
+/usr/lib/aarch64-linux-gnu/libcuda.so.580.105.08:/usr/local/cuda-13.0/compat/libcuda.so.1,\
+/usr/lib/aarch64-linux-gnu/libnvidia-ml.so.580.105.08:/usr/local/lib/libnvidia-ml.so.1 \
 --container-env=GLOG_vmodule=IntraNodeTransferRegistry=3,ExchangeOperator=3 \
 --container-env=GLOG_logtostderr=1 \
```
kingcrimsontianyu, line 240: In https://github.com/rapidsai/velox-testing/pull/299, I removed `container-env` and moved `GLOG_vmodule=IntraNodeTransferRegistry=3,ExchangeOperator=3,GLOG_logtostderr=1` to the `bash -c` region to ensure they are in effect at runtime. However, this resulted in extremely large worker logs being generated that fill up the head node disk quickly, so I temporarily commented them out. I think for now we need to reconsider the best values for these env vars.

presto/slurm/presto-nvl72/launch-run.sh
```diff
@@ -33,8 +37,16 @@ SCALE_FACTOR=""
 NUM_ITERATIONS="2"
 EXTRA_ARGS=()
 NUM_GPUS_PER_NODE="4"
-WORKER_IMAGE="presto-native-worker-gpu"
-COORD_IMAGE="presto-coordinator"
+USE_NUMA="1"
+VARIANT_TYPE="gpu"
+#WORKER_IMAGE="presto-native-worker-gpu"
+COORD_IMAGE="presto-coordinator-karth-Mar11"
+WORKER_IMAGE="presto-native-worker-gpu-karth-Mar11"
```
kingcrimsontianyu, line 44: Fyi, I added nsys to this image and created a new squashFS file:
```
WORKER_IMAGE="presto-native-worker-gpu-karth-Mar11-with-nsys-2025.5.1"
```

benchmark_reporting_tools/post_results.py
```diff
@@ -93,10 +93,12 @@ class BenchmarkMetadata:
     kind: str | None = None
     execution_number: int = 1
     worker_count: int | None = None
+    node_count: int | None = None
```
paul-aiyedun, line 96: Would it make sense to also capture the number of workers per node?

benchmark_reporting_tools/post_results.py
```diff
@@ -93,10 +93,12 @@ class BenchmarkMetadata:
...
     scale_factor: int | None = None
     gpu_count: int | None = None
     num_drivers: int | None = None
     gpu_name: str | None = None
+    image_digest: str | None = None
```
paul-aiyedun, line 101: I believe, by default, the images are deleted after 30 days. Is this information that we want to persist in the database?

benchmark_reporting_tools/post_results.py
```diff
@@ -449,6 +476,16 @@ def _query_sort_key(name: str):
         if v is not None
     }
 
+    engine_config_payload = engine_config.serialize() if engine_config else {}
+    if any(v is not None for v in (velox_branch, velox_repo, presto_branch, presto_repo)):
```
paul-aiyedun, line 480: Nit: Does `if velox_branch or velox_repo or presto_branch or presto_repo:` do the same thing? If so, I think that would be a simpler way of writing this.

presto/scripts/generate_presto_config.sh
```diff
@@ -155,7 +158,7 @@ fi
 
 # We want to propagate any changes from the original worker config to the new worker configs even if
 # we did not re-generate the configs.
-if [[ -n "$NUM_WORKERS" && "$VARIANT_TYPE" == "gpu" ]]; then
+if [[ -n "$NUM_WORKERS" && ( "$VARIANT_TYPE" == "gpu" || "$VARIANT_TYPE" == "cpu" ) ]]; then
```
paul-aiyedun, line 161: Remove the `VARIANT_TYPE` check?

presto/slurm/presto-nvl72/defaults.env
```diff
@@ -0,0 +1,36 @@
...
+# --- Workspace identifier (per-clone isolation) ---
+_vt_path="${VT_ROOT:-$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." 2>/dev/null && pwd)}"
+: "${VT_WORKSPACE:=$(echo "${_vt_path#${HOME}/}" | tr '/' '-')}"
+unset _vt_path
+
+# --- Scratch-based paths ---
+# DATA points at the canonical TPC-H parquet tree on this cluster (hardcoded
+# to prestouser's scratch because that's where the shared dataset lives).
+# Override DATA in the env to point at a different layout.
+: "${DATA:=/scratch/prestouser/tpch-rs-float-no-delta}"
```
paul-aiyedun, line 22: Why are we using different environment variables from those used in the repo e.g. `PRESTO_DATA_DIR`?

presto/slurm/presto-nvl72/enroot-decompress.sh
```diff
@@ -0,0 +1,27 @@
...
+# Auto-detecting decompressor for enroot image layer downloads.
+# Used as ENROOT_GZIP_PROGRAM to support both gzip and OCI tar+zstd layers.
+# Called by enroot as: enroot-decompress.sh -d -f -c  (args are ignored)
+
+tmp=$(mktemp)
+trap 'rm -f "$tmp"' EXIT
+
+# Peek at first 4 bytes to detect compression format without buffering the full stream
+dd bs=1 count=4 2>/dev/null > "$tmp"
+magic=$(od -A n -N 4 -t x1 "$tmp" | tr -d ' \n')
```
paul-aiyedun, line 11: Did you consider using the `file` command?

presto/slurm/presto-nvl72/gen-tpch-data.slurm
```diff
@@ -0,0 +1,64 @@
...
+
+# ==============================================================================
+# Defaults
+# ==============================================================================
+: "${SCALE_FACTOR:=100}"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${SCRIPT_DIR}/defaults.env"
+: "${OUTPUT_DIR:=/scratch/${USER}/${VT_WORKSPACE}/tpch-rs-float/scale-100-no-delta}"
+: "${PARALLELISM:=100}"
+: "${IMAGE:=ghcr.io/tomaugspurger/tpchgen-rs:tom-sync-upstream-clean}"
```
paul-aiyedun, line 32: Is there a reason for not using the data generation tool in velox-testing for this?

presto/slurm/presto-nvl72/get_avg_col.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+awk -F'|' '
```
paul-aiyedun, line 5: Can you please expand on the use case for this script (it does not appear to be referenced by other scripts added in this PR)?

The same question applies to the `get_luke_col.sh` script.

presto/slurm/presto-nvl72/run-analyze-tables.slurm
```diff
@@ -0,0 +1,108 @@
...
+# ==============================================================================
+export NUM_NODES=$SLURM_JOB_NUM_NODES
+export COORD=$(scontrol show hostnames $SLURM_JOB_NODELIST | head -n 1)
+export NUM_WORKERS=$((NUM_NODES * NUM_GPUS_PER_NODE))
+
+# Presto configuration
+export PORT=9200
+export CUDF_LIB=/usr/lib64/presto-native-libs
+
+# UCX Configuration
```
paul-aiyedun, line 82: Is analyze executed with UCX?

presto/slurm/presto-nvl72/run-sweep.sh
```diff
@@ -0,0 +1,99 @@
...
+CACHE_STATE="warm"
+VELOX_BRANCH="ibm-research-preview_2026_03_03_pr16201_pr16488"
+PRESTO_BRANCH="ibm-research-preview_2026_03_03_and_fixes_and_PR27215"
+VELOX_REPO="https://github.com/karthikeyann/velox"
+PRESTO_REPO="https://github.com/karthikeyann/presto"
+
+#VELOX_BRANCH="ibm-research-preview_2026_03_11"
+#PRESTO_BRANCH="ibm-research-preview_2026_03_11"
+#VELOX_REPO="https://github.com/IBM/velox"
+#PRESTO_REPO="https://github.com/prestodb/presto"
```
paul-aiyedun, line 41: Why are we hardcoding these values?

presto/slurm/presto-nvl72/launch-run.sh
```diff
@@ -33,8 +37,16 @@ SCALE_FACTOR=""
 NUM_ITERATIONS="2"
 EXTRA_ARGS=()
 NUM_GPUS_PER_NODE="4"
-WORKER_IMAGE="presto-native-worker-gpu"
-COORD_IMAGE="presto-coordinator"
+USE_NUMA="1"
+VARIANT_TYPE="gpu"
+#WORKER_IMAGE="presto-native-worker-gpu"
+COORD_IMAGE="presto-coordinator-karth-Mar11"
```
paul-aiyedun, line 43: Should this value be coming from an environment variable or parameter?

presto/slurm/presto-nvl72/run-analyze-tables.sh
paul-aiyedun, line 1: Instead of adding new scripts like this with similar functions as existing scripts, did you consider extending the existing scripts?

paul-aiyedun (commented):
Changes overall make sense to me. However, I had a number of questions.

kingcrimsontianyu:
Compared to the superseded PR, there is no [validation_results.py](https://github.com/rapidsai/velox-testing/pull/273/files#diff-306fe5fea4abc421f7f8dd50f3e492505e76d89f812405c49913c9bca0c32993), should we add it back?

presto/slurm/presto-nvl72/functions.sh
misiugodfrey: I've removed them, but left them in a comment above in case folks want to turn them on again.

presto/slurm/presto-nvl72/functions.sh
misiugodfrey: Makes sense.  I've removed this syntax.

presto/slurm/presto-nvl72/launch-run.sh
misiugodfrey: That's good to hear; although I imagine in most cases folks are going to profile an arbitrary image.  Once work is done to fully capture a profile, perhaps  we should add an option to install nsys in a given image so that any image can be profiled?

benchmark_reporting_tools/post_results.py
misiugodfrey: I'm not sure there's much use for a "number of workers per node" field since there is already a "number of workers" and "number of nodes"?  Unless there was some use-case for an uneven number of workers per node (not currently a supported use-case), then I think that value should always be derivable from the existing information.

benchmark_reporting_tools/post_results.py
misiugodfrey: Right now the Benchmarking DB requires an identifier for run submissions and we default to the image_digest.  I think even if the image is deleted this is useful for identifying an image uniquely.

benchmark_reporting_tools/post_results.py
misiugodfrey: I think there is a difference in the semantics if these values contain empty strings; but if the user is explicitly putting those in then perhaps  we should accept it.

presto/scripts/generate_presto_config.sh
misiugodfrey: Makes sense.  While we haven't tested our multi-node java-worker case, there's no reason it should be excluded.

presto/slurm/presto-nvl72/defaults.env
misiugodfrey: PRESTO_DATA_DIR and DATA are both used in the slurm scripts to refer to different things because of how the standard scripts expect them to be used.

DATA is the variable that posts to the host-side path (outside the containers) that gets bound to /user_data inside the container. This is understandably confusing, because this is what PRESTO_DATA_DIR does in the docker environment.

However, because scripts that rely on PRESTO_DATA_DIR are run in the slurm scripts inside containers, that value needs to be bound to the same value as the container sees. So PRESTO_DATA_DIR is hard-coded in these scripts to user_data.

This difference comes up because in the slurm environment, everything - including the scripts that would otherwise be run by the host - are run in containers. So in this context PRESTO_DATA_DIR needs to be an in-container path instead of a host-path; hence the need for a new host variable.

presto/slurm/presto-nvl72/enroot-decompress.sh
misiugodfrey: I did not experiment with the `file` command.  From what I looked up, there can be some variation in what it may output in this case, e.g. "gzip compressed data" vs "application/gzip".  We could probably grep through the output - since it will likely always contain "gzip", but I think the `od` version will always return the same bytes regardless.

That being said, I'm not sure just how variable `file` is, just that it can vary slightly.  Do you think the tradeoff is worth clarify in the source?

presto/slurm/presto-nvl72/gen-tpch-data.slurm
misiugodfrey: At time of writing I was under the impression that Tom's fork was our official datagen version.   Now that has changed I can update this to us our local one.

That being said, the local scripts require docker to install the fork, which is not available on these systems.   I have pushed a version of the required container to our velox-testing registry, so we can pull that instead and run it on our systems.

presto/slurm/presto-nvl72/get_avg_col.sh
misiugodfrey: These were scripts I was using to more easily copy output data into the spreadsheets we were using to track benchmarks (they re-formatted the relevant data for easy data-entry).  Now that we have the benchmarking DB I'll remove them.

presto/slurm/presto-nvl72/run-analyze-tables.slurm
misiugodfrey: No.  This was just a copy of the configuration since we were using the same images but with `cudf.enabled=false`.  I don't believe the UCX settings would come into play in this context, but I'll remove this to prevent confusion.

presto/slurm/presto-nvl72/run-sweep.sh
misiugodfrey: I'll replace them with required parameters for now.  Ideally I want these values to be auto-detected from the image so that a user can't accidentally report the wrong values - there is a PR up to store that information in our images - but until that comes these should be params.

presto/slurm/presto-nvl72/launch-run.sh
misiugodfrey: I've made these required params.

presto/slurm/presto-nvl72/run-analyze-tables.sh
misiugodfrey: While these scripts try to re-use the internal functions to accomplish the same logic; there is a lot of logic that isn't currently share-able between the slurm and non-slurm versions.  Specifically stuff like determining if workers and coordinators are up work differently.
We may be able to bring those closer together in the future, but it would require changes in the base versions that I think is better saved as future work.
That being said, these scripts do re-use the main functionality - such as setup_benchmark_tables.sh and analyze_tables.sh, so I don't think we are at risk of the semantics drifting apart.  It's just the node-management that's different due to the different context.

misiugodfrey:
> Compared to the superseded PR, there is no [validation_results.py](https://github.com/rapidsai/velox-testing/pull/273/files#diff-306fe5fea4abc421f7f8dd50f3e492505e76d89f812405c49913c9bca0c32993), should we add it back?

I did not include that validation in this PR so that it doesn't block this work (as the validation work is ongoing here: https://github.com/rapidsai/velox-testing/pull/275).  Once the validation PR lands, we can turn it on in the cluster scripts very easily, since it goes through the same paths as part of run_benchmark.sh (in fact, I think it should be pulled in automatically).

kingcrimsontianyu (approved):
Looks good!

benchmark_reporting_tools/post_results.py
paul-aiyedun: For instance, if we have 6 workers and 2 nodes, is there a guarantee that each node would have 3 workers, or can it be 4 workers on one node and 2 on another node (assuming each node has 4 GPUs)?

presto/slurm/presto-nvl72/enroot-decompress.sh
paul-aiyedun: I think the current implementation is fine. I just wanted to know if this was weighed against the `file` option.

benchmark_reporting_tools/post_results.py
misiugodfrey: Right now it's guaranteed to be an even split.  In CPU-mode you always get one worker per node.  In GPU-mode you always get `NUM_GPUS_PER_NODE` workers per node (default 4).

The API to the launch script requires you to specify number of nodes (-n) and optionally number of gpus (workers) per node (-g).  So the number of workers is dynamically calculated based on this and guarantees an even worker split among nodes.

paul-aiyedun (approved):

PR #314, opened 2026-04-21
Add support for data generation column encoding and compression override

paul-aiyedun:
* Add per-table, per-column codec control to the TPC-H data generation script.
* Auto-generate default codec definitions when no file is supplied by introspecting the TPC-H schema at a small scale factor:
  * Integers: DELTA_BINARY_PACKED, dictionary disabled.
  * High-cardinality strings: PLAIN, dictionary disabled.
  * Everything else: tpchgen-cli defaults.
* Skip row group size-tolerance checks for small tables (< 4 row groups) and tighten tolerance from 25% to 20%.
* Add codec_definitions_test.py covering default codec application, custom file loading with encoding/compression assertions, and validation error cases.
* Add codec_definition_template.json and test fixture JSON files

---

The following experiment was done to determine the best default encoding scheme to use:

## Experiment Summary

This experiment evaluates the impact of different Parquet column encoding and compression strategies on TPC-H query performance. The goal is to determine whether alternative encoding and compression schemes can improve query latency compared to the current baseline configuration used by `tpchgen-cli` and the [tpchgen-rs fork](https://github.com/TomAugspurger/tpchgen-rs/blob/a2419caad8cfc33f66d9213337592ce14ad3c938/scripts/generate_tpch.py).

All datasets were generated at TPC-H Scale Factor 1000 with decimal columns converted to 64-bit floats (`--convert-decimals-to-floats`). Each experiment variant was benchmarked by running all 22 TPC-H queries for 10 iterations, with Hive schemas dropped and recreated between variants to eliminate caching effects. The queries were executed on a Presto GPU (Velox-cuDF) instance with 4 workers. Each worker ran on an NVIDIA H200 GPU.

## Results (10 Iterations)

| # | Experiment Description | Encoding | Compression | Dictionary | Dataset Size | Median Total (ms) | vs Baseline |
|---|---|---|---|---|---|---|---|
| 1 | **Baseline**: DELTA_BINARY_PACKED for integer keys (NDV >= 0.01), PLAIN elsewhere, no compression, dictionary off for unique strings (NDV >= 0.99) | Mixed | UNCOMPRESSED | Selective | 345 GB | 51,157 | -- |
| 2 | **Upstream defaults**: No codec overrides, uses tpchgen-cli built-in defaults | Default | Default | Default | 343 GB | 51,789 | +1.2% |
| 3 | **Snappy compression only**: Baseline encodings with Snappy compression applied to all columns | Baseline | SNAPPY | Baseline | 308 GB | 47,226 | **-7.7%** |
| ~~4~~ | ~~**ZSTD compression only**: Baseline encodings with ZSTD compression applied to all columns~~ | ~~Baseline~~ | ~~ZSTD~~ | ~~Baseline~~ | ~~308 GB~~ | ~~48,733~~ | ~~**-4.7%**~~ |
| 5 | **No compression, no encoding**: All columns PLAIN + UNCOMPRESSED, dictionary off | PLAIN | UNCOMPRESSED | Off | 593 GB | 52,822 | +3.3% |
| 6 | **No encoding, default compression**: All columns forced to PLAIN encoding with dictionary off, default compression | PLAIN | Default | Off | 442 GB | 62,786 | +22.7% |
| 7 | **All dictionary, no compression**: Dictionary enabled for all columns (including high-cardinality, NDV >= 0.01 for integers / >= 0.5 for strings), no encoding overrides, UNCOMPRESSED. †Q9, Q21 failed (INTERNAL_ERROR); dictionary overflow on high-cardinality columns falls back to PLAIN, inflating dataset | Default | UNCOMPRESSED | On (all) | 431 GB | 60,572† | +49%† |
| 8 | **Type-based encoding v1**: DELTA_BINARY_PACKED for all integers, PLAIN for strings/decimals/dates, dictionary on for low-cardinality (NDV < 0.01), off for high-cardinality | Type-based | Default | Cardinality-based | 319 GB | 47,557 | **-7.0%** |
| 9 | **Type-based encoding v2**: DELTA_BINARY_PACKED for all integers (dictionary off), PLAIN for strings/decimal/float (dictionary off), dates retain default dictionary. †Q9, Q21 failed (INTERNAL_ERROR) | Type-based | Default | Dates only | 331 GB | 51,195† | +26%† |
| ~~10~~ | ~~**Type-based encoding + ZSTD**: Type-based v1 encoding combined with ZSTD compression~~ | ~~Type-based~~ |~~ZSTD~~ | ~~Cardinality-based~~ | ~~319 GB~~ | ~~47,899~~ | ~~**-6.4%**~~ |
| 11 | **Type-based encoding + Snappy**: Type-based v1 encoding combined with Snappy compression | Type-based | SNAPPY | Cardinality-based | 319 GB | 48,451 | **-5.3%** |
| ~~12~~ | ~~**Baseline encoding + ZSTD**: Original baseline encodings with ZSTD compression~~ | ~~Baseline~~ | ~~ZSTD~~ | ~~Baseline~~ | ~~308 GB~~ | ~~47,454~~ | ~~**-7.2%**~~ |
| 13 | **Snappy, favor dictionary**: Dictionary on for all strings unless NDV ratio >= 0.5, DELTA_BINARY_PACKED for integer keys (NDV >= 0.01), Snappy | Mixed | SNAPPY | Favor dict | 306 GB | 47,000 | **-8.1%** |
| 14 | **Snappy, favor dict + delta int**: Same as #13 but DELTA_BINARY_PACKED for ALL integer columns, dictionary off for integers | Mixed | SNAPPY | Favor dict | 294 GB | 46,484 | **-9.1%** |
| 15 | **Snappy, dict non-unique + delta int**: Same as #14 but only dictionary encode strings with NDV ratio < 0.99 (truly non-unique). Best result | Mixed | SNAPPY | Selective | 294 GB | 45,618 | **-10.8%** |
| 16 | **Snappy, dict 80% + delta int**: Same as #14 but dictionary encode strings only when NDV ratio < 0.80, off for near-unique and unique strings | Mixed | SNAPPY | Selective | 294 GB | 47,188 | **-7.8%** |
| 17 | **generate_tpch.py defaults**: Exact per-column reproduction of `generate_tpch.py` hardcoded settings (--no-delta-length-byte-array, --use-float-type): DELTA_BINARY_PACKED for integer keys, PLAIN for all strings, mixed SNAPPY/UNCOMPRESSED compression, dictionary off for most columns. †Q9, Q21 failed (INTERNAL_ERROR — CUDA OOM) | Mixed | Mixed | Mostly off | 294 GB | 48,849† | +23.7%† |

**†** Experiments 07, 09, and 17 completed 20/22 queries; Q9 and Q21 failed with INTERNAL_ERROR on the GPU workers. The "Median Total" and "vs Baseline" values are computed using only the 20 successful queries, compared against the same 20 queries from the baseline. Not directly comparable to full 22-query totals.

PR Revision Update:
The ZSTD related experiments have been invalidated due to a bug in the initial version of the PR where compression options besides UNCOMPRESSED were ignored and as a result, SNAPPY compression was always used in this case.


benchmark_data_tools/generate_data_files.py
```diff
@@ -197,6 +216,126 @@ def get_column_projection(column_metadata, convert_decimals_to_floats):
...
+    if not columns:
+        return []
+
+    args = []
+
+    uncompressed_cols = [
+        column["name"] for column in columns if column.get("compression", "").upper() == "UNCOMPRESSED"
+    ]
+    if uncompressed_cols:
+        args.append(f"--uncompressed-column-overrides={','.join(uncompressed_cols)}")
```
mattgara, line 258: Hmmm, AFAICT, based on 

https://github.com/rapidsai/velox-testing/pull/314/changes#diff-681da63d4a800f91a862e5e674a49851b158885022be118d8b7fc6effa45f59fR9

`SNAPPY` and `ZSTD` are supposed to be supported as well? It would seem this only supports `UNCOMPRESSED`?

benchmark_data_tools/generate_data_files.py
```diff
@@ -197,6 +216,126 @@ def get_column_projection(column_metadata, convert_decimals_to_floats):
...
+    """
+    conn = duckdb.connect()
+    conn.execute(f"INSTALL tpch; LOAD tpch; CALL dbgen(sf = {_SAMPLE_SF});")
+
+    tables = [row[0] for row in conn.execute("SHOW TABLES").fetchall()]
+    config = {"tables": []}
+
+    for table in sorted(tables):
+        total_rows = conn.execute(f"SELECT COUNT(*) FROM {table}").fetchone()[0]
+        assert total_rows > 0, f"Table '{table}' has no rows at SF {_SAMPLE_SF}"
```
mattgara: Consider using `raise` instead of `assert`, as it will likely behave better in the wild.

benchmark_data_tools/generate_data_files.py
```diff
@@ -48,13 +53,26 @@ def generate_partition(
     if convert_decimals_to_floats:
         command.extend(["--decimal-column-type", "f64"])
 
+    command.extend(get_tpchgen_codec_args(codec_defs, table))
+
     try:
         subprocess.run(command, check=True)
     except subprocess.CalledProcessError as e:
         print(f"Error generating TPC-H data: {e}")
```
mattgara: Should this continue to raise the exception and/or cleanup? It looks like we could get partially complete data if an exception occurs.

benchmark_data_tools/generate_data_files.py
```diff
@@ -197,6 +216,126 @@ def get_column_projection(column_metadata, convert_decimals_to_floats):
...
+                            "name": col_name,
+                            "encoding": "PLAIN",
+                            "dictionary": False,
+                        }
+                    )
+
+        if col_entries:
+            config["tables"].append({"name": table, "columns": col_entries})
+
+    conn.close()
```
mattgara: Consider putting `close` in a `finally`

mattgara (commented):
LGTM. Should we have the `generate_data_files.py` store details on what encoding/compressions were used in `metadata.json`?

misiugodfrey (approved):

benchmark_data_tools/generate_data_files.py
paul-aiyedun: Correct. I have made an update to raise an exception.

benchmark_data_tools/generate_data_files.py
paul-aiyedun: Good catch. There was a bug where other compression types were ignored. That should be fixed in the current PR revision. Also, the compression option has been moved to the table level.

benchmark_data_tools/generate_data_files.py
paul-aiyedun: Updated.

benchmark_data_tools/generate_data_files.py
paul-aiyedun: Updated to use a `with` statement.

paul-aiyedun:
> LGTM. Should we have the `generate_data_files.py` store details on what encoding/compressions were used in `metadata.json`?

Comprehensive changes to `metadata.json` will be applied in a subsequent PR (there are other updates besides the changes in this PR).

mattgara (approved):
LGTM, thanks.

PR #315, opened 2026-04-21
Remove privileged from docker-compose template

misiugodfrey:
Ran into an odd error where gpu worker images were failing to come up on a host with a bad (broken?) gpu.

The failing GPU was detected by the driver but ignored by `nvidia-smi`.  It looks like the docker-compose using `--privileged` was causing the broken gpu to be mounted in the container and `cuInit` would run and fail before it could be filtered by CUDA_VISIBLE_DEVICES (which is how we assign each container to use a different GPU).

This change prevents that issue, while still allowing the privileges necessary for NUMA binding (which I believe was the original impetus for the --privileged option).


paul-aiyedun (approved):

karthikeyann (approved):
Looks good to me. 
added @kingcrimsontianyu for final approval

bdice (approved):

presto/docker/docker-compose/template/docker-compose.native-gpu.yml.jinja
```diff
@@ -16,6 +16,7 @@ x-presto-native-worker-gpu: &gpu_worker_base
   cap_add:
     - IPC_LOCK
     - SYS_ADMIN  # Required for nsys GPU metrics profiling
+    - SYS_NICE   # Required for numactl --cpunodebind
```
kingcrimsontianyu: Sorry for the late review. Interesting problem and solution. 
I checked both the Docker and Linux numactl documentations, and I'm not quite confident that NUMA binding requires SYS_NICE. numactl sets its own process affinity and then replaces its own process image with the program launched. I think this is orthogonal to setting the process "nice" priority value.
I guess the original impetus for using `privileged` might be dropping system-wide page cache, and enabling certain options in nsys profile such as `--gpu-metrics-devices`.

kingcrimsontianyu: PS: Just noticed we have `SYS_ADMIN` already. That should be sufficient for `--gpu-metrics-devices`.

kingcrimsontianyu: I made a separate test with Docker, and can confirm the following:

- NUMA control with `numactl` does not require special capability such as `SYS_NICE`, or elevated privilege.
- System-wide page cache dropping requires elevated privilege. However, I believe our [current implementation](https://github.com/rapidsai/velox-testing/blob/main/common/testing/performance_benchmarks/cache_utils.py#L14) is to do that in a separate, unrelated container, which has nothing to do with the docker configuration files here.

So I think here we don't need `SYS_NICE` and can simply drop `privileged: true`. @misiugodfrey @karthikeyann @paul-aiyedun

misiugodfrey: My own testing found that the worker node crashes if I remove the SYS_NICE privilege from the docker container.  Specifically, the `launch_presto_servers.sh` entrypoint will fail on this call: `numactl --cpunodebind=0 --membind=0 fails with "numactl: This system does not support NUMA policy"`

From my understanding, docker's default seccomp profile will block calls like `mbind` unless the container has SYS_NICE privileges.  I think not all numactl invocations require such a privilege, `numactl --interleave --all` may work, for instance, but the particular binding that we currently use needs it.  I believe that `mbind` is the  main syscall that we are going through that requires these privileges: https://docs.docker.com/engine/security/seccomp/#:~:text=mbind,by%20CAP_SYS_NICE.

kingcrimsontianyu: I was testing a toy code in a devcontainer without the NICE permission. Perhaps this setup is more permissive. Great to learn about the actual Presto experiment and the official doc.

kingcrimsontianyu (approved):

PR #316, opened 2026-04-22
Add image labels

misiugodfrey:
This change updates our CI and build scripts to add labels into our docker  images to track which presto/velox branches the images were built with.  This is particularly useful since the scripts expect image names and tags to be uniform, so it can be easy to lose track of how an image was originally built.

This will be especially useful for benchmarking, as the post_results.py script can reference this to make sure we track exactly what was used to build an image that was used for a benchmark (follow-up PR).

The way these labels are filled is slightly different between the CI and a local build, only because the CI references SHAs only, whereas a local build will also reference the branch the CI belongs to for ease of use.

Example docker inspect on worker image (from CI):

```
"Labels": {
                ...
                "velox-testing.presto.branch": "",
                "velox-testing.presto.repository": "prestodb/presto",
                "velox-testing.presto.sha": "dae4091c7105d1370544bfcb11baf8f8a79256bc",
                "velox-testing.velox.branch": "",
                "velox-testing.velox.repository": "facebookincubator/velox",
                "velox-testing.velox.sha": "948e7be0619f8d123ae9d8fbd8a4f7b928492aec"
            }
```
Coord image (CI):
```
"Labels": {
                 ...
                "velox-testing.presto.branch": "",
                "velox-testing.presto.repository": "prestodb/presto",
                "velox-testing.presto.sha": "dae4091c7105d1370544bfcb11baf8f8a79256bc",
                "velox-testing.velox.branch": "",
                "velox-testing.velox.repository": "",
                "velox-testing.velox.sha": ""
            }
```
Deps image (Local):
```
"Labels": {
                ...
                "velox-testing.presto.branch": "master",
                "velox-testing.presto.repository": "https://github.com/prestodb/presto.git",
                "velox-testing.presto.sha": "91e9132db5abcbfb0d7ab2232602784c5f02d5b4",
                "velox-testing.velox.branch": "main",
                "velox-testing.velox.repository": "https://github.com/facebookincubator/velox.git",
                "velox-testing.velox.sha": "7ed90a53dee9e640cba92a644dae85fe52b07fb9"
}
```
Worker image (Local):
```
            "Labels": {
                ...
                "velox-testing.presto.branch": "master",
                "velox-testing.presto.repository": "https://github.com/prestodb/presto.git",
                "velox-testing.presto.sha": "91e9132db5abcbfb0d7ab2232602784c5f02d5b4",
                "velox-testing.velox.branch": "main",
                "velox-testing.velox.repository": "https://github.com/facebookincubator/velox.git",
                "velox-testing.velox.sha": "7ed90a53dee9e640cba92a644dae85fe52b07fb9"
}
```

### DIFF
--- a/.github/workflows/presto-build.yml
+++ b/.github/workflows/presto-build.yml
@@ -335,13 +335,12 @@ jobs:
             ENABLE_SCCACHE=ON
             GPU=${{ matrix.gpu_flag }}
             NUM_THREADS=31
-          labels: |
-            velox-testing.presto.sha=${{ inputs.presto_sha }}
-            velox-testing.presto.branch=
-            velox-testing.presto.repository=${{ inputs.presto_repository }}
-            velox-testing.velox.sha=${{ inputs.velox_sha }}
-            velox-testing.velox.branch=
-            velox-testing.velox.repository=${{ inputs.velox_repository }}
+            PRESTO_SHA=${{ inputs.presto_sha }}
+            PRESTO_BRANCH=
+            PRESTO_REPOSITORY=${{ inputs.presto_repository }}
+            VELOX_SHA=${{ inputs.velox_sha }}
+            VELOX_BRANCH=
+            VELOX_REPOSITORY=${{ inputs.velox_repository }}
           secret-files: |
             github_token=${{ env.SCCACHE_AUTH_DIR }}/github_token
             aws_credentials=${{ env.SCCACHE_AUTH_DIR }}/aws_credentials
--- a/.github/workflows/presto-build.yml
+++ b/.github/workflows/presto-build.yml
@@ -135,6 +135,11 @@ jobs:
           build-args: |
             CUDA_VERSION=${{ matrix.cuda_version }}
             ARM_BUILD_TARGET=generic
+          labels: |
+            velox-testing.presto.sha=${{ inputs.presto_sha }}
+            velox-testing.presto.repository=${{ inputs.presto_repository }}
+            velox-testing.velox.sha=${{ inputs.velox_sha }}
+            velox-testing.velox.repository=${{ inputs.velox_repository }}
           cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:cache-presto-deps-cuda${{ matrix.cuda_version }}-${{ matrix.arch }}
           cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:cache-presto-deps-cuda${{ matrix.cuda_version }}-${{ matrix.arch }},mode=max
           outputs: type=registry,compression=zstd,force-compression=true,compression-level=15
@@ -195,6 +200,11 @@ jobs:
           file: presto/docker/Dockerfile
           build-args: |
             PRESTO_VERSION=${{ inputs.presto_short_sha }}
+          labels: |
+            velox-testing.presto.sha=${{ inputs.presto_sha }}
+            velox-testing.presto.repository=${{ inputs.presto_repository }}
+            velox-testing.velox.sha=${{ inputs.velox_sha }}
+            velox-testing.velox.repository=${{ inputs.velox_repository }}
           outputs: type=registry,compression=zstd,force-compression=true,compression-level=15
           push: true
           tags: |
@@ -321,6 +331,11 @@ jobs:
             ENABLE_SCCACHE=ON
             GPU=${{ matrix.gpu_flag }}
             NUM_THREADS=31
+          labels: |
+            velox-testing.presto.sha=${{ inputs.presto_sha }}
+            velox-testing.presto.repository=${{ inputs.presto_repository }}
+            velox-testing.velox.sha=${{ inputs.velox_sha }}
+            velox-testing.velox.repository=${{ inputs.velox_repository }}
           secret-files: |
             github_token=${{ env.SCCACHE_AUTH_DIR }}/github_token
             aws_credentials=${{ env.SCCACHE_AUTH_DIR }}/aws_credentials
--- a/.github/workflows/presto-build.yml
+++ b/.github/workflows/presto-build.yml
@@ -137,8 +137,10 @@ jobs:
             ARM_BUILD_TARGET=generic
           labels: |
             velox-testing.presto.sha=${{ inputs.presto_sha }}
+            velox-testing.presto.branch=
             velox-testing.presto.repository=${{ inputs.presto_repository }}
             velox-testing.velox.sha=${{ inputs.velox_sha }}
+            velox-testing.velox.branch=
             velox-testing.velox.repository=${{ inputs.velox_repository }}
           cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:cache-presto-deps-cuda${{ matrix.cuda_version }}-${{ matrix.arch }}
           cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:cache-presto-deps-cuda${{ matrix.cuda_version }}-${{ matrix.arch }},mode=max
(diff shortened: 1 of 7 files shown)


paul-aiyedun (commented):
Can we add the labels directly in the Dockerfiles, so we can ensure that all build paths set the labels?

misiugodfrey:
> Can we add the labels directly in the Dockerfiles, so we can ensure that all build paths set the labels?

I've added support for that in the `native_build.dockerfile`, but the other dockerfiles are external.  So this will  hybridize between the two approaches now.

presto/scripts/build_centos_deps_image.sh
```diff
@@ -94,12 +94,32 @@ mkdir -p velox
...
+# Apply provenance labels; docker compose build --label is not universally supported.
+echo "Applying provenance labels..."
+echo "FROM ${COMPOSE_IMAGE_NAME}" | docker build --no-cache \
+  --label "velox-testing.presto.sha=${PRESTO_SHA}" \
+  --label "velox-testing.presto.branch=${PRESTO_BRANCH}" \
+  --label "velox-testing.presto.repository=${PRESTO_REPO}" \
+  --label "velox-testing.velox.sha=${VELOX_SHA}" \
+  --label "velox-testing.velox.branch=${VELOX_BRANCH}" \
+  --label "velox-testing.velox.repository=${VELOX_REPO}" \
+  -t "${COMPOSE_IMAGE_NAME}" -
```
paul-aiyedun: If we are not adding the labels to the original image, then I think we can put this logic in a Dockerfile. Also, does overriding the `$COMPOSE_IMAGE_NAME` image tag result in a dangling image? Do we need to clean up the original image after this build?

presto/scripts/start_presto_helper.sh
```diff
@@ -226,13 +226,52 @@ if (( ${#BUILD_TARGET_ARG[@]} )); then
...
+  )
+  for service in "${BUILD_TARGET_ARG[@]}"; do
+    case "$service" in
+      "$COORDINATOR_SERVICE") img="$COORDINATOR_IMAGE" ;;
+      "$JAVA_WORKER_SERVICE") img="$JAVA_WORKER_IMAGE" ;;
+      *) continue ;;
+    esac
+    echo "Applying provenance labels to ${img}..."
+    echo "FROM ${img}" | docker build --no-cache "${PROVENANCE_LABEL_ARGS[@]}" -t "${img}" -
+  done
```
paul-aiyedun, line 270: See above comment about using a Dockerfile.

presto/scripts/start_presto_helper.sh
```diff
@@ -226,13 +226,52 @@ if (( ${#BUILD_TARGET_ARG[@]} )); then
...
   docker compose --progress plain -f $DOCKER_COMPOSE_FILE_PATH build \
   $SKIP_CACHE_ARG --build-arg PRESTO_VERSION=$PRESTO_VERSION \
   --build-arg NUM_THREADS=$NUM_THREADS --build-arg BUILD_TYPE=$BUILD_TYPE \
   --build-arg CUDA_ARCHITECTURES=$CUDA_ARCHITECTURES \
+  --build-arg PRESTO_SHA="${PRESTO_SHA}" \
+  --build-arg PRESTO_BRANCH="${PRESTO_BRANCH}" \
+  --build-arg PRESTO_REPOSITORY="${PRESTO_REPO}" \
+  --build-arg VELOX_SHA="${VELOX_SHA}" \
+  --build-arg VELOX_BRANCH="${VELOX_BRANCH}" \
+  --build-arg VELOX_REPOSITORY="${VELOX_REPO}" \
```
paul-aiyedun: Consider moving these arguments to the docker compose file and making them required?

presto/docker/native_build.dockerfile
```diff
@@ -112,4 +112,17 @@ RUN mkdir /usr/lib64/presto-native-libs && \
 
 COPY velox-testing/presto/docker/launch_presto_servers.sh velox-testing/presto/docker/presto_profiling_wrapper.sh /opt
 
+ARG PRESTO_SHA=""
+ARG PRESTO_BRANCH=""
+ARG PRESTO_REPOSITORY=""
+ARG VELOX_SHA=""
+ARG VELOX_BRANCH=""
+ARG VELOX_REPOSITORY=""
```
paul-aiyedun: We should probably make these arguments required.

presto/scripts/build_centos_deps_image.sh
```diff
@@ -94,12 +94,32 @@ mkdir -p velox
 cp -r ../../velox/scripts velox
 cp -r ../../velox/CMake velox
 
+# Capture build provenance from sibling repos for embedding in Docker labels.
+PRESTO_SHA=$(git -C "${REPO_ROOT}/../presto" rev-parse HEAD 2>/dev/null || echo "")
+PRESTO_BRANCH=$(git -C "${REPO_ROOT}/../presto" rev-parse --abbrev-ref HEAD 2>/dev/null || echo "")
+PRESTO_REPO=$(git -C "${REPO_ROOT}/../presto" remote get-url origin 2>/dev/null || echo "")
+VELOX_SHA=$(git -C "${REPO_ROOT}/../velox" rev-parse HEAD 2>/dev/null || echo "")
+VELOX_BRANCH=$(git -C "${REPO_ROOT}/../velox" rev-parse --abbrev-ref HEAD 2>/dev/null || echo "")
+VELOX_REPO=$(git -C "${REPO_ROOT}/../velox" remote get-url origin 2>/dev/null || echo "")
```
paul-aiyedun: Consider moving this to a function that can be reused in `start_presto_helper.sh`.

.github/workflows/presto-build.yml
```diff
@@ -135,6 +135,13 @@ jobs:
           build-args: |
             CUDA_VERSION=${{ matrix.cuda_version }}
             ARM_BUILD_TARGET=generic
+          labels: |
```
paul-aiyedun, line 138: I believe the scripts that build the images should handle this based on the checked out repos. Why do we need to set the labels and build args here again?

paul-aiyedun (commented):
Changes overall look good to me. However, I had some code cleanup comments and a question about the workflow file updates.

presto/scripts/build_centos_deps_image.sh
misiugodfrey: I'll add a change to clean up the old image, as right now it is left dangling.

I don't know if we want to put this logic in a Dockerfile, since the original Dockerfile for this image is in the presto repo, not our own; unless you are suggesting we add a new Dockerfile to wrap it?

presto/scripts/start_presto_helper.sh
misiugodfrey: Similar problem as with deps.  The Dockerfile for coordinator/java-worker is outside of our repo/control.  We could add another wrapper Dockerfile for these, but I'm not sure if that's what you mean.

presto/scripts/start_presto_helper.sh
misiugodfrey: Sure; although I want to make sure they aren't required for cases where we are not building an image (just running an existing one).

presto/docker/native_build.dockerfile
misiugodfrey: Sounds good.

presto/scripts/build_centos_deps_image.sh
misiugodfrey: Moved to `common_funtions.sh`

.github/workflows/presto-build.yml
misiugodfrey: Right now our CI does not go through the same scripts (build_centos_deps_image.sh is not called - it calls `docker build buildx` directly).  I think this should be changed, but until those code-paths are unified we would need to change both.

presto/docker/docker-compose.common.yml
```diff
@@ -35,6 +35,12 @@ services:
       dockerfile: velox-testing/presto/docker/native_build.dockerfile
       args:
         - BASE_IMAGE=${DEPS_IMAGE:-presto/prestissimo-dependency:centos9-${USER}}
+        - PRESTO_SHA=${PRESTO_SHA:-}
+        - PRESTO_BRANCH=${PRESTO_BRANCH:-}
+        - PRESTO_REPOSITORY=${PRESTO_REPO:-}
+        - VELOX_SHA=${VELOX_SHA:-}
+        - VELOX_BRANCH=${VELOX_BRANCH:-}
+        - VELOX_REPOSITORY=${VELOX_REPO:-}
```
paul-aiyedun, line 43: Can we make these variables required?

paul-aiyedun (approved):
Besides the comment about required environment variables, changes look good to me.

presto/docker/docker-compose.common.yml
misiugodfrey: I think we should leave them required for now, or else we won't be able to use theses scripts to run images that were build without labels (i.e. anything built before this change).

PR #317, opened 2026-04-22
Snapshot logs and configs into benchmark output dir for isolation

misiugodfrey:
run_benchmark.sh now copies presto_logs/ and the generated config directory into the benchmark output dir at the end of each run.  post_results.py now reads exclusively from these bundled copies, removing the dependency on shared global paths that would be overwritten by subsequent runs.

This is mostly to avoid the issue where post_results would copy configs/logs based on the most recent run, even if we were pointing at different results directory.


benchmark_reporting_tools/post_results.py
```diff
@@ -589,16 +569,21 @@ async def _process_benchmark_dir(
...
     variant = _ENGINE_TO_VARIANT.get(benchmark_metadata.engine)
     if effective_config_dir is None:
         if variant:
-            effective_config_dir = _default_config_dir(variant)
-            if effective_config_dir:
-                print(f"  Auto-detected variant '{variant}' → config dir: {effective_config_dir}", file=sys.stderr)
+            bundled = benchmark_dir / "config" / variant
+            if bundled.is_dir():
+                effective_config_dir = bundled
+                print(f"  Using bundled config for variant '{variant}': {effective_config_dir}", file=sys.stderr)
```
paul-aiyedun, line 633: Why do we write this to `stderr`?

benchmark_reporting_tools/post_results.py
```diff
@@ -616,12 +601,13 @@ async def _process_benchmark_dir(
...
+    # Resolve logs directory: explicit override → bundled snapshot inside benchmark_dir
     effective_logs_dir = logs_dir
     if effective_logs_dir is None:
-        effective_logs_dir = _default_logs_dir()
-        if effective_logs_dir:
-            print(f"  Auto-detected logs dir: {effective_logs_dir}", file=sys.stderr)
+        bundled = benchmark_dir / "logs"
+        if bundled.is_dir():
+            effective_logs_dir = bundled
+            print(f"  Using bundled logs dir: {effective_logs_dir}", file=sys.stderr)
```
paul-aiyedun, line 663: Same question as above.

paul-aiyedun (approved):

benchmark_reporting_tools/post_results.py
misiugodfrey: I believe all of the diagnostics messages in this file are routed to `stderr` because `stdout` is used for the JSON payload (although I think that's only if the `--dry-run` option is specified).  @TomAugspurger  can probably comment further on the original intent.

I think changing this to use `stdout` for reporting would not affect any actual submission path, but these changes are in line with existing conventions of the file such as this: https://github.com/rapidsai/velox-testing/blob/d8d0990d695b7aa517f41ed4a787f0e4fac34d73/benchmark_reporting_tools/post_results.py#L489

TomAugspurger: I believe that's correct.

benchmark_reporting_tools/post_results.py
```diff
@@ -589,16 +569,21 @@ async def _process_benchmark_dir(
...
-    # Resolve config directory: explicit override → auto-detect from variant
+    # Resolve config directory: explicit override → bundled snapshot inside benchmark_dir
     effective_config_dir = config_dir
     variant = _ENGINE_TO_VARIANT.get(benchmark_metadata.engine)
     if effective_config_dir is None:
         if variant:
-            effective_config_dir = _default_config_dir(variant)
-            if effective_config_dir:
-                print(f"  Auto-detected variant '{variant}' → config dir: {effective_config_dir}", file=sys.stderr)
+            bundled = benchmark_dir / "config" / variant
```
mattgara, line 630: Apparently this is only valid python code if `benchmark_dir` and `variant` both override the `/` operator. Actually, we might be able to loosen that requirement to "at least one-of".

"config" is a python string that gets implicitly casted to a type to support the `/` operation.

My question here is, is this preferable to a more general solution like `os.path.join` which will handle strings without corner cases?

benchmark_reporting_tools/post_results.py
```diff
@@ -616,12 +601,13 @@ async def _process_benchmark_dir(
...
         engine_config = None
 
-    # Resolve logs directory: explicit override → auto-detect from repo
+    # Resolve logs directory: explicit override → bundled snapshot inside benchmark_dir
     effective_logs_dir = logs_dir
     if effective_logs_dir is None:
-        effective_logs_dir = _default_logs_dir()
-        if effective_logs_dir:
-            print(f"  Auto-detected logs dir: {effective_logs_dir}", file=sys.stderr)
+        bundled = benchmark_dir / "logs"
```
mattgara, line 660: Similar question here, we seem to be relying on the type of `benchmark_dir` to make this valid, will it work in all cases -- that is, assuming this is a parameter that is passed in, does the argument parsing know that the type of the `benchmark_dir` matters to form valid python code?

mattgara (commented):
Just had comments about the use of the `/` operator with what look to be like strings.

benchmark_reporting_tools/post_results.py
misiugodfrey: I think this is fine because `benchmark_dir` is explicitly typed as a `Path`, so the join operator should expect a `str` or another `Path` as the second argument (there are operators for both).

This is the same idom that is used on line 558 in the unchanged code, so I think it should be fine to leave it.

benchmark_reporting_tools/post_results.py
misiugodfrey: Same case, I think.  `benchmark_dir` is explicitly set as a `Path` type,  so it will always be valid.

mattgara (approved):

PR #319, opened 2026-04-23
Handle both flat and nested scale_factor in metadata.json

misiugodfrey:
Datasets with a newer metadata format nest scale_factor under an "options" key. Update get_scale_factor_from_file() to try the top-level key first and fall back to options.scale_factor, and update error messages in Presto and Spark Gluten test utils to document both formats.


devavret (approved):

paul-aiyedun (commented):
Please hold off on this. There is ongoing work to unify the metadata format.

misiugodfrey:
Closed because we want to run on a unified format.  If we run into this error when parsing the data it should be an indicator that the data should be called into question.

PR #320, opened 2026-04-23
Fix stop_presto.sh failing when PRESTO_IMAGE_TAG is not set

misiugodfrey:
After adding `:?` enforcement to docker-compose files, stop_presto.sh would fail immediately if PRESTO_IMAGE_TAG wasn't already in the environment.  stop_presto.sh was still executing correctly, but would end with an error.

Apply the same `USER:-latest` default used in start_presto_helper.sh and run_benchmark.sh.


kjmph (approved):

mattgara (approved):
LGTM

simoneves (approved):
Fixes the issue for me

PR #323, opened 2026-04-24
Add Parquet format/type overrides to match reference TPC-H dataset

kjmph:
Builds on #314's per-column encoding/compression overrides by adding
three dataset-wide knobs that determine the Parquet wire format and
the integer width of nationkey/regionkey columns. Required to
reproduce layouts used as reference datasets elsewhere, where a
mismatch between the generated schema and the CREATE TABLE DDL (int32
vs int64) causes silent casts, and where Parquet v2 Data Page format
is what downstream readers assume for the encodings #314 configures.

* Add --parquet-version {v1,v2} to generate_data_files.py, default v2.
  tpchgen-cli's default is v1 (broader compatibility). v2 emits Data
  Page V2 headers, which cuDF's Parquet reader relies on for efficient
  decode of DELTA_BINARY_PACKED integer columns (the encoding 314's)
  default codec defs apply to every integer column).
* Add --nationkey-type {i32,i64} (default i32) and --regionkey-type
  {i32,i64} (default i32). tpchgen-cli defaults both to i64; these
  columns are low cardinality (25 nations, 5 regions) so i32 halves
  their bytes and matches the INTEGER width used in reference TPC-H
  DDL.
* Plumb the three flags through
  presto/scripts/setup_benchmark_data_and_tables.sh so the top-level
  benchmark setup script accepts the same options and forwards them to
  generate_data_files.py.
* Extend the DataGenArgs fixture in
  benchmark_data_tools/tests/common_fixtures.py with matching defaults
  so the existing codec/decimal/row-group tests exercise the new
  values automatically.

### DIFF
--- a/benchmark_data_tools/generate_data_files.py
+++ b/benchmark_data_tools/generate_data_files.py
@@ -29,6 +29,9 @@ def generate_partition(
     approx_row_group_bytes,
     convert_decimals_to_floats,
     codec_defs,
+    parquet_version,
+    nationkey_type,
+    regionkey_type,
 ):
     if verbose:
         print(f"Generating '{table}' partition: {partition}")
@@ -49,6 +52,12 @@ def generate_partition(
         "parquet",
         "--parquet-row-group-bytes",
         str(approx_row_group_bytes),
+        "--parquet-version",
+        parquet_version,
+        "--nationkey-type",
+        nationkey_type,
+        "--regionkey-type",
+        regionkey_type,
     ]
 
     if convert_decimals_to_floats:
@@ -126,6 +135,9 @@ def generate_data_files_with_tpchgen(args, codec_defs):
                         args.approx_row_group_bytes,
                         args.convert_decimals_to_floats,
                         codec_defs,
+                        args.parquet_version,
+                        args.nationkey_type,
+                        args.regionkey_type,
                     )
                 )
             max_partitions = num_partitions if num_partitions > max_partitions else max_partitions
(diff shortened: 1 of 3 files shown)


benchmark_data_tools/generate_data_files.py
```diff
@@ -415,5 +427,30 @@ def build_default_codec_defs():
         default=None,
         help="Path to a JSON file specifying per-table/per-column encoding, compression, and dictionary settings.",
     )
+    parser.add_argument(
+        "--parquet-version",
```
paul-aiyedun, line 431: I think this should always be V2. The current code has a bug where the file format shows V1, even though the default uses V2 encodings.

benchmark_data_tools/generate_data_files.py
```diff
@@ -415,5 +427,30 @@ def build_default_codec_defs():
...
+        help="Arrow type for c_nationkey, n_nationkey, s_nationkey. Default: i32.",
+    )
+    parser.add_argument(
+        "--regionkey-type",
+        type=str,
+        required=False,
+        default="i32",
+        choices=["i32", "i64"],
+        help="Arrow type for n_regionkey, r_regionkey. Default: i32.",
+    )
```
paul-aiyedun, line 454: I believe the current DDLs (https://github.com/rapidsai/velox-testing/blob/9b74ba7d2c3e4515a48a70fe4f5e7151dd358c55/presto/testing/common/schemas/tpch/region.sql#L2 and https://github.com/rapidsai/velox-testing/blob/9b74ba7d2c3e4515a48a70fe4f5e7151dd358c55/presto/testing/common/schemas/tpch/nation.sql#L2) still set these columns to int64/bigint. Does changing the dataset alone result in the expected behavior?

paul-aiyedun (commented):
Changes overall make sense to me. However, I had a question about possible mismatches between the CREATE TABLE DDL and generated datasets.

benchmark_data_tools/generate_data_files.py
kjmph: Yeah, I'll have to look that up. Do you have a pointer to the bug? Does this still exist?

benchmark_data_tools/generate_data_files.py
kjmph: I'll have to refresh. I generated this data a while back and have kept using it. Since this PR popped back on my radar, I'll go sit down and look.

kjmph: I was commenting on this PR earlier today, and I realize this thread wasn't finished. As we can see, if we follow the `setup_benchmark_tables.sh` path, we see that we generate the DDL from the Parquet files, not using the DDL that is checked in:
https://github.com/rapidsai/velox-testing/blob/main/presto/scripts/setup_benchmark_tables.sh#L56

However, if the dataset was generated in some other manner, then if Parquet is `i32`, it is widened to `BIGINT`, tested here:
https://github.com/facebookincubator/velox/blob/main/velox/dwio/parquet/tests/reader/ParquetReaderWideningTest.cpp#L596-L602

Although, I can't find a corresponding GPU test, so maybe this would be a problem on GPU, but again, that would be from a dataset not generated with this methodology.

PR #324, opened 2026-04-24
Quality of life improvements when benchmarking

kjmph:
Brings the CPU variant up to parity with GPU for multi-worker benchmarking: new jinja-rendered compose template, per-worker NUMA pinning, and host-aware auto-tuning in generate_presto_config.sh that derives memory / driver / cache sizing from the detected topology (NPROC, NUMA nodes, SMT ratio) so the same scripts produce sensible configs on any host. Also adds an optional --wait flag to block on worker registration after start, a cleaner stop_presto.sh that only tears down the running variant (no more stale-file "Resource is still in use" warnings), small fixes around per-worker log naming, and hive.file-splittable for the CPU path.

Per-file rationale is preserved in the PR review comments.


karthikeyann:
@misiugodfrey to review and merge this.

PR #325, opened 2026-04-24
Fix uploading failes query runs in benchmark

karthikeyann:
fix error NoneType while reading benchmark_results.json with some failed queries

### DIFF
--- a/benchmark_reporting_tools/post_results.py
+++ b/benchmark_reporting_tools/post_results.py
@@ -401,21 +401,9 @@ def _build_submission_payload(
         is_failed = query_name in failed_queries
 
         if times is None:
-            if is_failed:
-                query_logs.append(
-                    {
-                        "query_name": query_name.lstrip("Q"),
-                        "execution_order": execution_order,
-                        "runtime_ms": None,
-                        "status": "error",
-                        "extra_info": {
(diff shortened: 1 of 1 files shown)


Avinash-Raj (approved):

TomAugspurger (approved):
Down at https://github.com/rapidsai/velox-testing/pull/325/changes#diff-955cdbcd697eefa97dc47ade2fba8f1936b0df876257975e933d1ae46b5e0980R439, there's a section that appears to be trying to handle this

```
    # Handle failed queries that may not have times
    for query_name, error_info in failed_queries.items():
```

is that dead code now?

karthikeyann:
what does `execution_order` mean?

TomAugspurger:
It's a counter giving the actual order queries are executed in. This is useful when running multiple iterations of multiple queries, since you can distinguish between

1. `query=1 iteration=1, query=1 iteration=2, query=2 iteration=1, query=2 iteration=2`, and
2. `query=1 iteration=1, query=2 iteration=1, query=1 iteration=2, query=2 iteration=2`

karthikeyann:
@TomAugspurger I updated the code and minimized the diff. kept old code.

PR #329, opened 2026-04-29
Add ssh server to dev container

devavret:
Allow sshing to the dev container from host to use nsight systems ui on the host to profile in container


.devcontainer/cuda13.1/devcontainer.json
```diff
@@ -38,6 +38,7 @@
     "--ulimit", "nofile=500000"
   ],
   "postCreateCommand": "/home/coder/velox-testing/scripts/devcontainer/post-create",
+  "postStartCommand": "/bin/bash /home/coder/velox-testing/scripts/devcontainer/start-sshd",
```
bdice: Instead of calling `start-sshd` directly, let's use a `post-start` script that calls `start-sshd`.

scripts/devcontainer/initialize
```diff
@@ -0,0 +1,21 @@
...
+
+mkdir -p \
+    "${workspace_parent}/.aws" \
+    "${workspace_parent}/.cache" \
+    "${workspace_parent}/.config/pip" \
+    "${workspace_parent}/.local/share/velox-testing-cuda${cuda_version}-venvs"
+
+authorized_keys="${workspace_parent}/.config/velox-testing-authorized_keys"
+if compgen -G "${host_home}/.ssh/*.pub" >/dev/null; then
+    cat "${host_home}"/.ssh/*.pub > "${authorized_keys}"
```
bdice, line 18: @trxcllnt Do you have advice on how to do this better? I know you've done some work with SSH agent forwarding. I don't know if that's the right tool to use here.

trxcllnt: Yeah you shouldn't need to do this. The host's SSH agent socket will be automatically forwarded to the devcontainer by the VSCode extension. If you aren't using VSCode, you should mount the host socket and set the `SSH_AUTH_SOCK` envvar in the docker run command like this:

```shell
docker run -v "$SSH_AUTH_SOCK:/tmp/ssh_auth_socket" -e SSH_AUTH_SOCK=/tmp/ssh_auth_socket ...
```

That will ensure any ssh operations inside the container use your host's agent, including using its authorized keys.

trxcllnt:
What about installing the [`sshd` feature](https://github.com/devcontainers/features/tree/main/src/sshd) instead?

scripts/devcontainer/initialize
devavret: @trxcllnt  but I'm trying to ssh **from** the host **to** the container. What you described sounds to me like ssh from inside the container to something external will be automatically handled. Am I misunderstanding this?

trxcllnt: You're right, I was misunderstanding the reason for this code. Can you take a look at using https://github.com/devcontainers/features/tree/main/src/sshd instead?

devavret: Codex actually suggested this before I changed it to install OpenSSH server. I thought I'd have to otherwise add it to each of the two jsons. I didn't know this is the canonical way of doing this.

.devcontainer/cuda13.1/devcontainer.json
devavret: start-sshd itself is a script. Do you want me to rename it to be a more generic `post-start`?

bdice (approved):
This seems fine to me. I'd like a final approval from @trxcllnt regarding the usage of sshd and SSH keys.

trxcllnt (approved):

devavret:
How do I merge this? Squash and merge is greyed out and CI isn't starting.

PR #330, opened 2026-04-30
Enable GDS, nsys, metrics collection for cluster usage

kingcrimsontianyu:
This PR adds optional capabilities to the Presto-Velox TPC-H benchmark runner on the NVL72 EPG cluster, all controlled by new flags on `launch-run.sh`. Default behavior is unchanged except that GDS is now on by default.

## GDS I/O (`--disable-gds` to opt out, on by default)

Workers run with `KVIKIO_COMPAT_MODE=OFF` so KvikIO uses GPU Direct Storage. With `--disable-gds`, workers fall back to POSIX I/O via KvikIO compat mode.

## Tunable worker env vars (`--worker-env-file`)

Env vars to be set in each worker container can now be declared in a sourced file rather than buried in the bash scripts. Defaults live in `worker.env` (currently `KVIKIO_TASK_SIZE=16MiB`, `KVIKIO_NTHREADS=16`); override the path with `--worker-env-file`.

## nsys profiling (`-p, --profile`)

Captures one `.nsys-rep` per query for a single worker (selectable via `--nsys-worker-id`). The worker image must include the `nsys` CLI. After pytest exits, the slurm job waits up to 10 minutes for nsys to finish flushing reports before tearing down the containers.

## Metrics collection (`-m, --metrics`)

After each query, pytest pulls per-query stats from the coordinator's REST API and writes them to `result_dir/metrics/<query>.json`.

## Nsys report and metrics uploading

Updates the `post_results.py` code so that nsys report and metrics can be uploaded to the online database. In particular, S3 is used to upload the large size nsys report.

## Other changes

- New `-q, --queries LIST` flag forwards a comma-separated query list through to pytest, useful for narrowing profile/metrics runs.
- README updated with full parameter documentation for `launch-run.sh`.
- `run_benchmark.sh` gains `--profile-script-path` so the slurm path can supply its own profiler functions instead of the docker default.

This PR supersedes #299


presto/slurm/presto-nvl72/run-presto-benchmarks.sh
```diff
@@ -68,6 +68,8 @@ cp -r ${LOGS}/cli.log ${SCRIPT_DIR}/result_dir/summary.txt
 echo "Collecting configs and logs into result directory..."
 collect_results
 
+wait_for_nsys_report_generation
```
kjmph, line 71: I believe `run_queries` will return non-zero if any of the tests fail. That means we will early exit, and won't get to this new `wait_for_nsys_report_generation` function. I mention because the PR comment said: `partial captures are still copied out.`. So, I think we intend for this to always execute? There are multiple ways to capture run_queries so this doesn't happen.. Happy to suggest a method!

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -237,17 +298,81 @@ ${worker_hive}:/opt/presto-server/etc/catalog/hive.properties,\
...
+            rm ${vt_nsys_report_dir}/.nsys_stop_token_w${worker_id}_\${qid}
+            ${nsys_bin} stop
+            echo \"Worker ${worker_id}: nsys stop exit code: \$?\"
+        done
+        echo \"Worker ${worker_id}: nsys subshell done, all queries profiled\"
+    ) &
+
+    echo \"Worker ${worker_id}: Nsight System program at ${nsys_bin}\"
+    echo \"Worker ${worker_id}: running nsys launch\"
+    ${nsys_bin} launch ${nsys_launch_opts} numactl --cpubind=${numa_node} --membind=${numa_node} /usr/bin/presto_server --etc-dir=/opt/presto-server/etc
```
kjmph: Other parts of this commit guard numctl with `USE_NUMA`, shouldn't this be guarded as well?

presto/slurm/presto-nvl72/run-presto-benchmarks.sh
kingcrimsontianyu: Thanks. I now think things should be kept simple here, so have removed the partial capture statement from the PR description.

presto/slurm/presto-nvl72/functions.sh
kingcrimsontianyu: Good catch! Added.

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -205,6 +209,32 @@ function run_worker {
     mkdir -p ${worker_data}/hive/data/user_data
     mkdir -p ${VT_ROOT}/.hive_metastore
 
+    local vt_worker_env_file="/var/worker_env_file"
+    local vt_cufile_log_dir="/var/log/cufile"
+    local vt_cufile_log="${vt_cufile_log_dir}/cufile_worker_${worker_id}.log"
+
+    local gds_mounts=""
+    if [[ "${ENABLE_GDS}" == "1" ]]; then
+        export MELLANOX_VISIBLE_DEVICES=all
```
kingcrimsontianyu, line 219: With pyxis hooks fixed on the cluster, now we only need to include this env var on the first compute node where `srun` for container creation is to be executed. There is no longer a need for IB-related bind mounts, and even no need for mounting `/run/udev`. But It seems that we still must mount `/dev/nvidia-fs*`.

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -364,8 +495,9 @@ function run_queries {
     export PRESTO_DATA_DIR=/var/lib/presto/data/hive/data/user_data; \
     export MINIFORGE_HOME=/workspace/miniforge3; \
     export HOME=/workspace; \
+    export NSYS_TOKEN_DIR=/workspace/presto/slurm/presto-nvl72/logs; \
```
kjmph: I was looking at this again.. It looks like this is dependent on the mounted directory structure. Are we guaranteeing that to remain in place for these tests? I'm used to the more specific mount style, where this is accessed via /var/log/nsys or something, and inside the chroot, we are agnostic to where it is sourced from. I'm not sure this needs to change, yet pointing it out as it does carry the expectation of the `velox-testing` directory structure.

kingcrimsontianyu: Thanks. `/var/log/nsys` has already been used for the worker container. Now using it for both the worker and CLI containers.

benchmark_reporting_tools/post_results.py
```diff
@@ -405,6 +408,9 @@ def _query_sort_key(name: str):
 
     for query_name in query_names:
         times = raw_times[query_name]
+        # Allow incomplete test results to be posted
```
TomAugspurger: https://github.com/rapidsai/velox-testing/pull/325 (just merged) has a similar change.

benchmark_reporting_tools/post_results.py
```diff
@@ -503,17 +509,64 @@ def _build_http_client(api_url: str, api_key: str, timeout: float) -> httpx.Asyn
     )
 
 
+async def _s3_presigned_put(
+    upload_url: str,
+    required_headers: dict[str, Any],
+    content: bytes,
+    timeout: float,
+) -> tuple[int, str]:
+    headers = {str(k): str(v) for k, v in required_headers.items()}
```
TomAugspurger, line 532: Out of curiosity, why are the `str` calls necessary? Is `presign.get("required_headers")` not already strings?

benchmark_reporting_tools/post_results.py
```diff
@@ -503,17 +509,64 @@ def _build_http_client(api_url: str, api_key: str, timeout: float) -> httpx.Asyn
...
+    async with httpx.AsyncClient(timeout=timeout) as s3_client:
```
TomAugspurger, line 533: Are you able to pass in the original `https.AsyncClient` here? Or do we need a new one for connecting to S3? I guess if we set a base URL or auth header on the original client it might not be usable for the S3 requests.

presto/slurm/presto-nvl72/worker.env
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
TomAugspurger, line 1: Just a note that in cudf-polars, we've been putting "important" environment variables in the `BenchmarkRun.extra_info.environment` field. `extra_info` is just a bag for arbitrary key-value pairs. We can elevate that to the top-level of the model if desired.

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -205,6 +210,32 @@ function run_worker {
...
     mkdir -p ${VT_ROOT}/.hive_metastore
 
+    local vt_worker_env_file="/var/worker_env_file"
+    local vt_cufile_log_dir="/var/log/cufile"
+    local vt_cufile_log="${vt_cufile_log_dir}/cufile_worker_${worker_id}.log"
+
+    local gds_mounts=""
+    if [[ "${ENABLE_GDS}" == "1" ]]; then
+        export MELLANOX_VISIBLE_DEVICES=all
+        gds_mounts="/etc/cufile.json:/etc/cufile.json:ro"
```
TomAugspurger, line 220: IIUC, the contents of this `/etc/cufile.json` are important for understanding if GDS is enabled / how it's configured.

Would it make sense to upload it as an asset?

benchmark_reporting_tools/post_results.py
kingcrimsontianyu: `str` for `k` can be removed since the type annotation indicates it is a string. I wasn't sure about `v` so added the `str`.

benchmark_reporting_tools/post_results.py
kingcrimsontianyu: Thanks for the info. Just merged into main and removed my version. The placeholder value in the main (`None`) makes sense to log the failed queries.

presto/slurm/presto-nvl72/functions.sh
kingcrimsontianyu: When GDS is enabled, cuFile specific log files `cufile_worker_<id>.log` will be generated and uploaded along with the worker logs. This cuFile log file contains the actual GDS status and detailed configs. So I think we probably don't need the `cufile.json` config file in addition.

TomAugspurger: Makes sense, thanks.

benchmark_reporting_tools/post_results.py
```diff
@@ -522,7 +572,13 @@ async def _upload_log_files(
     Returns:
         List of asset IDs from the uploaded files
     """
-    log_files = sorted(benchmark_dir.glob("*.log"))
+    log_files = sorted(effective_logs_dir.glob("*.log"))
+    log_files.extend(sorted(effective_logs_dir.glob("*.out")))
+    log_files.extend(sorted(effective_logs_dir.glob("*.err")))
```
kingcrimsontianyu, line 594: Fyi: Updated the scripts to generate SLURM output and error files in the `logs` directory and upload them to the database, so that we will know the nodelist used for the run. @misiugodfrey

presto/slurm/presto-nvl72/worker.env
kingcrimsontianyu: Good to know!

benchmark_reporting_tools/post_results.py
kingcrimsontianyu: That's right.
Just tried, and `PUT` failed with the error:
```text
...
Only one auth mechanism allowed; only the X-Amz-Algorithm query parameter, Signature query string parameter or the Authorization header should be specified
...
```
The original client has the field `Authorization: Bearer {api_key}` which conflicts with presigned URL authorization.

TomAugspurger (approved):

karthikeyann (approved):

PR #333, opened 2026-05-04
Fix CI image tag collisions and remove staging nightly variant

### LINKED ISSUE #332: Fix CI image tag collisions between duplicate build variants
# CI Image Variant Collision Proposal

## Problem

The Velox `upstream` and `staging` nightly variants can currently resolve to the same Velox commit. When that happens, they publish the same GHCR image tags and race each other in the `merge-manifests` cleanup path.

The reported Velox failure is consistent with this sequence:

1. `staging` pushes arch-specific tags such as `velox-<velox_sha>-gpu-cuda12.9-<date>-amd64`.
2. `upstream` creates the same multi-arch manifest from the same arch-specific tags.
3. `upstream` deletes the shared `-amd64` and `-arm64` intermediate tags.
4. `staging` later tries to create its multi-arch manifest from those tags and fails because they no longer exist.

The same collision exists for Presto images when variants resolve to the same Presto and Velox commit pair. `presto-coordinator` is especially exposed because its tag only contains the Presto short SHA and date, so any two Presto variants with the same Presto commit share coordinator image tags even if their Velox inputs differ.

## Current Tag Shape

The build workflows use commit/date tags for final multi-arch images, and append `-${arch}` only for intermediate per-arch images:

| Image | Multi-arch tag | Intermediate tags |
|---|---|---|
| Velox deps | `velox-deps-${VELOX_SHA}-cuda${CUDA_VERSION}-${DATE}` | `...-${ARCH}` |
| Velox build | `velox-${VELOX_SHA}-{cpu,gpu-cuda*}-${DATE}` | `...-${ARCH}` |
| Presto deps | `presto-deps-${PRESTO_SHA}-velox-${VELOX_SHA}-cuda${CUDA_VERSION}-${DATE}` | `...-${ARCH}` |
| Presto worker | `presto-${PRESTO_SHA}-velox-${VELOX_SHA}-{cpu,gpu-cuda*}-${DATE}` | `...-${ARCH}` |
| Presto coordinator | `presto-coordinator-${PRESTO_SHA}-${DATE}` | `...-${ARCH}` |

`build_variant` is threaded into `velox-build.yml` and `presto-build.yml`, but it is currently only used to decide which nightly variant updates `latest` tags. It is not part of the image tags used by the build, merge, test, or cleanup steps.

## Regression Timeline

The issue appears to be the result of multiple individually reasonable changes interacting:

1. [`26f352f`](https://github.com/rapidsai/velox-testing/commit/26f352f45591022d8c19e656814482b0622c8540) split CI entrypoints and introduced nightly fan-out wrappers. This made `upstream`, `pinned`, and `staging` run concurrently as separate reusable workflow calls.
2. [`d0d3c67`](https://github.com/rapidsai/velox-testing/commit/d0d3c670127fc9476fd6a1a12ce71189a760e467) threaded `build_type` and `build_variant` through the workflow call chain, but kept the durable image tag shape commit/date-based and used the variant only for `latest` tag ownership.
3. [`ea85623`](https://github.com/rapidsai/velox-testing/commit/ea85623b140e37debf544f6ffe321f46a6d07ed5) added deletion of arch-specific intermediate tags after the merge jobs. This created the deletion side of the race for variants that share tags.
4. [`24a2452`](https://github.com/rapidsai/velox-testing/commit/24a245274f9b64dfad15301822dc169d1ad89537) changed deletion to the GitHub Packages API and moved deps deletion to the final build merge job. This made cleanup effective and likely made the latent variant collision fail more reliably.

## Recommended Direction

Disable redundant `staging` jobs in nightly CI until `staging` is testing a distinct commit set again.

This is the smallest change that addresses the reported failures and removes duplicate CI cost. It also matches the current signal value: when `STABLE_VELOX_COMMIT` resolves to the same commit as `facebookincubator/velox:main`, Velox `staging` is not testing anything different from `upstream`. When `STABLE_PRESTO_COMMIT` and `STABLE_VELOX_COMMIT` resolve to the same inputs as another Presto variant, Presto `staging` is likewise redundant.

The immediate implementation is to delete the `staging` jobs in:

- `.github/workflows/velox-nightly.yml`
- `.github/workflows/presto-nightly.yml`

If keeping manual staging runs is useful, leave `velox.yml` and `presto.yml` unchanged. Operators can still run a staging build manually with `variant_label: staging` and explicit refs.

### 1. Disable Nightly Staging While It Is Redundant

Delete `staging` from scheduled nightly fan-out while keeping manual staging builds available through the reusable workflow entrypoints.

Implementation details:

- Delete the `staging` job in `.github/workflows/velox-nightly.yml`.
- Delete the `staging` job in `.github/workflows/presto-nightly.yml`.
- Leave `.github/workflows/velox.yml` and `.github/workflows/presto.yml` unchanged so `workflow_dispatch` can still run staging-like builds with explicit refs.
- Update `.github/workflows/README.md` to say scheduled nightly CI currently runs Velox `upstream` and Presto `upstream`/`pinned`, while staging remains a manual workflow mode.
- Re-add scheduled staging only when it is expected to test distinct resolved commits.

Pros:

- Smallest effective change.
- Eliminates the current collisions for Velox and Presto staging.
- Avoids spending CI capacity on duplicate builds/tests/benchmarks.
- Does not change published image tag contracts.

Cons:

- Does not fix the underlying tag namespace collision if two non-staging variants resolve to identical inputs.
- Requires a clear process for re-enabling staging when it becomes meaningful again.

Expected result: the current Velox and Presto staging collisions stop immediately, and CI no longer spends nightly capacity on duplicate builds/tests/benchmarks.

### 2. Add Run IDs To Intermediate Tags And Manual Final Tags

Use a run-scoped intermediate namespace for every build, and include the run ID in final multi-arch tags for manual builds. Scheduled nightly builds keep the existing stable final tag names. For example:

- Intermediate: `velox-${SHA}-gpu-cuda12.9-${DATE}-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}-${ARCH}`
- Nightly final: `velox-${SHA}-gpu-cuda12.9-${DATE}`
- Manual final: `velox-${SHA}-gpu-cuda12.9-${DATE}-${GITHUB_RUN_ID}`

The build jobs would push run-scoped arch tags. The merge jobs would read those run-scoped arch tags and publish final multi-arch tags based on `build_type`. Cleanup would delete the same run-scoped intermediate tags.

Implementation details:

- Add workflow-call inputs for `github_run_id` and `github_run_attempt` to `.github/workflows/velox-build.yml` and `.github/workflows/presto-build.yml`.
- Pass `${{ github.run_id }}` and `${{ github.run_attempt }}` from `.github/workflows/velox.yml` into `velox-build.yml`.
- Pass `${{ github.run_id }}` and `${{ github.run_attempt }}` from `.github/workflows/presto.yml` into `presto-build.yml`.
- In `velox-build.yml`, append the run suffix before `-${{ matrix.arch }}` for every arch-specific pushed tag and for every merge input reference.
- In `presto-build.yml`, append the same run suffix before `-${{ matrix.arch }}` for `presto-deps`, Presto worker, and `presto-coordinator` arch-specific tags and merge input references.
- In `velox-build.yml`, append `-${{ inputs.github_run_id }}` to final multi-arch tags when `inputs.build_type == 'manual'`.
- In `presto-build.yml`, append `-${{ inputs.github_run_id }}` to final multi-arch tags when `inputs.build_type == 'manual'`.
- Update `Determine deps image tag` steps so worker builds consume the run-scoped deps intermediates from the same workflow run.
- Update `Delete intermediate arch-specific tags` `base-tags` so cleanup targets only the current run's intermediates.
- Pass the resolved final tag suffix into `velox-test.yml`, `velox-benchmark.yml`, and `presto-test.yml` from `velox.yml` and `presto.yml`, so manual build-and-test runs test the run-scoped final tags they just published.
- Preserve existing final multi-arch tags for scheduled nightly builds and downstream users.

Pros:

- Directly fixes the race described in the logs.
- Preserves the existing durable nightly multi-arch image names used by downstream consumers.
- Protects against variant collisions and repeated manual runs for the same SHA/date.
- Avoids relying on semantic variant names for uniqueness.
- Prevents manual builds from repointing nightly final tags for the same SHA/date.

Cons:

- Manual image tags become longer and must include the GitHub run ID when used outside the build-and-test workflow.
- Manual image tags are no longer predictable from only commit SHA and date; users must know the workflow run ID.
- Scheduled nightly runs with identical resolved inputs still publish the same final tags; this is acceptable when the inputs are identical, but the final tag remains last-writer-wins.

Expected result: intermediate arch tags become private to a workflow run, so one merge job cannot delete another run's merge inputs. Nightly final tags remain stable and backward-compatible, while manual final tags cannot overwrite nightly outputs.

## Proposed Plan

1. Remove scheduled `staging` jobs from `velox-nightly.yml` and `presto-nightly.yml`.
2. Preserve manual staging builds through `velox.yml` and `presto.yml`.
3. Add run-scoped suffixes to all intermediate arch-specific image tags in `velox-build.yml` and `presto-build.yml`.
4. Add `github.run_id` to final multi-arch image tags for manual builds only.
5. Keep scheduled nightly final multi-arch image tags unchanged.
6. Validate by running a manual Velox workflow and a manual Presto workflow with identical refs in overlapping time windows, then confirm both merge jobs succeed, their run-scoped intermediates are cleaned up, and manual final tags do not overwrite nightly final tags.

The first step removes the current redundant scheduled collision source. The second step fixes the underlying intermediate-tag race for any future overlapping runs with identical resolved inputs and prevents manual builds from republishing nightly final tags.

bdice:
## Summary
- Add per-run intermediate image tags and an optional run ID suffix for final tags so manual builds are collision-safe.
- Prevent manual builds from updating stable `*-latest-*` tags while preserving scheduled nightly final tag names.
- Drop the `staging` nightly jobs and remove `build_variant: staging` from CI image workflows; the `STABLE_*` repo variables are no longer consumed by these workflows.
- Propagate run ID suffixes through Velox tests, Velox benchmarks, and Presto tests, and document the new tag behavior.

Closes #332


.github/actions/resolve-inputs/action.yml
```diff
@@ -33,6 +33,9 @@ outputs:
   date:
     description: 'Resolved build date'
     value: ${{ steps.resolve.outputs.date }}
+  image_tag_suffix:
```
bdice: Let's call it "run id suffix" or similar. Applies throughout.

.github/workflows/presto-build.yml
```diff
@@ -41,6 +41,19 @@ on:
         type: string
         required: false
         default: 'manual'
+      github_run_id:
+        description: 'GitHub Actions run ID used to isolate temporary tags'
```
bdice: "used to make temporary tags unique"

And similar throughout.

.github/workflows/presto-build.yml
```diff
@@ -143,7 +156,7 @@ jobs:
           outputs: type=registry,compression=zstd,force-compression=true,compression-level=15
           push: true
           tags: |
-            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:presto-deps-${{ inputs.presto_short_sha }}-velox-${{ inputs.velox_short_sha }}-cuda${{ matrix.cuda_version }}-${{ inputs.date }}-${{ matrix.arch }}
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:presto-deps-${{ inputs.presto_short_sha }}-velox-${{ inputs.velox_short_sha }}-cuda${{ matrix.cuda_version }}-${{ inputs.date }}-${{ inputs.github_run_id }}-${{ inputs.github_run_attempt }}-${{ matrix.arch }}
```
bdice: I don't think the run attempt is needed here. If we rerun a job it should produce the same tag output.

Applies throughout.

.github/workflows/presto.yml
```diff
@@ -55,6 +55,11 @@ on:
       run_tests: *run_tests
       build_type: *build_type
       build_variant: *build_variant
+      final_image_tag_suffix:
```
bdice: Same here, this should continue to reference the optional run ID (included for manual runs).

.github/workflows/README.md
```diff
@@ -110,6 +110,8 @@ Multi-arch (amd64/arm64) Docker images for Velox and Presto are built and publis
   | `pinned` | Presto's pinned Velox submodule | `prestodb/presto:master` |
   | `staging` | `$STABLE_VELOX_REPO:$STABLE_VELOX_COMMIT` | `$STABLE_PRESTO_REPO:$STABLE_PRESTO_COMMIT` |
 
+  `staging` remains available through manual `workflow_dispatch` runs, but is not scheduled nightly while it resolves to the same commits as other variants.
```
bdice: We can purge `staging` for now, we don't need to include it anywhere.

.github/actions/resolve-inputs/action.yml
```diff
@@ -33,6 +33,9 @@ outputs:
   date:
     description: 'Resolved build date'
     value: ${{ steps.resolve.outputs.date }}
+  run_id_suffix:
+    description: 'Run ID suffix after the resolved build date, if present'
```
bdice:
```suggestion
    description: 'Run ID suffix, if present'
```

.github/actions/resolve-inputs/action.yml
```diff
@@ -50,18 +53,28 @@ runs:
           TAG="${IMAGE_TAG##*:}"
           echo "Parsing tag: ${TAG}"
 
-          # Presto format: presto-{SHA}-velox-{SHA}-...-{DATE}[-arch]
+          # Presto format: presto-{SHA}-velox-{SHA}-...-{DATE}[-suffix]
```
bdice:
```suggestion
          # Presto format: presto-{SHA}-velox-{SHA}-...-{DATE}{RUN_ID_SUFFIX}
```

.github/actions/resolve-inputs/action.yml
```diff
@@ -50,18 +53,28 @@ runs:
...
             echo "date=${BASH_REMATCH[3]}" >> "$GITHUB_OUTPUT"
+            suffix="${BASH_REMATCH[4]}"
+            if [[ ! "${suffix}" =~ ^-[0-9]+$ ]]; then
+              suffix=""
+            fi
+            echo "run_id_suffix=${suffix}" >> "$GITHUB_OUTPUT"
             echo "Parsed presto image: presto=${BASH_REMATCH[1]}, velox=${BASH_REMATCH[2]}, date=${BASH_REMATCH[3]}"
 
-          # Velox format: velox-{SHA}-...-{DATE}[-arch]
+          # Velox format: velox-{SHA}-...-{DATE}[-suffix]
```
bdice:
```suggestion
          # Velox format: velox-{SHA}-...-{DATE}{RUN_ID_SUFFIX}
```

.github/workflows/presto-build.yml
```diff
@@ -37,10 +37,19 @@ on:
...
+        description: 'One of: [upstream, pinned, manual]'
         type: string
         required: false
         default: 'manual'
+      github_run_id:
+        description: 'GitHub Actions run ID used to make temporary tags unique'
+        type: string
+        required: true
+      final_run_id_suffix:
+        description: 'Optional run ID suffix appended to final multi-arch image tags'
```
bdice: Why is this "final"? Can we keep it consistent and call it `run_id_suffix` like the rest?

.github/workflows/presto-build.yml
```diff
@@ -364,23 +376,26 @@ jobs:
           BUILD_DATE: ${{ inputs.date }}
           BUILD_TYPE: ${{ inputs.build_type }}
           BUILD_VARIANT: ${{ inputs.build_variant }}
+          GITHUB_RUN_ID: ${{ inputs.github_run_id }}
+          FINAL_RUN_ID_SUFFIX: ${{ inputs.final_run_id_suffix }}
         run: |
           create_manifest() {
             local tag="$1"
+            local final_tag="${tag}${FINAL_RUN_ID_SUFFIX}"
```
bdice: Don't construct this here, construct the tag in the call to `create_manifest`.

.github/workflows/presto-build.yml
```diff
@@ -408,29 +423,32 @@ jobs:
           BUILD_DATE: ${{ inputs.date }}
           BUILD_TYPE: ${{ inputs.build_type }}
           BUILD_VARIANT: ${{ inputs.build_variant }}
+          GITHUB_RUN_ID: ${{ inputs.github_run_id }}
+          FINAL_RUN_ID_SUFFIX: ${{ inputs.final_run_id_suffix }}
```
bdice:
```suggestion
          RUN_ID_SUFFIX: ${{ inputs.run_id_suffix }}
```

.github/workflows/presto-build.yml
```diff
@@ -364,23 +376,27 @@ jobs:
...
           BUILD_TYPE: ${{ inputs.build_type }}
           BUILD_VARIANT: ${{ inputs.build_variant }}
+          RUN_ID_SUFFIX: ${{ inputs.run_id_suffix }}
+          INTERMEDIATE_TAG_SUFFIX: ${{ inputs.intermediate_tag_suffix }}
         run: |
           create_manifest() {
             local tag="$1"
+            local final_tag="$2"
             echo "Creating multi-arch manifest: ${tag}"
-            docker buildx imagetools create -t "${REGISTRY}/${IMAGE_NAME}:${tag}" \
```
misiugodfrey, line 382: The syntax for `create_manifest()` is a little confusing to me now.  It looks like we always call it with parameters `${tag}` and `${tag}${RUN_ID_SUFFIX}` (the `final_tag` value always contains the `tag` value).  If this is the case, can we split these into `${tag}` and `${RUN_ID_SUFFIX}` and compose the final_tag inside as necessary?  I think this would be more clear.

I'd also like to suggest that functions like `create_manifest()` get put into a script that can be referenced in multiple locations, because right now I think we are defining the same function in multiple places and it would be good to deduplicate.

I'm fine with marking that as a future PR, if you think it is out of scope for this one.

.github/workflows/presto-build.yml
```diff
@@ -364,23 +376,27 @@ jobs:
...
+            create_manifest "${tag}" "${tag}${RUN_ID_SUFFIX}"
           done
-          if [[ "${BUILD_TYPE}" == "nightly" && "${BUILD_VARIANT}" == "pinned" ]]; then
+          if [[ "${BUILD_TYPE}" == "nightly" && "${BUILD_VARIANT}" == "pinned" && -z "${RUN_ID_SUFFIX}" ]]; then
             for cuda in 12.9 13.1; do
               docker buildx imagetools create \
                 -t "${REGISTRY}/${IMAGE_NAME}:presto-deps-latest-cuda${cuda}" \
-                "${REGISTRY}/${IMAGE_NAME}:presto-deps-${PRESTO_SHORT_SHA}-velox-${VELOX_SHORT_SHA}-cuda${cuda}-${BUILD_DATE}-amd64" \
-                "${REGISTRY}/${IMAGE_NAME}:presto-deps-${PRESTO_SHORT_SHA}-velox-${VELOX_SHORT_SHA}-cuda${cuda}-${BUILD_DATE}-arm64"
+                "${REGISTRY}/${IMAGE_NAME}:presto-deps-${PRESTO_SHORT_SHA}-velox-${VELOX_SHORT_SHA}-cuda${cuda}-${BUILD_DATE}${INTERMEDIATE_TAG_SUFFIX}-amd64" \
```
misiugodfrey: This `docker buildx imagetools create` could be expressed as a `create_manifest` call, no?  Is there a reason we want to keep it separate?

.github/actions/resolve-inputs/action.yml
```diff
@@ -50,18 +53,28 @@ runs:
           TAG="${IMAGE_TAG##*:}"
           echo "Parsing tag: ${TAG}"
 
-          # Presto format: presto-{SHA}-velox-{SHA}-...-{DATE}[-arch]
+          # Presto format: presto-{SHA}-velox-{SHA}-...-{DATE}{RUN_ID_SUFFIX}
```
misiugodfrey: Minor note, but it seems to break convention that these RUN_ID_SUFFIX and INTERMEDIATE_TAG_SUFFIX contain hyphens, whereas all other variables used in this way to not.  Is there a particular case where we expect these values to be empty/optional so that we want them to be silently omitted?

misiugodfrey (commented):
There's some duplication that's starting to accumulate in these files that I would like to export into scripts, although I understand if we want to perform that in a separate PR.

I am not sure I understand the intent behind the optional-ness of the run_id and why we need both an intermediary run_id and a general one.  What situation are we trying to avoid by always having a single unique run_id?

bdice:
> I am not sure I understand the intent behind the optional-ness of the run_id and why we need both an intermediary run_id and a general one.  What situation are we trying to avoid by always having a single unique run_id?

There are two separate purposes here:

`run_id_suffix` is for final/public multi-arch image tags. It is intentionally optional because scheduled nightlies should keep the existing stable date-based tag names, while manual runs need a run-ID suffix so they do not collide with scheduled or other manual final tags. There are other possible solutions -- we could always include run IDs in the nightly image tags too -- but I didn't think that would be useful enough to break the current conventions.

`intermediate_tag_suffix` is for internal arch-specific tags used as manifest inputs. It is intentionally always present and more specific because those tags are temporary scratch state that can collide across overlapping runs before cleanup. It includes run ID, run attempt, and a sanitized variant label so one run cannot publish or delete another run’s intermediate `-amd64` / `-arm64` inputs.

So a single suffix would lose one of those properties: always appending it would change scheduled final tag names, while making it optional would leave intermediate tags collision-prone.

.github/workflows/presto-build.yml
bdice: Resolved by 7334df2.

`create_manifest` now lives in `scripts/create_manifest.sh`, is sourced from the workflow steps, and takes `tag` plus `RUN_ID_SUFFIX` instead of a pre-composed final tag.

.github/workflows/presto-build.yml
bdice: Resolved by aa96422.

The stable `latest` manifests now use the shared manifest helper via `create_manifest_from_arch_tags`, so the raw `docker buildx imagetools create` call only exists in `scripts/create_manifest.sh`.

.github/actions/resolve-inputs/action.yml
bdice: We have `-${RUN_ID}` for manual builds but it's empty for nightlies.

bdice:
Closed and reopened to kick the CI state. It's not wanting to run for some reason.

misiugodfrey (approved):

PR #334, opened 2026-05-05
Replace variant sed blocks with override files for CPU/GPU Presto configs

misiugodfrey:
## Summary

- Replaces ad-hoc `sed`/`echo` post-processing in `generate_presto_config.sh` with a structured override file system: variant-specific files under `docker/config/template/overrides/{cpu,gpu}/` are appended to the pbench-generated configs after rendering
- Moves CPU-specific settings (exchange tuning, split sizing, pushdown filters, `cudf.enabled=false`) and GPU-specific settings (`cudf.*`, `parquet.reader.*`, `hive.file-splittable=false`, optimizer flags) out of the shared templates and into their respective override directories
- Base templates now contain only the settings common to all variants; adding new per-variant config options no longer requires touching the generator script
- Added new options observed to speed up presto-cpu benchmarking on 8-node 1k/3k runs.

## Test plan

- [x] Run `generate_presto_config.sh` with `VARIANT_TYPE=gpu` — verify generated config files match those from previous benchmark runs.
- [x] Run with `VARIANT_TYPE=cpu` — verify generated config files match those for in updated POC benchmarks
- [x] Run with `VARIANT_TYPE=java` — verify no regression (java has no override directory, so only the base template is used)


presto/docker/config/template/etc_coordinator/catalog/hive.properties
```diff
@@ -14,12 +14,6 @@ hive.metastore.catalog.dir=file:/var/lib/presto/data/hive/metastore
 # state and clean up artifacts without manual intervention.
 hive.allow-drop-table=true
 
-# Control whether Presto can split files for parallel reads. Disable when the
-# file compression/format isn't splittable to avoid read failures. TPCH Parquet
-# test data commonly uses SNAPPY compression that isn't splittable at the file
-# level here, hence this must be false.
-hive.file-splittable=false
```
misiugodfrey, line 21: We want this option on for presto-cpu, but off for gpu - moved to overrides.

presto/docker/config/template/etc_coordinator/config_native.properties
```diff
@@ -25,10 +25,6 @@ node-scheduler.max-pending-splits-per-task=2000
 node-scheduler.max-splits-per-node=2000
 
 # Optimizer flags
-# New option in 0.282 to control inference of NOT NULL columns in joins (undocumented).
-#optimizer.joins-not-null-inference-strategy=USE_FUNCTION_METADATA
```
misiugodfrey, line 29: These options had previously been disabled, but appear to help in cpu-mode.  Moved to overrides.

presto/docker/config/template/etc_worker/catalog/hive.properties
```diff
@@ -13,16 +13,3 @@ hive.metastore.catalog.dir=file:/var/lib/presto/data/hive/metastore
...
-
-# Control whether Presto can split files for parallel reads. Disable when the
-# file compression/format isn't splittable to avoid read failures. TPCH Parquet
-# test data commonly uses SNAPPY compression that isn't splittable at the file
-# level here, hence this must be false.
-hive.file-splittable=false
-
-# Parquet read options
-# Limit (in bytes) on total number of bytes to be returned per read, or 0 if there is no limit
-parquet.reader.chunk-read-limit=0
```
misiugodfrey, line 25: The `parquet.reader*` options are cudf-exclusive.  Moved to overrides.

presto/docker/config/template/etc_worker/config_native.properties
```diff
@@ -32,28 +32,4 @@ system-mem-shrink-gb=20
 # overwritten to "false" in multi-node settings.
 single-node-execution-enabled=true
 
-# Enable cuDF (CPU mode will ignore this setting)
```
misiugodfrey, line 35: All the cudf.* options are ignored in by cpu-workers.  Moved to overrides.

presto/docker/config/template/overrides/cpu/etc_coordinator/catalog/hive.properties
```diff
@@ -0,0 +1,14 @@
+hive.file-splittable=true
+
+hive.max-outstanding-splits=1800
+hive.max-initial-splits=600
+hive.max-initial-split-size=128MB
+hive.max-split-size=256MB
+hive.split-loader-concurrency=32
+hive.pushdown-filter-enabled=true
+
+hive.parquet.pushdown-filter-enabled=true
```
misiugodfrey, line 10: The `hive.parquet*` options are only used by the coordinator.  Right now our `etc_common/catalog/hive` paths get overwritten instead of appended to, so the options above this are duplicated in both the coordinator and worker paths.  Not sure if we want to change that - but I think that's outside the scope of this PR.

presto/docker/config/template/overrides/cpu/etc_worker/config_native.properties
```diff
@@ -0,0 +1,18 @@
+# cuDF has no effect in CPU mode; disable to suppress startup warnings.
+cudf.enabled=false
+
+exchange.http-client.enable-connection-pool=true
```
misiugodfrey, line 4: These new options were all found to help in benchmarking (and are recommended by IBM).  They were not found to have much impact until we boosted the max-buffer-size on a busy cluster.

presto/scripts/generate_presto_config.sh
```diff
@@ -48,11 +48,9 @@ function duplicate_worker_configs() {
 
   # Some configs should only be applied if we are in a multi-worker environment.
   if [[ ${NUM_WORKERS} -gt 1 ]]; then
-    sed -i "s+single-node-execution-enabled.*+single-node-execution-enabled=false+g" ${coord_native_config}
-    sed -i "s+single-node-execution-enabled.*+single-node-execution-enabled=false+g" ${worker_native_config}
+    sed -i "s+single-node-execution-enabled.*+single-node-execution-enabled=false+g" ${coord_native_config} ${worker_native_config}
```
misiugodfrey, line 51: Use the same sed multi-file syntax as is used below.

presto/scripts/generate_presto_config.sh
```diff
@@ -123,37 +121,16 @@ EOF
     sed -i "s|hive.metastore.catalog.dir=.*|hive.metastore.uri=${HIVE_METASTORE_URI}|" "${CONFIG_DIR}/etc_coordinator/catalog/hive.properties" "${CONFIG_DIR}/etc_worker/catalog/hive.properties"
   fi
 
-  COORD_CONFIG="${CONFIG_DIR}/etc_coordinator/config_native.properties"
```
misiugodfrey, line 126: Removed these sed commands in favour of the file overrides.

simoneves (approved):
LGTM

misiugodfrey:
A quick note: Because the `overrides`  directory is in `templates` it gets parsed and duplicated by pbench - creating a duplicate folder in `generated`.  We could move the folder to prevent this... but I think it may be useful in the future in case we want to use any pbench variables in the override files.  I'm fine with going either way on this.

presto/scripts/generate_presto_config.sh
```diff
@@ -123,37 +121,16 @@ EOF
...
+  # Apply variant-specific override files by appending them to the generated configs
+  OVERRIDES_DIR="${SCRIPT_DIR}/../docker/config/template/overrides/${VARIANT_TYPE}"
+  if [[ -d "${OVERRIDES_DIR}" ]]; then
+    while IFS= read -r -d '' src_file; do
+      rel_path="${src_file#${OVERRIDES_DIR}/}"
+      dest_file="${CONFIG_DIR}/${rel_path}"
+      if [[ -f "${dest_file}" ]]; then
+        cat "${src_file}" >> "${dest_file}"
+      fi
+    done < <(find "${OVERRIDES_DIR}" -type f -print0)
```
simoneves: Ew. But OK.

misiugodfrey: I'm open to an alternate syntax if we want to change it.

simoneves: Maybe at least do the `find` in advance into a variable with an explanatory name so that it's not so complex a one-liner

presto/scripts/generate_presto_config.sh
mattgara: It looks like we are picking from `sed` commands or file overrides to determine what config gets used. Reading this, its a bit hard to follow where one should look to see a definitive config given a mode. 

Have we considered storing the configuration we want in a python dictionary then writing it to a config file? A nice property of this is that `json` files naturally map to python dictionaries and vice-versa, so it gives a concise way to store these representations on disk as well.

presto/scripts/generate_presto_config.sh
misiugodfrey: How about this syntax using `mapfile`?

paul-aiyedun (approved):

presto/scripts/generate_presto_config.sh
misiugodfrey: I don't think we want to go witha python dictionary for this.  We could refactor this to be more clear (Paul has suggested we could use jinja like we do for the docker config generation), but  I think anything we do will be a larger refactor outside the scope of this PR.

PR #338, opened 2026-05-07
Bake provenance JSON into images and auto-populate post_results.py args

### LINKED ISSUE #381: Add provenance information to images generated with scripts or CI
We want to easily be able to identify which branches/forks/commits were used to build any individual worker/coordinator image.  While this can be done with docker labels, those will be lost if the images are converted to a sqsh image, so the easiest way is just to have a provenance.json file inside the image that tracks this information.

This file can then be used by post_results.py to automatically upload accurate provenance information when posting benchmarks.

misiugodfrey:
## Summary

Capture the Presto/Velox source that built an image and thread it into benchmark reports so results are traceable to their build. Follow-up to #316 (image labels).

- Writes `/opt/velox-testing/provenance.json` into each Presto image at build time (both `native_build.dockerfile` and `provenance_labels.dockerfile`). Unlike the OCI labels from #316, a file is readable from inside a running container.
- Each launch script (`launch_presto_servers.sh`, `launch_coordinator.sh`) copies that file into the shared `LOGS_DIR` at container startup as `worker_provenance.json` / `coordinator_provenance.json`, so `run_context.py` running on the host (Docker) or inside the coordinator container (SLURM/Enroot) can read it via the existing logs bind-mount. The in-container `/opt/velox-testing/provenance.json` path is kept as a fallback for callers running directly inside an image that has the file baked in.
- SLURM `functions.sh`: adds `${LOGS}:/opt/presto-server/logs` to both coordinator and worker `srun --container-mounts` so the same scheme works across the cluster, mirrors the provenance `cp` in the inline worker bash and `COORD_SCRIPT`, and exports `LOGS_DIR=/opt/presto-server/logs` in `run_queries` so the in-container pytest looks at the mounted path.
- `run_context.py` reads the provenance file and merges `presto_sha/branch/repo` and `velox_sha/branch/repo` into the benchmark context dict written to `benchmark_result.json`. Empty values are dropped, and the resolver prefers `worker_provenance.json` (full presto+velox fields) over `coordinator_provenance.json` (presto only on CI builds).
- `post_results.py` submission:
  - Auto-populates `--velox-branch`, `--velox-repo`, `--presto-branch`, `--presto-repo` from the `benchmark_result.json` context; explicit CLI args still take precedence.
  - Emits `presto_sha`/`velox_sha` as their own `engine_config` entries alongside branch/repo. Provenance fields are added individually with empties omitted, so a coordinator-only image contributes just its presto entries.
  - Derives `query_engine.commit_hash` from the SHAs as `presto-<sha>-velox-<sha>` (or `presto-<sha>` when velox is absent) when `--commit-hash` is not given; explicit `--commit-hash` still wins, and it stays `unknown` when no SHAs exist.
  - Older result files without provenance are handled gracefully (fields absent, `commit_hash` = `unknown`, no error).
- `presto-build.yml`:
  - Replaces the `labels:` input on the deps and coordinator build steps with a second `provenance_labels.dockerfile` wrapper step — buildx `labels:` applies OCI metadata only and does not execute `RUN` steps, so the provenance file would not have been written via that path.
  - Fixes the deps provenance step to reference the tag the deps image is actually pushed under (it was missing the `-<build_variant>-<run_id>` segments, so buildx failed with `not found`).

## Notes / follow-ups

- On CI-built images the branch fields are empty (CI knows only the SHA), so a SLURM run submits populated repos + SHAs with `commit_hash = presto-<sha>-velox-<sha>`, and branch fields only appear for locally-built images that baked a branch.
- The new `engine_config` keys (`presto_sha`/`velox_sha`, and the existing branch/repo keys) have so far only been exercised via `--dry-run`; server-side acceptance should be confirmed with a real submission.

## Test plan

- [x] Build a new native worker image: `presto/scripts/start_native_cpu_presto.sh`
- [x] Confirm provenance file exists: `docker run --rm presto-native-worker-cpu:$USER cat /opt/velox-testing/provenance.json`
- [x] Run a short benchmark: `presto/scripts/run_benchmark.sh -b tpch -s bench_sf1`
- [x] Confirm `benchmark_result.json` context has `velox_branch`, `velox_repo`, `presto_branch`, `presto_repo` populated
- [x] Run `post_results.py --dry-run` without branch/repo args; confirm payload `engine_config` shows values auto-populated from the result file
- [x] Run with `--velox-branch override`; confirm CLI value wins over the label
- [x] Test with an older result file (no provenance in context); confirm the fields are absent from `engine_config` and `commit_hash` is `unknown`, without error
- [x] Trigger a CI build via the updated `presto-build.yml`; confirm all deps/coordinator/worker images build and push with the provenance step green (run 29847632984)
- [x] Verify baked `provenance.json` in the CI images: coordinator has presto-only fields; worker has both presto and velox (repo + sha), branches empty
- [x] Pull a CI image into a SLURM/Enroot context and run a quick benchmark; confirm the provenance fields populate via the shared logs path and a real (non-dry-run) `post_results.py` submission is accepted with `commit_hash = presto-<sha>-velox-<sha>` and the separate sha/repo entries

Closes: #381


.github/workflows/presto-build.yml
```diff
@@ -118,28 +118,39 @@ jobs:
...
       - name: Build and push Presto deps image
-        id: push
         uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83  # v6.18.0
         with:
           context: presto-deps-context
           file: presto/presto-native-execution/scripts/dockerfiles/centos-dependency.dockerfile
           build-args: |
             CUDA_VERSION=${{ matrix.cuda_version }}
             ARM_BUILD_TARGET=generic
-          labels: |
```
misiugodfrey, line 129: Docker labels were lost if we convert docker images to sqsh images, so instead we store that information in a provenance.json file which is preserved inside the image regardless of conversion.

benchmark_reporting_tools/post_results.py
```diff
@@ -387,8 +393,10 @@ def _build_submission_payload(
     validation_results: dict | None = None,
     velox_branch: str | None = None,
     velox_repo: str | None = None,
+    velox_sha: str | None = None,
```
misiugodfrey, line 401: Usually only branch or sha is used, but decided it was better to have separate values than try to overload the meaning of one value.

benchmark_reporting_tools/post_results.py
```diff
@@ -413,8 +421,15 @@ def _build_submission_payload(
     # Use placeholders for version info if not provided
     if version is None:
         version = "unknown"
+    # Derive commit_hash from the image provenance SHAs when not explicitly given:
+    # "presto-<sha>-velox-<sha>" (worker), "presto-<sha>" (coordinator, no velox).
     if commit_hash is None:
-        commit_hash = "unknown"
```
misiugodfrey, line 417: `commit_hash` doesn't apply well to presto-velox builds since they are combination, but this is a top-level value in the benchmarking DB, so it seems useful to fill it with a concatenation of both.

patdevinwilson (commented):
Nice approach — baking a readable `provenance.json` (instead of relying on OCI labels that disappear in `.sqsh` conversion) and surfacing it via `LOGS_DIR` is clean. A few nits below; otherwise LGTM once the real submission path is confirmed.

presto/docker/provenance_labels.dockerfile
```diff
@@ -13,3 +13,8 @@ LABEL velox-testing.presto.sha=${PRESTO_SHA} \
       velox-testing.velox.sha=${VELOX_SHA} \
       velox-testing.velox.branch=${VELOX_BRANCH} \
       velox-testing.velox.repository=${VELOX_REPOSITORY}
+RUN mkdir -p /opt/velox-testing && \
+    printf '{"presto_sha":"%s","presto_branch":"%s","presto_repo":"%s","velox_sha":"%s","velox_branch":"%s","velox_repo":"%s"}\n' \
```
patdevinwilson: Minor robustness nit: building the JSON with `printf`/`%s` will produce invalid JSON if a branch/repo ever contains `"`, `\`, or a newline. SHAs are safe, but branch names aren't guaranteed. Consider writing via Python/`jq` (`jq -n --arg ...`) so values are escaped, or document that only plain git refs are supported.

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -373,6 +383,10 @@ echo \"Worker ${worker_id}: CUFILE_LOGFILE_PATH=\${CUFILE_LOGFILE_PATH:-unset}\"
 echo \"Worker ${worker_id}: KVIKIO_TASK_SIZE=\${KVIKIO_TASK_SIZE:-unset}\"
 echo \"Worker ${worker_id}: KVIKIO_NTHREADS=\${KVIKIO_NTHREADS:-unset}\"
 
+if [ -f /opt/velox-testing/provenance.json ]; then
+    cp /opt/velox-testing/provenance.json /opt/presto-server/logs/worker_provenance.json
```
patdevinwilson, line 411: With multi-worker runs, every worker writes the same path `worker_provenance.json`. That's fine if images are identical (last writer wins with the same content), but worth a one-line comment so nobody later tries to make this per-worker without also updating `run_context.py`.

presto/testing/performance_benchmarks/run_context.py
```diff
@@ -153,6 +153,40 @@ def _get_gpu_name_from_worker_logs() -> str | None:
...
+    container and the pytest-running coord container mount ${LOGS} at the same path.
+
+    Prefer the worker file (full presto+velox fields) over the coordinator (presto
+    only). Fall back to the in-image path for callers running directly inside a
+    container that has the file baked in.
+    """
+    logs_dir = os.environ.get("LOGS_DIR")
+    if logs_dir:
+        ld = Path(logs_dir)
+        for name in ("worker_provenance.json", "coordinator_provenance.json"):
```
patdevinwilson, line 183: Preferring worker over coordinator provenance here matches the CI shape (worker has presto+velox; coordinator is presto-only). Good call. One edge case: if only the coordinator file is present (e.g. worker copy failed), you still get useful presto fields — nice fallback chain.

benchmark_reporting_tools/post_results.py
```diff
@@ -413,8 +421,15 @@ def _build_submission_payload(
...
+        commit_parts = []
+        if presto_sha:
+            commit_parts.append(f"presto-{presto_sha}")
+        if velox_sha:
+            commit_parts.append(f"velox-{velox_sha}")
+        commit_hash = "-".join(commit_parts) if commit_parts else "unknown"
```
patdevinwilson, line 440: Agree with your note that `commit_hash` is awkward for a dual-repo build. The `presto-<sha>-velox-<sha>` form is clear for humans/DB. One thing to confirm with a real (non-dry-run) submission: does the benchmarking DB impose a max length / charset on `commit_hash`? Full SHAs make this ~85 chars.

benchmark_reporting_tools/post_results.py
```diff
@@ -746,6 +766,15 @@ async def _process_benchmark_dir(
         )
         return 1
 
+    # Fall back to image provenance labels captured in the benchmark context.
+    velox_branch = velox_branch or benchmark_metadata.velox_branch
+    velox_repo = velox_repo or benchmark_metadata.velox_repo
+    presto_branch = presto_branch or benchmark_metadata.presto_branch
+    presto_repo = presto_repo or benchmark_metadata.presto_repo
+    # SHAs come only from the baked image provenance (no CLI override).
+    velox_sha = benchmark_metadata.velox_sha
```
patdevinwilson, line 783: Asymmetry is intentional but slightly surprising: branch/repo allow CLI override, SHAs do not. Worth a sentence in the `--help` text (or PR description is enough) so operators don't expect `--velox-sha`.

.github/workflows/presto-build.yml
```diff
@@ -118,28 +118,31 @@ jobs:
...
           push: true
           tags: |
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:presto-deps-${{ inputs.presto_short_sha }}-velox-${{ inputs.velox_short_sha }}-cuda${{ matrix.cuda_version }}-${{ inputs.date }}-${{ inputs.build_variant }}-${{ github.run_id }}-${{ matrix.arch }}
 
+      - name: Apply provenance labels to Presto deps image
+        id: push
+        uses: ./velox-testing/.github/actions/apply-provenance-labels
+        with:
+          base_image: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:presto-deps-${{ inputs.presto_short_sha }}-velox-${{ inputs.velox_short_sha }}-cuda${{ matrix.cuda_version }}-${{ inputs.date }}-${{ inputs.build_variant }}-${{ github.run_id }}-${{ matrix.arch }}
+          tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:presto-deps-${{ inputs.presto_short_sha }}-velox-${{ inputs.velox_short_sha }}-cuda${{ matrix.cuda_version }}-${{ inputs.date }}-${{ inputs.build_variant }}-${{ github.run_id }}-${{ matrix.arch }}
```
patdevinwilson, line 140: Good catch fixing the deps tag reconstruction. Overwriting the same tag with the provenance-wrapped image means the non-provenance digest is briefly published then replaced — attestation correctly uses `steps.push` after this step. If anything external could pull between the two pushes, consider pushing provenance under a temp tag then moving, but probably fine for this registry flow.

presto/testing/performance_benchmarks/run_context.py
kjmph: I think these fixed filenames can cause us to report provenance from an earlier run. `start_presto_helper.sh` archives only `*.log`, so `worker_provenance.json` survives between starts. If the next run uses Java, which doesn't write this file, or an older worker image without baked provenance, we will still prefer the stale worker file over the current coordinator file.

Could we clear both provenance files when setting up `LOGS_DIR`, or make the filenames run-specific using `SERVER_START_TIMESTAMP`? Clearing them at startup seems like the smaller fix.

benchmark_reporting_tools/post_results.py
```diff
@@ -746,6 +766,15 @@ async def _process_benchmark_dir(
         )
         return 1
 
+    # Fall back to image provenance labels captured in the benchmark context.
+    velox_branch = velox_branch or benchmark_metadata.velox_branch
+    velox_repo = velox_repo or benchmark_metadata.velox_repo
```
kjmph, line 779: I think we should normalize these repository values before persisting them. Local builds populate them from the raw output of `git remote get-url origin`, and HTTPS remotes can contain credentials such as `https://user:TOKEN@github.com/...`. This fallback would submit that value in engine_config. It would also already be present in the baked file, shared logs, and benchmark result.. yet, we definitely don't want them in the engine_config.

Could we strip URL userinfo in `capture_build_provenance` and store either `owner/repo` or a credential-free URL? Sanitizing at capture time seems preferable because it protects every downstream artifact, not just the API submission.

presto/docker/launch_coordinator.sh
```diff
@@ -7,6 +7,13 @@ set -e
 LOGS_DIR="/opt/presto-server/logs"
 mkdir -p "${LOGS_DIR}"
 : "${SERVER_START_TIMESTAMP:?SERVER_START_TIMESTAMP must be set before starting the container}"
+
+# Surface baked-in image provenance in the shared logs dir so the host-side
+# pytest (Docker) or in-container pytest (SLURM) can read it via LOGS_DIR.
+if [ -f /opt/velox-testing/provenance.json ]; then
+  cp /opt/velox-testing/provenance.json "${LOGS_DIR}/coordinator_provenance.json"
```
kjmph, line 14: Should we mirror this in `launch_java_worker.sh`? The Java worker image also gets provenance baked into it, but its launcher never copies that file into the shared logs directory, so Java benchmarks always fall back to the coordinator's provenance. This also was noted in my earlier comment in regards to `run_context.py`.

presto/docker/provenance_labels.dockerfile
kjmph: Note, the same printf pattern in `native_build.dockerfile` would need the corresponding change.

presto/testing/performance_benchmarks/run_context.py
misiugodfrey: I've changed it so that the startup scripts clear the provenance files from `LOGS_DIR` at startup.

.github/workflows/presto-build.yml
misiugodfrey: I think we can leave it for now.  I don't think we have anything pulling between the two pushes, and even for stuff that may in the future (nightly runs), it would be a rare occurrence without much consequence.

benchmark_reporting_tools/post_results.py
misiugodfrey: I've added a credential-stripping step to `capture_build_provenance` that should handle the https cases.

presto/docker/launch_coordinator.sh
misiugodfrey: I've added a mirror for `launch_java_worker.sh`.

benchmark_reporting_tools/post_results.py
misiugodfrey: I think I'll switch these to using short hashes, as it avoids this issue and is more readable as well.  I don't think there's any reason we need to use the full hashes.

misiugodfrey: Switched commit_hash to 12-char short SHAs (presto-<short>-velox-<short>, ~38 chars) to bound the length and sidestep any DB column limit. The full 40-char SHAs are still submitted in the separate presto_sha/velox_sha engine_config entries, so no precision is lost.

patdevinwilson (approved):
LGTM

kjmph (approved):

benchmark_reporting_tools/post_results.py
bdice: I feel like it’d be cleaner (and safer) to store something like “facebookincubator/velox” and not a raw git remote. Just assume we’re working with GitHub repos and pass the org/user and repo name. Maybe fetch that with gh.

bdice (approved):

PR #339, opened 2026-05-07
Add s2 bundled to devcontainer

devavret:
Velox added a new dependency called s2geometry in https://github.com/facebookincubator/velox/commit/fcb63d379fb99661a9dbd9d0e5cbff52c92d9f15 and devcontainer build-velox was broken without it


bdice (approved):
I verified this fix. Merging.

PR #345, opened 2026-05-15
Fix manual run IDs and speed up arch tag cleanup

bdice:
## Summary
- Pass the parent run ID into called test and benchmark workflows so manual build images resolve to the run that produced them.
- Speed up arch-specific tag cleanup by resolving multiple tags in one lookup path and stopping package-version pagination once all requested tags are found.
- Use merged multi-arch deps images as build contexts so deps arch-specific intermediates can be cleaned up in the deps merge jobs.

The cleanup step was taking more than **6 minutes** to run because it was fetching _all_ pages of the container registry, once for _every_ image being deleted. The updates in `ci/delete-arch-tags.sh` now fetch only enough pages needed to resolve the IDs for the desired tags, which now only takes a second or two to run.

Since that is so much faster, I also moved the `presto-deps` image cleanup and refactored the other `presto` image builds to use the merged manifests rather than the arch-specific intermediates.

## Testing
- Presto manual dispatch: https://github.com/rapidsai/velox-testing/actions/runs/25946083183
- Velox manual dispatch: https://github.com/rapidsai/velox-testing/actions/runs/25946083203

### DIFF
--- a/.github/actions/delete-arch-tags/action.yml
+++ b/.github/actions/delete-arch-tags/action.yml
@@ -23,10 +23,13 @@ runs:
         IMAGE_NAME: ${{ inputs.image-name }}
         GITHUB_TOKEN: ${{ github.token }}
       run: |
-        # Process each base tag line
+        tags=()
         while IFS= read -r base_tag; do
           [[ -z "${base_tag}" ]] && continue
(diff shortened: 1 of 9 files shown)


Avinash-Raj (approved):
LGTM.

PR #349, opened 2026-05-18
CI - Dispatch presto tests irrespective of the target variant

Avinash-Raj:
Currently we can test only manual builds through presto-test CI manual dispatch. This PR extends it to cover any type of build variants (manual, pinned, upstream).


Avinash-Raj:
It seems like the relevant dispatch event properly pulls up the image which is relevant to the input build variant -> https://github.com/rapidsai/velox-testing/actions/runs/26048080250/job/76577453313

bdice (approved):

PR #353, opened 2026-05-22
CI - Retry docker pull on transient ghcr.io failures

Avinash-Raj:
## Summary

- Wrap `docker pull` in `presto-test.yml`'s "Pull and retag images" step in a small exponential-backoff retry helper, so a one-off `ghcr.io 502` no longer fails the whole integration matrix.

- Add the helper as `docker_pull_with_retry` in `scripts/common.sh` so other workflow steps can reuse it via `source scripts/common.sh`.


Avinash-Raj:
We also use docker pull in `velox-test.yml`, so I think I have to modify that as well.

bdice (approved):

Avinash-Raj:
/ok to test https://github.com/rapidsai/velox-testing/pull/353/commits/02114fa810eae224d077d9d52e1fd777e4561d72

PR #358, opened 2026-06-08
Add --label and --notes options to post_results.py

misiugodfrey:
## Summary

- Adds `--label` (repeatable) and `--notes` CLI options to `post_results.py`, wiring them through to the `labels` and `notes` fields in the benchmark submission API payload
- Fields are omitted from the payload when not provided, matching the API's optional semantics

## Test plan

- [x] `--dry-run` with `--label foo --label bar --notes "some note"` to confirm fields appear in printed payload
- [x] Live post to verify labels and notes are recorded on the benchmark run

### DIFF
--- a/benchmark_reporting_tools/post_results.py
+++ b/benchmark_reporting_tools/post_results.py
@@ -29,6 +29,16 @@ Usage:
         --identifier-hash abc123 \
         --cache-state warm
 
+    # With labels and notes:
+    python benchmark_reporting_tools/post_results.py /path/to/benchmark_output \
+        --sku-name PDX-H100 \
+        --storage-configuration-name pdx-lustre-sf-100 \
+        --benchmark-name tpch \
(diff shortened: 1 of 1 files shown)


benchmark_reporting_tools/post_results.py
```diff
@@ -366,6 +389,8 @@ def _build_submission_payload(
     velox_repo: str | None = None,
     presto_branch: str | None = None,
     presto_repo: str | None = None,
+    labels: list[str] | None = None,
+    notes: str | None = None,
```
Avinash-Raj, line 393: nit, Instead of passing multiple function arguments, we could use a dataclass to encapsulate them, making it easier to serialize and deserialize the arguments when needed.

Avinash-Raj:
LGTM:-)

Avinash-Raj (approved):

PR #359, opened 2026-06-09
Retry presto-deps dependency downloads on transient network failures

Avinash-Raj:
The `presto-deps` image build runs many unguarded `curl/wget/dnf` downloads from upstream `Presto/Velox `setup scripts, which intermittently fail on transient network blips. Inject retry/timeout config into the staged dependency Dockerfile so those downloads become resilient without forking the upstream scripts:

- `CURL_OPTIONS` (honoured by velox's `wget_and_untar`) for curl tarball fetches
- `dnf.conf` retries/timeout/minrate for dnf installs (incl. the CUDA repo)
- `/etc/wgetrc` tries/wait,retry for direct wget downloads

A grep guard fails the step loudly if the upstream Dockerfile anchor changes.

This [failure](https://github.com/rapidsai/velox-testing/actions/runs/27120166056/job/80035237774) should be addressed by this PR, which allows the curl command to retry on network failures.

### DIFF
--- a/.github/workflows/presto-build.yml
+++ b/.github/workflows/presto-build.yml
@@ -117,12 +117,28 @@ jobs:
           cp -r velox/scripts presto-deps-context/velox/scripts
           cp -r velox/CMake presto-deps-context/velox/CMake
 
+          # Harden the dependency-install layer against transient network blips.
+          # The upstream Presto/Velox setup scripts run many unguarded downloads
+          # (curl tarballs, dnf repos, wget mirrors); rather than fork those
+          # scripts we inject retry/timeout config into the staged Dockerfile:
+          #   - CURL_OPTIONS is honoured by velox's wget_and_untar (curl -L ...)
+          #   - dnf.conf retries/timeout/minrate cover dnf installs (incl. CUDA repo)
+          #   - /etc/wgetrc covers the remaining direct `wget` downloads (e.g. minio)
+          dockerfile=presto-deps-context/scripts/dockerfiles/centos-dependency.dockerfile
+          sed -i \
+            -e '/^ENV UCX_VERSION/a ENV CURL_OPTIONS="--retry 5 --retry-delay 5 --retry-all-errors --retry-connrefused --connect-timeout 30"' \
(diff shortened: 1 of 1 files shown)


.github/workflows/presto-build.yml
```diff
@@ -117,12 +117,28 @@ jobs:
...
+          # Harden the dependency-install layer against transient network blips.
+          # The upstream Presto/Velox setup scripts run many unguarded downloads
+          # (curl tarballs, dnf repos, wget mirrors); rather than fork those
+          # scripts we inject retry/timeout config into the staged Dockerfile:
+          #   - CURL_OPTIONS is honoured by velox's wget_and_untar (curl -L ...)
+          #   - dnf.conf retries/timeout/minrate cover dnf installs (incl. CUDA repo)
+          #   - /etc/wgetrc covers the remaining direct `wget` downloads (e.g. minio)
+          dockerfile=presto-deps-context/scripts/dockerfiles/centos-dependency.dockerfile
+          sed -i \
+            -e '/^ENV UCX_VERSION/a ENV CURL_OPTIONS="--retry 5 --retry-delay 5 --retry-all-errors --retry-connrefused --connect-timeout 30"' \
```
Avinash-Raj, line 129: Since this append happens after `ENV UCX_VERSION` line and we believe that the line should always be exists on the presto's centos deps dockerfile.

.github/workflows/presto-build.yml
```diff
@@ -117,12 +117,28 @@ jobs:
...
+            -e '/^ENV UCX_VERSION/a RUN { echo retries=10; echo timeout=60; echo minrate=500; } >> /etc/dnf/dnf.conf && { echo tries=5; echo timeout=30; echo waitretry=5; echo retry_connrefused=on; echo read_timeout=300; } >> /etc/wgetrc' \
+            "$dockerfile"
+          grep -q '^ENV CURL_OPTIONS=' "$dockerfile" || { echo "::error::failed to inject network-retry config (anchor 'ENV UCX_VERSION' not found in upstream dockerfile)"; exit 1; }
```
Avinash-Raj, line 132: Else it won't build the deps image.

PR #360, opened 2026-06-12
Improve compute sanitizer workflow

bdice:
## Summary

- Change `velox-compute-sanitizer-trigger.yaml` into a lightweight scheduler/manual dispatcher that starts separate `velox-compute-sanitizer-run.yaml` workflows for `memcheck`, `racecheck`, and `synccheck`.
- Move test discovery into `velox-compute-sanitizer-run.yaml`, matching the cuDF pattern where the run workflow owns discovery and execution.
- Keep the compute-sanitizer default image on `velox-latest-gpu-cuda13.1` and copy the clearer GHCR image guidance into the manual dispatch inputs.
- Update the workflow README to describe the new split.

## Context

The previous `trigger` workflow was not just a trigger: it discovered tests and ran the full weekly sanitizer matrix through reusable workflow calls. The cuDF workflow uses a cleaner split where the trigger workflow schedules separate run workflows, and the run workflow handles discovery plus execution. This PR reshapes Velox in that direction while preserving its current weekly tool coverage.

## Validation

- `yq e . .github/workflows/velox-compute-sanitizer-trigger.yaml`
- `yq e . .github/workflows/velox-compute-sanitizer-run.yaml`
- `git diff --check`
- Commit hooks: whitespace, EOF, codespell, zizmor

`actionlint` was not installed in this local environment.

### DIFF
--- a/.github/workflows/velox-compute-sanitizer-run.yaml
+++ b/.github/workflows/velox-compute-sanitizer-run.yaml
@@ -33,7 +33,7 @@ on:
       velox_image:
         required: true
         type: string
-        description: "Full Velox GPU container image reference"
+        description: 'Velox GPU image to test (e.g. ghcr.io/rapidsai/velox-testing-images:velox-e0169a0-gpu-cuda13.1-20260318). Leave empty to use velox-latest-gpu-cuda13.1.'
 
 env:
   REGISTRY: ghcr.io
--- a/.github/workflows/velox-compute-sanitizer-run.yaml
+++ b/.github/workflows/velox-compute-sanitizer-run.yaml
@@ -1,20 +1,7 @@
-name: compute-sanitizer-run
+name: Compute Sanitizer
+run-name: "Compute Sanitizer ${{ inputs.tool_name }}"
 
 on:
-  workflow_call:
-    inputs:
-      tool_name:
-        required: true
-        type: string
-        description: "Compute sanitizer tool to run (memcheck, racecheck, initcheck, synccheck)"
-      test_names:
-        required: true
-        type: string
-        description: "JSON array of test names to run"
-      velox_image:
-        required: true
-        type: string
-        description: "Full Velox GPU container image reference"
   workflow_dispatch:
     inputs:
       tool_name:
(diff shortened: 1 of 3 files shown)


bdice:
Tested here and it works as expected.
- https://github.com/rapidsai/velox-testing/actions/runs/28039903879
- https://github.com/rapidsai/velox-testing/actions/runs/28039916922
- https://github.com/rapidsai/velox-testing/actions/runs/28039915424
- https://github.com/rapidsai/velox-testing/actions/runs/28039914786

.github/workflows/velox-compute-sanitizer-run.yaml
```diff
@@ -27,20 +14,74 @@ on:
...
+            -e GITHUB_OUTPUT=/tmp/gh_output \
+            "${IMAGE}" bash -c "bash /tmp/discover.sh && cat /tmp/gh_output" >> "${GITHUB_OUTPUT}"
+
   run-sanitizer-tests:
-    name: ${{ inputs.tool_name }} / ${{ matrix.test_name }}
+    name: Run ${{ inputs.tool_name }} on ${{ matrix.test_name }}
+    needs:
+      - resolve-image
+      - discover-sanitizer-tests
+    if: ${{ !cancelled() }}
```
misiugodfrey, line 84: Does `!cancelled()` override the default behavior of "skip if dependent job failed"?

My understanding of the intent here is that we want to run this even if `discover-sanitizer-tests` was skipped (i.e. if tests to run were manually specified), but I don't think we want to run if it, or  the resolve-image steps failed.

Perhaps something like: ```if: ${{ !cancelled() && (inputs.test_names != '' }}``` conveys the intend more explicitly?

.github/workflows/velox-compute-sanitizer-trigger.yaml
```diff
@@ -1,97 +1,48 @@
...
-    needs: discover-tests
-    permissions:
-      contents: read
-      packages: read
-    uses: ./.github/workflows/velox-compute-sanitizer-run.yaml
-    with:
-      tool_name: "synccheck"
-      test_names: ${{ needs.discover-tests.outputs.tests }}
-      velox_image: ${{ needs.discover-tests.outputs.velox_image }}
+          gh workflow run velox-compute-sanitizer-run.yaml "${args[@]}"
```
misiugodfrey, line 48: IIUC this job will now be green as long as the sub-jobs have been dispatched - regardless of whether they pass or not.  Is this ok with how we intend to view/report these results?
I don't think we want to leave the triggering job spinning as it waits for the other jobs to finish; I just want to make sure we aren't going to miss failures if they do occur by looking at the wrong reporting.  If those jobs are hooked into slack or something then that's probably fine.

.github/workflows/velox-compute-sanitizer-trigger.yaml
```diff
@@ -1,97 +1,48 @@
...
+      fail-fast: false
+      matrix:
+        tool_name:
+          - memcheck
+          - racecheck
+          - synccheck
     steps:
-      - name: Checkout code
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
```
misiugodfrey, line 35: I don't think `trigger-sanitizer` needs the  checkout action any more?  Since it's just calling `gh workflow run` now, I think we can remove this action (and possibly the `contents: read` requirement).

misiugodfrey (commented):
Looks good.  Just a few minor comments for your consideration.

.github/workflows/velox-compute-sanitizer-trigger.yaml
kingcrimsontianyu, line 35: I'm not familiar with workflow management with yaml in general. Claude flagged a minor issue. Not sure if it makes sense or not. Just fyi:
>velox-compute-sanitizer-trigger.yaml:35 — the actions/checkout step is unnecessary. The job's only real step is gh workflow run …, which is a pure REST call (uses GH_TOKEN + the auto-set GITHUB_REPOSITORY); it never reads the working tree. The checkout — and the contents: read permission it forces at line 26 — is dead work that can be removed, leaving the job with just actions: write.

.github/workflows/velox-compute-sanitizer-trigger.yaml
```diff
@@ -1,97 +1,48 @@
...
+    - cron: "0 10 * * 6"  # Weekly on Saturday at 10:00 UTC
   workflow_dispatch:
     inputs:
       velox_image:
-        description: 'Velox GPU image to test (e.g. ghcr.io/rapidsai/velox-testing-images:velox-e0169a0-gpu-cuda13.1-20260318). Leave empty to use velox-latest-gpu-cuda12.9.'
+        description: "Velox GPU image to test (e.g. ghcr.io/rapidsai/velox-testing-images:velox-e0169a0-gpu-cuda13.1-20260318). Leave empty to use velox-latest-gpu-cuda13.1."
         type: string
         required: false
-        default: ''
+        default: ""
```
kingcrimsontianyu, line 16: Is there any difference between using single/double quotes in this yaml file? Or is this simply to maintain format consistency and use double quotes wherever possible?

.github/workflows/velox-compute-sanitizer-trigger.yaml
bdice: Yeah, this succeeds if the trigger succeeds. I think that’s fine. We can’t spin-wait here because the actual runs take too long.

.github/workflows/velox-compute-sanitizer-trigger.yaml
bdice: Probably so. I will check that.

.github/workflows/velox-compute-sanitizer-run.yaml
bdice: I will check into this.

PR #363, opened 2026-06-25
Use decimal TPCH data by default

bdice:
## Summary

- Remove the default `--convert-decimals-to-floats` flag from TPC-H data generation in CI benchmark/test workflows.
- Update SLURM and script examples so generated TPCH data keeps decimal columns by default.
- Leave the explicit conversion flag available for legacy callers that still need float columns.

## Why

Velox upstream now supports decimal operations, so the default TPCH data path no longer needs to force decimals into floats.


bdice:
Update on the failing benchmark run: the earlier YAML issue is fixed, and the workflow now gets past TPCH data generation. The remaining failure is in the Velox benchmark binary itself.

The failing job reaches `Run TPC-H GPU benchmarks`, starts query 1, and aborts while constructing the Q1 plan:

```text
Reason: Scalar function signature is not supported: minus(DOUBLE, DECIMAL(15, 2)).
Function: resolveScalarFunctionType
File: /workspace/velox/velox/parse/TypeResolver.cpp
...
TpchQueryBuilder::getQ1Plan()
```

So this PR is now exposing the intended decimal data path: `velox-testing` generates TPCH parquet with real decimal columns, and Velox's TPCH benchmark query builder still has an expression that mixes a `DOUBLE` operand with a `DECIMAL(15, 2)` operand. That fix should be in Velox, likely in the benchmark Q1 plan/query construction or expression typing/casting behavior, rather than in `velox-testing`.

bdice:
Recent `presto` workflow run for this branch: https://github.com/rapidsai/velox-testing/actions/runs/28459435154

paul-aiyedun (commented):
The infrastructure changes look good to me. @simoneves Are the TPC-H failures expected (https://github.com/rapidsai/velox-testing/actions/runs/28459435154/job/84353854672)?

simoneves:
> The infrastructure changes look good to me. @simoneves Are the TPC-H failures expected (https://github.com/rapidsai/velox-testing/actions/runs/28459435154/job/84353854672)?

They certainly are not. I'll try to test this locally ASAP.

simoneves (commented):
Investigating @paul-aiyedun 's comment about TPC-H failures.

bdice:
Triggered a new `presto` workflow run after keeping TPC-DS generated decimals converted to floats while preserving TPCH decimal coverage: https://github.com/rapidsai/velox-testing/actions/runs/28481484156

bdice:
Updated `rapidsai/velox-testing:codex/use-tpch-decimals` to the PR head `2f57818` and triggered a new `presto` workflow run: https://github.com/rapidsai/velox-testing/actions/runs/28482057355

bdice:
@paul-aiyedun @simoneves My agent said the previous run might've been on an older commit that didn't have the right fixes in place yet. I pushed again and triggered a new run. Also I filed #365 to track adding decimal support to the TPC-DS queries.

bdice:
Committed decimal-output-scale-aware result tolerance (`24f3e4b`) and triggered a new `presto` workflow run: https://github.com/rapidsai/velox-testing/actions/runs/28492309047

bdice:
The previous run https://github.com/rapidsai/velox-testing/actions/runs/28482057355/job/84422107755 failed because the reference values from DuckDB TPC-H are in `DOUBLE` while Presto TPC-H is in `DECIMAL` (as intended). I pushed 24f3e4b, which sets the comparison tolerance to `0.5 * 10^-scale`, so that it counts `DOUBLE` results that would round to the same `DECIMAL` value as equivalent. It seems to make tests pass locally, we'll see how the current test run in https://github.com/rapidsai/velox-testing/actions/runs/28492309047 goes.

@simoneves @paul-aiyedun This should be ready for review now.

simoneves (approved):
All stlll good with this branch of `velox-testing`, at least with TPC-H (DECIMAL SF50).

I do not have local TPC-DS data at this time, so I haven't tested that. I presume I can create some using the scripts. I'll look into that.

simoneves:
@bdice may I have some guidance on specifically what else I should be testing, please? :)

bdice:
@simoneves This PR implements decimal for (TPC-H, Presto). The other cases are:

- TPC-DS, Presto: Failing with decimal (passes with `double`). Needs to be investigated.
- TPC-H, Velox: Not supported upstream, Velox's TPC-H implementation is hardcoded to use `DOUBLE`. We _could_ tweak the upstream tests to use decimal, but it's probably low priority given that we now have Presto TPC-H coverage.

PR #364, opened 2026-06-30
Derive Presto integration schemas from Parquet data

bdice:
Auto-created Presto integration test tables now derive their Hive schema from the generated Parquet data instead of static SQL files. This keeps the table schema aligned with the generated data when decimal columns are preserved or converted to doubles.

Validated locally with the TPCH decimal repro path against the CI image stack.

### DIFF
--- a/presto/testing/integration_tests/create_hive_tables.py
+++ b/presto/testing/integration_tests/create_hive_tables.py
@@ -3,8 +3,15 @@
 
 import argparse
 import os
+import sys
+import tempfile
+from pathlib import Path
 
 import prestodb
+
+sys.path.append(str(Path(__file__).resolve().parents[3] / "benchmark_data_tools"))
+
+from generate_table_schemas import generate_table_schemas
 
 
 def create_tables(presto_cursor, schema_name, schemas_dir_path, data_sub_directory):
@@ -18,6 +25,12 @@ def create_tables(presto_cursor, schema_name, schemas_dir_path, data_sub_directo
                 file_path=f"/var/lib/presto/data/hive/data/{data_sub_directory}/{table_name}", schema=schema_name
             )
         )
+
+
+def create_tables_from_data(presto_cursor, schema_name, benchmark_type, data_dir_path, data_sub_directory):
+    with tempfile.TemporaryDirectory() as schemas_dir:
+        generate_table_schemas(benchmark_type, schemas_dir, data_dir_path, False)
+        create_tables(presto_cursor, schema_name, schemas_dir, data_sub_directory)
 
 
 def get_table_schemas(schemas_dir):
--- a/presto/testing/integration_tests/common_fixtures.py
+++ b/presto/testing/integration_tests/common_fixtures.py
@@ -28,9 +28,11 @@ def setup_and_teardown(request, presto_cursor):
 
     should_create_tables = not has_schema_name
     if should_create_tables:
-        schemas_dir = get_abs_file_path(__file__, f"../common/schemas/{benchmark_type}")
         data_sub_directory = f"integration_test/{benchmark_type}"
-        create_hive_tables.create_tables(presto_cursor, schema_name, schemas_dir, data_sub_directory)
+        data_dir = get_abs_file_path(__file__, f"../../../common/testing/integration_tests/data/{benchmark_type}")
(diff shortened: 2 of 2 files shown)


bdice:
Triggered a `presto` workflow run for `parquet-derived-integration-schemas` using the decimal repro Presto/Velox inputs: https://github.com/rapidsai/velox-testing/actions/runs/28460989763

presto/testing/integration_tests/create_hive_tables.py
```diff
@@ -3,9 +3,16 @@
 
 import argparse
 import os
+import sys
+import tempfile
+from pathlib import Path
 
 import prestodb
 
+sys.path.append(str(Path(__file__).resolve().parents[3] / "benchmark_data_tools"))
```
simoneves, line 12: Ew... but OK

simoneves (approved):

presto/testing/integration_tests/create_hive_tables.py
```diff
@@ -20,6 +27,12 @@ def create_tables(presto_cursor, schema_name, schemas_dir_path, data_sub_directo
         )
 
 
+def create_tables_from_data(presto_cursor, schema_name, benchmark_type, data_dir_path, data_sub_directory):
+    with tempfile.TemporaryDirectory() as schemas_dir:
+        generate_table_schemas(benchmark_type, schemas_dir, data_dir_path, False)
```
misiugodfrey, line 32: I think we may want to re-think how we represent the table schema we are using.

The benchmarks currently do something similar, where they use `generate_table_schemas` to create schema files in a temp directory, read them, and then delete them (using slightly different logic).  These are effectively workarounds to use the existing `generate_table_schemas` API for generating files, which doesn't seem to be in use anymore.

I think we should revisit whether we want `generate_table_schemas` to be writing to files at all, because if we use auto-detection here, then I don't think these generated files are used anymore.  I believe the only reason `generate_table_schemas` created files was so that they could be checked in for a consistent base for these integration tests,  and if the integration tests are going to detect the schema from the data files then they are dead artifacts (after this change, I don't think anything uses `common/schemas` anymore.

If we want to commit fully to detecting types from the data files (which is fine with me), then I think we should refactor `generate_table_schemas` to just return a schema type directly rather than write to a file and remove the static schemas from the repo.  It's probably a good idea to have @paul-aiyedun chime in too, since I believe he wrote the initial schema system and may know the motivations for that choice better.

I'm also ok with landing this as is and removing the legacy stuff as a followup to unblock testing, but I do want us to be aware that this change is effectively deprecating a section of the repo that will live on otherwise.

paul-aiyedun (changes requested):
We can regenerate the schemas and sample test files used for integration tests using the `generate_test_files.sh` script. My understanding is that we now want to consistently have our tests use decimals, in which case, schema and test file regeneration is all that is needed here. If there is a requirement to test both decimal and floating point columns, then we can add parameterized tests for this.

misiugodfrey:
> We can regenerate the schemas and sample test files used for integration tests

@paul-aiyedun Is there a reason we want to continue using static schemas rather than relying on the data files?  To detect changes if the generation dependencies did something differently?

paul-aiyedun:
> > We can regenerate the schemas and sample test files used for integration tests
> 
> @paul-aiyedun Is there a reason we want to continue using static schemas rather than relying on the data files? To detect changes if the generation dependencies did something differently?

The integration/correctness tests cover well-defined scenarios, and in order to reliably catch regressions, the behavior they verify needs to stay stable. This is why the table schemas are kept static and committed to the codebase. One can tell exactly what schemas are under test from any given commit, rather than having to infer them from dynamically generated data. For the same reason, a small sample of the data was committed alongside the schemas, making the exact data easy to see. However, larger datasets cannot practically be stored or reviewed in the repo and so have to be generated dynamically. Even when data is generated dynamically, the schema staying static keeps what is being tested visible. Per my above comment, if the scope of the tests has changed, then we should update the test suite/framework to match the new scope.

bdice:
> If there is a requirement to test both decimal and floating point columns, then we can add parameterized tests for this.

The team discussed this a bit more today and the consensus was that we want to stop testing decimals-as-floats, we only want to test "true" decimals. I'll close this, since it doesn't seem aligned with that goal.

PR #371, opened 2026-07-12
Build in devcontainers in CI

bdice:
This enables devcontainer builds in CI for verification of builds.


.github/workflows/pr.yaml
```diff
@@ -37,27 +37,164 @@ jobs:
...
+        uses: devcontainers/ci@513af61f4de4f75d37e4438f184ba4358f0fc1ca  # v0.3.1900000450
+        with:
+          push: never
+          checkoutPath: velox-testing
+          configFile: .devcontainer/cuda${{ matrix.cuda }}/devcontainer.json
+          imageName: docker.io/rapidsai/velox-testing-devcontainer
+          imageTag: cuda${{ matrix.cuda }}-${{ matrix.arch }}
+          cacheFrom: ghcr.io/rapidsai/velox-testing/devcontainer:latest-cuda${{ matrix.cuda }}
+          env: |
+            AWS_ACCESS_KEY_ID=${{ env.AWS_ACCESS_KEY_ID }}
```
mattgara, line 173: This looks to pass a 12-hour AWS session plus an sccache auth token into code from the PR checkout. I'm not sure in what context the dev containers are supposed to be used, but could we consider moving this credentialed build behind a manually dispatched workflow on main and a protected trusted-pr-build environment, with a manually maintained trusted team as required reviewers? This is mostly to prevent someone accessing said credentials in their PR's container unbeknownst to us.

mattgara (commented):
One issue related to credentials exposure, but otherwise LGTM.

.github/workflows/pr.yaml
bdice: @trxcllnt Can you comment on this? I think I followed what we're doing everywhere in our CI workflows. Is this a concern?

trxcllnt: Yeah what @bdice has done here is fine. We do this everywhere, and the guidance is the same as the requirement for using NVIDIA's GitHub Actions runners: don't run PRs from third-party sources without reviewing them first.

mattgara (approved):

karthikeyann:
@bdice 3 CI checks are failing.  Needs addressing.

karthikeyann (approved):

PR #373, opened 2026-07-15
Add distributed CTAS output for Presto benchmarks

misiugodfrey:
## Summary

- add an opt-in `--run-as-ctas-queries` mode that executes benchmark queries as Hive CTAS operations, avoiding result-set transfer through the coordinator during measured execution
- use a worker-visible `PRESTO_CTAS_SCRATCH_DIR` for temporary distributed output while keeping source data read-only
- after timing and profiling, normalize each retained CTAS result into the same `--output-dir[/tag]/query_results/qN.parquet` layout produced by the default benchmark path
- merge multi-part Parquet output and reconstruct resolvable SQL `ORDER BY` clauses before validation
- keep validation and result posting independent of whether a query ran normally or as CTAS
- use native CTAS column aliases (`CREATE TABLE ... (c1, c2, ...) AS <original SELECT>`) without rewriting explicit SELECT expressions; wildcard projections retain their original output names
- retain the first iteration as the query result and remove later-iteration scratch tables
- update Docker and Slurm mounts, configuration examples, help text, documentation, and focused tests

## Output layout

Both execution modes now produce:

```text
<output-dir>[/<tag>]/
└── query_results/
    ├── q1.parquet
    ├── q2.parquet
    └── ...
```

In CTAS mode, workers first write temporary `qN/` datasets beneath `PRESTO_CTAS_SCRATCH_DIR`. Normalization happens after the measured query execution and before validation.

### Current revision

- [x] Focused comparison, validation, normalization, and CTAS fixture tests: `35 passed`
- [x] Multi-part normalization test produces one ordered `qN.parquet`
- [x] Single-part and empty-result normalization tests
- [x] Native CTAS alias-list tests, including TPC-H Q18
- [x] Parsed all 22 built-in TPC-H and 99 built-in TPC-DS queries
- [x] Confirmed all ten TPC-DS wildcard fallbacks expand to uniquely named columns using the bundled test schemas
- [x] Ruff on changed Python modules and tests
- [x] `bash -n` on changed benchmark and Slurm shell scripts
- [x] `run_benchmark.sh --help` exposes the renamed flag and scratch variable
- [x] `git diff --check`
- [x] Live Docker smoke using the final normalization and native CTAS alias-list implementation
- [x] Live multi-worker run confirming a multi-part `qN/` dataset is normalized into one `query_results/qN.parquet`
- [ ] Live TPC-DS CTAS run
- [x] Repeat the Slurm smoke with the final interface and normalization flow

### Earlier CTAS revision

- [x] Docker native-CPU TPC-H Q1 smoke with two CTAS iterations and separate read-only source/writable result mounts
- [x] Live TPC-H Q18 CTAS run
- [x] Docker multi-worker CTAS run producing multiple Parquet parts
- [x] Live empty-result CTAS query
- [x] Docker native-GPU worker path
- [x] Slurm deployment path
- [x] Full TPC-H query suite

### DIFF
--- a/common/testing/normalize_ctas_results.py
+++ b/common/testing/normalize_ctas_results.py
@@ -41,7 +41,9 @@ def _source_output_names(query_sql: str, physical_names: list[str]) -> list[str]
     has_wildcard = any(_is_wildcard_projection(expression) for expression in select.expressions)
     if len(select.expressions) != len(physical_names) or has_wildcard:
         return physical_names
-    return [expression.alias_or_name or physical_names[position] for position, expression in enumerate(select.expressions)]
+    return [
+        expression.alias_or_name or physical_names[position] for position, expression in enumerate(select.expressions)
+    ]
 
 
 def _write_canonical_result(query_dir: Path, target: Path, query_sql: str | None) -> None:
--- a/common/testing/normalize_ctas_results.py
+++ b/common/testing/normalize_ctas_results.py
@@ -19,7 +19,7 @@ from sqlglot import exp
 # Allow importing from the repo root when this file is executed directly.
 sys.path.insert(0, str(Path(__file__).parent.parent.parent))
 
-from common.testing.result_comparison import _get_orderby_sort_spec, _restore_orderby
+from common.testing.result_comparison import get_orderby_sort_spec, restore_orderby
 from common.testing.test_utils import get_queries
 
 
@@ -50,7 +50,7 @@ def _write_canonical_result(query_dir: Path, target: Path, query_sql: str | None
 
     if parquet_parts and query_sql is not None:
         column_names = _source_output_names(query_sql, pq.read_schema(parquet_parts[0]).names)
-        sort_spec = _get_orderby_sort_spec(query_sql, column_names)
+        sort_spec = get_orderby_sort_spec(query_sql, column_names)
 
     temporary = target.with_name(f".{target.name}.tmp-{os.getpid()}")
     try:
@@ -61,7 +61,7 @@ def _write_canonical_result(query_dir: Path, target: Path, query_sql: str | None
         else:
             frame = pd.read_parquet(query_dir)
             if sort_spec[0]:
-                frame = _restore_orderby(frame, *sort_spec)
+                frame = restore_orderby(frame, *sort_spec)
             frame.to_parquet(temporary, index=False)
         os.replace(temporary, target)
     finally:
--- a/common/testing/normalize_ctas_results.py
+++ b/common/testing/normalize_ctas_results.py
@@ -30,6 +30,13 @@ def _query_directories(source_dir: Path) -> list[Path]:
     )
 
 
+def _is_wildcard_projection(expression: exp.Expression) -> bool:
+    projection = expression.this if isinstance(expression, exp.Alias) else expression
+    return isinstance(projection, exp.Star) or (
+        isinstance(projection, exp.Column) and isinstance(projection.this, exp.Star)
+    )
+
+
 def _source_output_names(query_sql: str, physical_names: list[str]) -> list[str]:
     """Return names from an explicit SELECT projection for ORDER BY lookup."""
     # Explicit CTAS projections are stored with unique positional names (c1,
@@ -38,11 +45,7 @@ def _source_output_names(query_sql: str, physical_names: list[str]) -> list[str]
     # correct physical column positions without changing the stored schema.
     parsed = sqlglot.parse_one(query_sql, read="presto")
     select = next(parsed.find_all(exp.Select))
-    has_wildcard = any(
-        isinstance(projection := expression.this if isinstance(expression, exp.Alias) else expression, exp.Star)
-        or (isinstance(projection, exp.Column) and isinstance(projection.this, exp.Star))
-        for expression in select.expressions
-    )
+    has_wildcard = any(_is_wildcard_projection(expression) for expression in select.expressions)
     if len(select.expressions) != len(physical_names) or has_wildcard:
         return physical_names
     return [expression.alias_or_name or physical_names[position] for position, expression in enumerate(select.expressions)]
--- a/common/testing/normalize_ctas_results.py
+++ b/common/testing/normalize_ctas_results.py
@@ -1,14 +1,11 @@
-#!/usr/bin/env python3
 # SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 """Convert temporary distributed CTAS tables into canonical qN.parquet files."""
 
-import argparse
 import os
 import re
 import shutil
-import sys
 from pathlib import Path
 
 import pandas as pd
@@ -16,11 +13,7 @@ import pyarrow.parquet as pq
 import sqlglot
 from sqlglot import exp
 
-# Allow importing from the repo root when this file is executed directly.
-sys.path.insert(0, str(Path(__file__).parent.parent.parent))
-
 from common.testing.result_comparison import get_orderby_sort_spec, restore_orderby
-from common.testing.test_utils import get_queries
 
 
 def _query_directories(source_dir: Path) -> list[Path]:
@@ -93,32 +86,3 @@ def normalize_ctas_results(
         shutil.rmtree(query_dir)
         normalized.append(target)
     return normalized
-
-
-def parse_args() -> argparse.Namespace:
-    parser = argparse.ArgumentParser(description=__doc__)
-    parser.add_argument("--source-dir", required=True, type=Path, help="Directory containing temporary qN CTAS tables.")
-    parser.add_argument(
-        "--output-dir",
-        required=True,
-        type=Path,
-        help="Canonical query_results directory that will receive qN.parquet files.",
-    )
-    parser.add_argument("--benchmark-type", required=True, choices=["tpch", "tpcds"])
-    parser.add_argument("--queries-file", default=None, help="Optional custom query-definition JSON file.")
-    return parser.parse_args()
-
-
-if __name__ == "__main__":
-    args = parse_args()
-    try:
-        outputs = normalize_ctas_results(
-            args.source_dir,
-            args.output_dir,
-            get_queries(args.benchmark_type, args.queries_file),
-        )
-    except (OSError, ValueError) as error:
-        print(f"Error: Failed to normalize CTAS results: {error}", file=sys.stderr)
-        sys.exit(1)
-    for output in outputs:
-        print(f"Normalized CTAS result: {output}")
--- a/common/testing/normalize_ctas_results.py
+++ b/common/testing/normalize_ctas_results.py
@@ -32,6 +32,10 @@ def _query_directories(source_dir: Path) -> list[Path]:
 
 def _source_output_names(query_sql: str, physical_names: list[str]) -> list[str]:
     """Return names from an explicit SELECT projection for ORDER BY lookup."""
+    # Explicit CTAS projections are stored with unique positional names (c1,
+    # c2, ...), but ORDER BY still refers to names from the original SELECT.
+    # Recover those logical names so ORDER BY expressions can be mapped to the
+    # correct physical column positions without changing the stored schema.
     parsed = sqlglot.parse_one(query_sql, read="presto")
     select = next(parsed.find_all(exp.Select))
     has_wildcard = any(
--- a/common/testing/normalize_ctas_results.py
+++ b/common/testing/normalize_ctas_results.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Convert temporary distributed CTAS tables into canonical qN.parquet files."""
+
+import argparse
+import os
+import re
+import shutil
+import sys
+from pathlib import Path
+
+import pandas as pd
+import pyarrow.parquet as pq
+import sqlglot
+from sqlglot import exp
+
+# Allow importing from the repo root when this file is executed directly.
+sys.path.insert(0, str(Path(__file__).parent.parent.parent))
+
+from common.testing.result_comparison import _get_orderby_sort_spec, _restore_orderby
+from common.testing.test_utils import get_queries
+
+
+def _query_directories(source_dir: Path) -> list[Path]:
+    return sorted(
+        (path for path in source_dir.iterdir() if path.is_dir() and re.fullmatch(r"q\d+", path.name)),
+        key=lambda path: int(path.name[1:]),
+    )
+
+
+def _source_output_names(query_sql: str, physical_names: list[str]) -> list[str]:
+    """Return names from an explicit SELECT projection for ORDER BY lookup."""
+    parsed = sqlglot.parse_one(query_sql, read="presto")
+    select = next(parsed.find_all(exp.Select))
+    has_wildcard = any(
+        isinstance(projection := expression.this if isinstance(expression, exp.Alias) else expression, exp.Star)
+        or (isinstance(projection, exp.Column) and isinstance(projection.this, exp.Star))
+        for expression in select.expressions
+    )
+    if len(select.expressions) != len(physical_names) or has_wildcard:
+        return physical_names
+    return [expression.alias_or_name or physical_names[position] for position, expression in enumerate(select.expressions)]
+
+
+def _write_canonical_result(query_dir: Path, target: Path, query_sql: str | None) -> None:
+    parquet_parts = sorted(query_dir.rglob("*.parquet"))
+    sort_spec: tuple[list[int], list[bool], list[bool]] = ([], [], [])
+
+    if parquet_parts and query_sql is not None:
+        column_names = _source_output_names(query_sql, pq.read_schema(parquet_parts[0]).names)
+        sort_spec = _get_orderby_sort_spec(query_sql, column_names)
+
+    temporary = target.with_name(f".{target.name}.tmp-{os.getpid()}")
+    try:
+        # A single unordered part is already a canonical Parquet file. Copying
+        # it avoids decoding and rewriting large results.
+        if len(parquet_parts) == 1 and not sort_spec[0]:
+            shutil.copy2(parquet_parts[0], temporary)
+        else:
+            frame = pd.read_parquet(query_dir)
+            if sort_spec[0]:
+                frame = _restore_orderby(frame, *sort_spec)
+            frame.to_parquet(temporary, index=False)
+        os.replace(temporary, target)
+    finally:
+        temporary.unlink(missing_ok=True)
+
+
+def normalize_ctas_results(
+    source_dir: Path,
+    output_dir: Path,
+    queries: dict[str, str],
+) -> list[Path]:
+    """Convert each committed qN CTAS table into one canonical Parquet file."""
+    query_dirs = _query_directories(source_dir)
+    if not query_dirs:
+        raise ValueError(f"No committed CTAS result tables found in {source_dir}")
+
+    output_dir.mkdir(parents=True, exist_ok=True)
+    normalized: list[Path] = []
+    for query_dir in query_dirs:
+        target = output_dir / f"{query_dir.name}.parquet"
+        _write_canonical_result(query_dir, target, queries.get(query_dir.name.upper()))
+        shutil.rmtree(query_dir)
+        normalized.append(target)
+    return normalized
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--source-dir", required=True, type=Path, help="Directory containing temporary qN CTAS tables.")
+    parser.add_argument(
+        "--output-dir",
+        required=True,
+        type=Path,
+        help="Canonical query_results directory that will receive qN.parquet files.",
+    )
+    parser.add_argument("--benchmark-type", required=True, choices=["tpch", "tpcds"])
+    parser.add_argument("--queries-file", default=None, help="Optional custom query-definition JSON file.")
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+    args = parse_args()
+    try:
+        outputs = normalize_ctas_results(
+            args.source_dir,
+            args.output_dir,
+            get_queries(args.benchmark_type, args.queries_file),
+        )
+    except (OSError, ValueError) as error:
+        print(f"Error: Failed to normalize CTAS results: {error}", file=sys.stderr)
+        sys.exit(1)
+    for output in outputs:
+        print(f"Normalized CTAS result: {output}")
--- a/common/testing/result_comparison.py
+++ b/common/testing/result_comparison.py
@@ -446,7 +446,6 @@ def compare_result_frames(
     expected: pd.DataFrame,
     query_sql: str,
     actual_column_types: list[str] | None = None,
-    actual_order_preserved: bool = True,
 ) -> None:
     """
     Full comparison pipeline. Raises AssertionError on any mismatch.
@@ -458,8 +457,7 @@ def compare_result_frames(
       1. Validate column count and capture expected column names.
       2. Normalize dtypes / value types.
       3. Validate row count.
-      4. Parse ORDER BY / LIMIT from SQL and reconstruct logical order when
-         the actual rows came from unordered distributed storage.
+      4. Parse ORDER BY / LIMIT from SQL.
       5. Validate each engine respected ORDER BY (per-frame, strict equality).
       6. With LIMIT: peel the tied tail in engine order, then canonical-sort
          each piece. Compare front fully; compare tail's ORDER BY values only.
@@ -482,13 +480,9 @@ def compare_result_frames(
     if len(actual) != len(expected):
         raise AssertionError(f"Row count mismatch: {len(actual)} (actual) vs {len(expected)} (expected)")
 
-    # 4. Parse ORDER BY / LIMIT. A distributed Parquet dataset has no physical
-    # row-order guarantee, so reconstruct its logical SQL order before applying
-    # the same ordered-result checks used for coordinator-returned results.
-    sort_col_indices, ascending, nulls_first = _get_orderby_sort_spec(query_sql, expected_col_names)
+    # 4. Parse ORDER BY / LIMIT.
+    sort_col_indices, ascending, _ = _get_orderby_sort_spec(query_sql, expected_col_names)
     limit = get_limit(query_sql)
-    if not actual_order_preserved:
-        actual = _restore_orderby(actual, sort_col_indices, ascending, nulls_first)
 
     # 5. Per-frame ORDER BY validation — engine order is intact here.
     _validate_orderby(actual, sort_col_indices, ascending)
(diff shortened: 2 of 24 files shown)


common/testing/validate_results.py
```diff
@@ -50,16 +51,22 @@ def validate(
     passed = failed = not_validated = expected_failures = 0
 
     if query_numbers is not None:
-        result_files = sorted(f for q in query_numbers for f in [results_dir / f"q{q}.parquet"] if f.exists())
+        result_files = []
+        for q_num in query_numbers:
```
misiugodfrey: Multiple workers can each write their own Parquet part file, so we now have the results as a directory.

common/testing/result_comparison.py
```diff
@@ -34,22 +34,19 @@
...
-    """
-    Extract ORDER BY column positions and directions from SQL using sqlglot.
-
-    Returns (indices, ascending) where ascending[i] is True iff indices[i]
-    is sorted ASC. Returns ([], []) when there is no ORDER BY, or when any
-    ORDER BY expression is too complex to map to a result column (CASE,
-    aggregate, etc.).
-    """
-    expr = sqlglot.parse_one(query_sql)
+def _get_orderby_sort_spec(
```
misiugodfrey: We were able to skip null handling when we had a single parquet file, but now since we are merging the files back based on order-by info, we need to know where nulls should go.

common/testing/result_comparison.py
```diff
@@ -288,6 +299,32 @@ def _validate_orderby(df: pd.DataFrame, sort_col_indices: list[int], ascending:
         prior_changed = prior_changed | value_changed
 
 
+def _restore_orderby(
```
misiugodfrey: To merge multiple parquet files we now need to know the hierarchy of the order by columns from the original sql query.

misiugodfrey:
@devavret I ran some performance benchmarks using these changes to make sure I saw the same performance increase you detailed in the issue on Q11.  I saw about a %25 reduction in time on this query, although this was running a 4-gpu tpch-1k run with an older image; so I'm not sure how much that differs from your original observations.
From the issue, it looks like we should be expecting much more of a gain, although I don't know exactly what context the POC was in, so let me know if further changes are needed to realize the full gain.

presto/scripts/run_benchmark.sh
```diff
@@ -43,6 +43,8 @@ OPTIONS:
     --profile-script-path   Path to a custom profiler functions script. Defaults to ./profiler_functions.sh.
     --skip-drop-cache       Skip dropping system caches before each benchmark query (dropped by default).
     --skip-analyze-check    Skip checking that ANALYZE TABLE has been run on all tables (checked by default).
+    --write-results-to-file Write query results with distributed Hive CTAS operations instead of returning them
```
paul-aiyedun: Do we plan on running/publishing benchmarks for the CTAS queries?

misiugodfrey: That would be a question for @devavret .  I believe we are allowed to, but I don't know if we have explicitly declared intentions.

devavret: Yes, that's the primary use case. The reason we want this CTAS option is to speed up two queries.

devavret:
> @devavret I ran some performance benchmarks using these changes to make sure I saw the same performance increase you detailed in the issue on Q11. I saw about a %25 reduction in time on this query, although this was running a 4-gpu tpch-1k run with an older image; so I'm not sure how much that differs from your original observations.
> From the issue, it looks like we should be expecting much more of a gain, although I don't know exactly what context the POC was in, so let me know if further changes are needed to realize the full gain.

@misiugodfrey If you see Q11 going from 50s to 3s on sf30k 8 nodes then this PR is sufficient.

devavret (approved):

common/testing/result_comparison.py
```diff
@@ -76,10 +73,25 @@ def get_orderby_col_indices(query_sql: str, expected_col_names: list[str]) -> tu
                 f"ORDER BY expression {key.sql()!r} couldn't be mapped to a result column; "
                 "ORDER BY validation and tie-boundary handling will be skipped for this query."
             )
-            return [], []
+            return [], [], []
 
         sort_col_indices.append(resolved_idx)
         ascending.append(not is_desc)
+        nulls_first.append(bool(ordered.args.get("nulls_first")))
```
patdevinwilson, line 80: `bool(ordered.args.get("nulls_first"))` treats unspecified NULLS as `False` (NULLS LAST). Presto's default is NULLS LAST for ASC but **NULLS FIRST for DESC**. So for `ORDER BY score DESC` (no explicit NULLS clause), restore will put nulls last while the engine put them first, and the subsequent `_validate_orderby` / LIMIT tie handling can disagree with expected.

Suggest something like:
```python
nulls_first_arg = ordered.args.get("nulls_first")
if nulls_first_arg is None:
    nulls_first.append(bool(is_desc))  # Presto default
else:
    nulls_first.append(bool(nulls_first_arg))
```
And add a test for bare `ORDER BY x DESC` with nulls.

presto/scripts/run_benchmark.sh
```diff
@@ -321,6 +354,55 @@ echo "Using PRESTO_IMAGE_TAG: $PRESTO_IMAGE_TAG"
 BENCHMARK_TEST_DIR=${TEST_DIR}/performance_benchmarks
 pytest -q -s ${BENCHMARK_TEST_DIR}/${BENCHMARK_TYPE}_test.py ${PYTEST_ARGS[*]}
```
patdevinwilson: `set -e` is on (line 6). If pytest fails, we never reach `normalize_ctas_result_permissions` or validation — and for CTAS that also means permission fixup / result posting from Slurm may see unreadable files.

Same pattern as #382 would help:
```bash
pytest_exit=0
pytest ... || pytest_exit=$?
# ... normalize + validate ...
exit ${pytest_exit}
```

presto/docker/config/template/etc_common/catalog/hive_output.properties
```diff
@@ -0,0 +1,10 @@
+# Hive catalog dedicated to benchmark CTAS output. Its file metastore root is
+# mounted separately from PRESTO_DATA_DIR so source data can remain read-only.
+connector.name=hive-hadoop2
+hive.metastore=file
+hive.metastore.catalog.dir=file:/var/lib/presto/data/hive/benchmark_output
+hive.allow-drop-table=true
+# Native workers otherwise stage writes under their container-private /tmp,
+# which prevents the coordinator from committing the files. Write directly to
+# the shared result mount; interrupted runs are cleaned before the next run.
+hive.temporary-staging-directory-enabled=false
```
patdevinwilson, line 10: Good call disabling temporary staging so native workers write straight to the shared mount. Worth confirming (maybe already did) that coordinator commit still works for multi-worker CTAS with this setting — if an interrupted run leaves partial `qN/` dirs, the fixture's `shutil.rmtree` + DROP TABLE path looks covered.

presto/docker/docker-compose.common.yml
```diff
@@ -3,7 +3,8 @@ services:
     volumes:
       - ./.hive_metastore:/var/lib/presto/data/hive/metastore
       - ../../common/testing/integration_tests/data:/var/lib/presto/data/hive/data/integration_test
-      - ${PRESTO_DATA_DIR:-/dev/null}:/var/lib/presto/data/hive/data/user_data
+      - ${PRESTO_DATA_DIR:-/dev/null}:/var/lib/presto/data/hive/data/user_data:ro
```
patdevinwilson, line 6: Making `PRESTO_DATA_DIR` `:ro` is a nice hardening win for the CTAS split. Just flagging: any existing workflow that still wrote into `user_data` (not via `hive_output`) will now fail — hopefully none remain.

presto/testing/performance_benchmarks/common_fixtures.py
```diff
@@ -124,6 +268,8 @@ def benchmark_query_function(query_id):
                     parquet_path = results_dir / f"{query_id.lower()}.parquet"
                     df.to_parquet(parquet_path, index=False)
 
+                result.append(cursor.stats["elapsedTimeMillis"])
```
patdevinwilson: Timing is now recorded **after** `fetchall()` for CTAS (good — waits for commit). For non-CTAS, `elapsedTimeMillis` is also after fetch, matching prior behavior for iteration 0 and changing later iterations (previously timed before fetch on iter>0). Intentional? If the goal is "time until results ready", this is better; if comparing to historical non-CTAS numbers for iter>0, they may shift slightly.

common/testing/validate_results.py
```diff
@@ -101,7 +120,13 @@ def validate(
             not_validated += 1
             continue
 
-        status, msg = validate_query_result(query_id, actual, expected, query_sql)
+        status, msg = validate_query_result(
+            query_id,
+            actual,
+            expected,
+            query_sql,
+            actual_order_preserved=result_file.is_file(),
```
patdevinwilson: `actual_order_preserved=result_file.is_file()` is a clean way to distinguish single-file vs distributed dataset dirs. For an empty CTAS table that only has `.prestoSchema` (your q15 test), a brief comment near the `read_parquet` call about directory/empty-schema behavior would help future readers.

Also: CI is failing `ruff format` on this file and `performance_benchmark_fixtures_test.py` — a format pass should unblock merge.

patdevinwilson (changes requested):
Solid feature overall — separate writable `hive_output` catalog + reconstructing ORDER BY for unordered Parquet datasets is the right shape. Blocking items: style CI (`ruff format`) is red, and I think implicit NULLS placement for DESC is wrong in `_restore_orderby`. Details inline.

presto/scripts/run_benchmark.sh
```diff
@@ -57,6 +59,8 @@ OPTIONS:
                             Set to the full host path of the expected results directory
                             (e.g. PRESTO_EXPECTED_RESULTS_DIR=/data/sf100_expected).
                             Warning (not error) if set but the directory does not exist.
+    PRESTO_OUTPUT_DIR       Writable host directory used for distributed CTAS result datasets. This must be set
```
paul-aiyedun: We already write outputs (including result sets) to `--output-dir`. I think adding another `PRESTO_OUTPUT_DIR` variable here can be confusing.

presto/scripts/run_benchmark.sh
paul-aiyedun: Can we use a more direct parameter name e.g. `--run-as-ctas-queries`?

presto/scripts/run_benchmark.sh
```diff
@@ -232,6 +241,26 @@ if [[ -z ${SCHEMA_NAME} ]]; then
   exit 1
 fi
 
+if [[ "${WRITE_RESULTS_TO_FILE}" == "true" ]]; then
```
paul-aiyedun: Similar to the above [comment](https://github.com/rapidsai/velox-testing/pull/373/changes#r3667491771), we currently write to `--output-dir`, so why do we need special handling here and in `normalize_ctas_result_permissions`?

common/testing/validate_results.py
paul-aiyedun: Can we instead normalize the Parquet files and not have the validation logic be concerned with the details of how the result set is generated?

common/testing/result_comparison.py
paul-aiyedun: I believe this [comment](https://github.com/rapidsai/velox-testing/pull/373/changes#r3667557596) also applies here.

common/testing/result_comparison.py
misiugodfrey: Are you sure Presto changes NULLS FIRST/LAST based on the order?  I was under the impression it was always last unless explicitly changed and as far as I can tell the documentation supports this: https://prestodb.github.io/docs/current/sql/select.html#order-by-clause:~:text=The%20default%20null%20ordering%20is%20NULLS%20LAST%2C%20regardless%20of%20the%20ordering%20direction.

If I'm missing something (or looking at the wrong docs), please link me to the appropriate documentation and I will change it.

presto/testing/ctas_test.py
paul-aiyedun, line 1: I don't think it is common to have tests for test fixtures?

common/testing/result_comparison.py
patdevinwilson: You're right — I was wrong here. Presto's docs are clear:

> The default null ordering is NULLS LAST, regardless of the ordering direction.

(https://prestodb.github.io/docs/current/sql/select.html#order-by-clause)

I was mixing that up with other engines (e.g. Postgres / Spark defaults that do flip with ASC/DESC). For Presto, `bool(ordered.args.get("nulls_first"))` → `False` when unspecified correctly matches NULLS LAST. Thanks for the correction.

presto/docker/docker-compose.common.yml
misiugodfrey: I don't think any of our existing workflows should be writing to user_data.  Our data-gen workflow acts outside of this container approach, so we should be good.

If anything was writing to the data like this, I think it would be considered a bug.

patdevinwilson: Makes sense — thanks for confirming. If something *was* writing through this mount, treating that as a bug (and failing loudly under `:ro`) is the right call. No change needed from me.

presto/testing/performance_benchmarks/common_fixtures.py
misiugodfrey: I'll preserve the original behaviour when we are not using CTAS.

presto/testing/performance_benchmarks/common_fixtures.py
```diff
@@ -1,20 +1,149 @@
...
+
+
+@dataclass(frozen=True)
+class CtasResults:
+    catalog: str
+    schema: str
+
+
+def ctas_schema_name(tag):
+    if tag and not re.fullmatch(r"[A-Za-z0-9_]+", tag):
```
paul-aiyedun: What is the difference between the regular benchmark tag and this tag?

presto/testing/performance_benchmarks/common_fixtures.py
```diff
@@ -1,20 +1,149 @@
...
+        raise ValueError("CTAS result tags must contain only alphanumeric and underscore characters")
+    return f"benchmark_results_{tag.lower()}" if tag else "benchmark_results"
+
+
+def strip_trailing_semicolon(query):
+    return query.rstrip().removesuffix(";").rstrip()
+
+
+def ctas_table_name(query_id, iteration_num):
```
paul-aiyedun: Are we creating a new table for each iteration?

presto/testing/performance_benchmarks/common_fixtures.py
```diff
@@ -1,20 +1,149 @@
...
+    return query_id.lower() if iteration_num == 0 else f"{query_id.lower()}_iteration_{iteration_num + 1}"
+
+
+def alias_ctas_output_expressions(query):
```
paul-aiyedun: Can you please expand on the reason for this? I would have thought that CTAS query construction would be more straightforward i.e. `CREATE TABLE ... AS {original SELECT query}`.

presto/docker/config/template/etc_common/catalog/hive_output.properties
paul-aiyedun, line 1: How do the settings here interact with `etc_worker/catalog/hive.properties` and `etc_coordinator/catalog/hive.properties`?

paul-aiyedun (commented):
I think adding a CTAS execution mode in order to understand exchange overhead makes sense. However, I think there is some interface divergence in the current implementation.

presto/testing/performance_benchmarks/common_fixtures.py
misiugodfrey: That was the original approach, but some of the queries contained result sets with unnamed expressions (legal) but can't be expressed in CTAS because it needs unique column names.

For instance, TPCH Q18 has:
```
SELECT
      c_name,
      c_custkey,
      o_orderkey,
      o_orderdate,
      o_totalprice,
      sum(l_quantity)
 ```
 
 And that `sum(l_quantity)` needs an explicit column name.  This function is handling that.

Comment (paul-aiyedun) [reply presto/testing/performance_benchmarks/common_fixtures.py]: Got it. Do we care about the specific column names? If not, I think we can simplify this logic by changing the CREATE TABLE clause i.e. `CREATE TABLE {table name} (c1, c2, c3, ...) AS {original SELECT query}`. This also reduces the chances of unexpected changes to the SELECT query.

File: presto/scripts/run_benchmark.sh (hunk 11 above)
Comment (misiugodfrey) [reply presto/scripts/run_benchmark.sh]: I can rename this to something separate `PRESTO_CTAS_SCRATCH_DIR`, but the problem is this directory needs to be mounted in the container when it starts - so using `--output-dir` isn't viable, since that variable isn't specified by the user until the containers are already running (at least in the docker path).

File: presto/scripts/run_benchmark.sh (hunk 4 above)
Comment (misiugodfrey) [reply presto/scripts/run_benchmark.sh]: Sounds good.

**Comment by misiugodfrey** [comment]
I've updated the PR to have the CTAS results stored in a scratch location and then normalize them to a single parquet file (if multiple files were created).  This should keep the rest of the verification unchanged, and adds a separate "result normalization" step that should run rarely (we should not often create multiple result files).

File: presto/docker/config/template/etc_common/catalog/hive_output.properties
Comment (misiugodfrey) [reply presto/docker/config/template/etc_common/catalog/hive_output.properties:1]: It's a separate catalog and should not interact with them at all beyond inheriting some shared settings.

File: presto/testing/performance_benchmarks/common_fixtures.py (hunk 13 above)
Comment (misiugodfrey) [reply presto/testing/performance_benchmarks/common_fixtures.py]: It's the same tag; it's just used here for separating potentially concurrent runs.  Probably not a major concern, so I can remove it for simplicity.

File: presto/testing/performance_benchmarks/common_fixtures.py (hunk 14 above)
Comment (misiugodfrey) [reply presto/testing/performance_benchmarks/common_fixtures.py]: Yes.  The first table is retained for normalization and validation and every other iteration creates and then drops a table.

File: presto/testing/ctas_test.py
Comment (misiugodfrey) [reply presto/testing/ctas_test.py:1]: These tests aren't really "testing the test", so much as testing the aliasing, naming, and schema logic that happen to be used in that fixture.  I have found it useful to catch regressions with these - but perhaps we should organize them differently?

File: presto/scripts/run_benchmark.sh [hunk 16]
```diff
@@ -232,6 +244,21 @@ if [[ -z ${SCHEMA_NAME} ]]; then
   exit 1
 fi
 
+if [[ "${RUN_AS_CTAS_QUERIES}" == "true" ]]; then
+  if [[ -z "${PRESTO_CTAS_SCRATCH_DIR}" ]]; then
+    echo "Error: PRESTO_CTAS_SCRATCH_DIR must be set when --run-as-ctas-queries is enabled." >&2
+    exit 1
+  fi
+  mkdir -p "${PRESTO_CTAS_SCRATCH_DIR}"
```
paul-aiyedun: Should `PRESTO_CTAS_SCRATCH_DIR` already exist (e.g. from the server startup bind)?

presto/scripts/run_benchmark.sh
```diff
@@ -331,6 +413,21 @@ else
...
+    --output-dir "${EFFECTIVE_BENCHMARK_DIR}/query_results"
+    --benchmark-type "${BENCHMARK_TYPE}"
+  )
+  if [[ -n ${QUERIES_FILE} ]]; then
+    NORMALIZE_ARGS+=(--queries-file "${QUERIES_FILE}")
+  fi
+  echo "[CTAS] Normalizing temporary results into ${EFFECTIVE_BENCHMARK_DIR}/query_results"
+  python "${SCRIPT_DIR}/../../common/testing/normalize_ctas_results.py" \
+    "${NORMALIZE_ARGS[@]}" || CTAS_NORMALIZATION_EXIT=$?
+fi
```
paul-aiyedun: Can this be done as a tear down step of the benchmark test suite?

presto/scripts/run_benchmark.sh
```diff
@@ -350,7 +447,10 @@ VALIDATE_REQUIREMENTS="${SCRIPT_DIR}/../testing/requirements.txt"
 # Resolve reference results directory.
 # PRESTO_EXPECTED_RESULTS_DIR env var is the implicit fallback (warning if missing).
 # Explicit --reference-results-dir was already validated before the benchmark ran.
-if [[ -n ${PRESTO_EXPECTED_RESULTS_DIR} && ! -d ${PRESTO_EXPECTED_RESULTS_DIR} ]]; then
+VALIDATION_EXIT=0
+if [[ "${RUN_AS_CTAS_QUERIES}" == "true" && ( ${CTAS_PERMISSION_EXIT} -ne 0 || ${CTAS_NORMALIZATION_EXIT} -ne 0 ) ]]; then
+  echo "[Validation] Skipped because CTAS result normalization failed."
```
paul-aiyedun: In what case is this expected?

presto/scripts/run_benchmark.sh
```diff
@@ -319,7 +350,58 @@ fi
 echo "Using PRESTO_IMAGE_TAG: $PRESTO_IMAGE_TAG"
 
 BENCHMARK_TEST_DIR=${TEST_DIR}/performance_benchmarks
-pytest -q -s ${BENCHMARK_TEST_DIR}/${BENCHMARK_TYPE}_test.py ${PYTEST_ARGS[*]}
+PYTEST_EXIT=0
+pytest -q -s ${BENCHMARK_TEST_DIR}/${BENCHMARK_TYPE}_test.py ${PYTEST_ARGS[*]} || PYTEST_EXIT=$?
+
+ensure_ctas_scratch_readable() {
```
paul-aiyedun: Out of curiosity, would executing `chown -R $USER:$USER $PRESTO_CTAS_SCRATCH_DIR` have the same effect?

presto/scripts/run_multiple_benchmarks.sh
```diff
@@ -76,13 +81,21 @@ if [[ -z ${PRESTO_DATA_DIR} ]]; then
     exit 1
 fi
 
+if [[ "${RUN_AS_CTAS_QUERIES}" == "true" && -z "${PRESTO_CTAS_SCRATCH_DIR:-}" ]]; then
+    echo "Error: PRESTO_CTAS_SCRATCH_DIR is required with --run-as-ctas-queries."
+    exit 1
+fi
+
+BENCHMARK_OUTPUT_ARGS=()
+[[ "${RUN_AS_CTAS_QUERIES}" == "true" ]] && BENCHMARK_OUTPUT_ARGS+=(--run-as-ctas-queries)
```
paul-aiyedun: Nit: Consistently use if statements for this type of logic.

presto/docker/docker-compose.common.yml
```diff
@@ -3,7 +3,8 @@ services:
...
+      - ${PRESTO_CTAS_SCRATCH_DIR:-/dev/null}:/var/lib/presto/data/hive/benchmark_output
```
paul-aiyedun: Nit: Consistently use CTAS scratch i.e. `ctas_scratch_output` instead of `benchmark_output`.

presto/testing/ctas_test.py
paul-aiyedun: I think it would make sense to separate out that logic into its own module with unit tests.

common/testing/normalize_ctas_results.py
```diff
@@ -0,0 +1,121 @@
...
+    """Return names from an explicit SELECT projection for ORDER BY lookup."""
+    # Explicit CTAS projections are stored with unique positional names (c1,
+    # c2, ...), but ORDER BY still refers to names from the original SELECT.
+    # Recover those logical names so ORDER BY expressions can be mapped to the
+    # correct physical column positions without changing the stored schema.
+    parsed = sqlglot.parse_one(query_sql, read="presto")
+    select = next(parsed.find_all(exp.Select))
+    has_wildcard = any(
+        isinstance(projection := expression.this if isinstance(expression, exp.Alias) else expression, exp.Star)
+        or (isinstance(projection, exp.Column) and isinstance(projection.this, exp.Star))
```
paul-aiyedun: Nit: Can we unwrap this in a function? The nesting makes it a little bit hard to follow.

presto/testing/performance_benchmarks/common_fixtures.py
```diff
@@ -1,20 +1,119 @@
...
+    """Return unique CTAS aliases without modifying the original SELECT.
+
+    A wildcard's expanded column count depends on source schema metadata, so
+    wildcard projections keep their original output names instead.
+    """
+    parsed = sqlglot.parse_one(strip_trailing_semicolon(query), read="presto")
+    select = next(parsed.find_all(exp.Select))
+    if any(
+        isinstance(projection := expression.this if isinstance(expression, exp.Alias) else expression, exp.Star)
+        or (isinstance(projection, exp.Column) and isinstance(projection.this, exp.Star))
```
paul-aiyedun: Same comment about moving this logic to a function.

presto/testing/performance_benchmarks/common_fixtures.py
```diff
@@ -100,29 +199,52 @@ def benchmark_query(request, presto_cursor, benchmark_queries, benchmark_result_
...
+                    if iteration_num == 0:
+                        rows = cursor.fetchall()
+                        columns = [desc[0] for desc in cursor.description]
+                        df = pd.DataFrame(rows, columns=columns)
+
+                        # Save to Parquet format to match expected results
+                        results_dir = Path(f"{bench_output_dir}/query_results")
+                        results_dir.mkdir(parents=True, exist_ok=True)
+                        parquet_path = results_dir / f"{query_id.lower()}.parquet"
+                        df.to_parquet(parquet_path, index=False)
```
paul-aiyedun: Can we move this to a function?

paul-aiyedun (commented):
Changes overall look good to me. I just had a few code cleanup comments.

patdevinwilson (approved):

presto/scripts/run_benchmark.sh
misiugodfrey: Makes sense.  I'll remove the mkdir as that can hide a missing docker mount.

presto/scripts/run_benchmark.sh
misiugodfrey: I think that's a good suggestion, but it runs into issues because of the docker permissions.  Velox is writing these files in a mode where pytest can't read them unless we repair the permissions in run_benchmark.sh.

Ideally we could fix this by updating the docker configs, but it would not be trivial and there are a lot of potential side-effects we would have to sort through.

So I think this  is something to pursue, but as part of a different item that has an appropriate scope, because it requires it's own investigation.

presto/scripts/run_benchmark.sh
misiugodfrey: Not "expected", but it could happen if we ran out of disk space, or there was a docker permissions issue, etc...

We don't want validation to "pass" because something failed and we read left-over results from an earlier run or something similar.

presto/scripts/run_benchmark.sh
misiugodfrey: It could, but I think it's less reliable (there are more assumptions about the system config, I think).  Is there a particular reason you would prefer `chown`?

presto/scripts/prepare_ctas_results.sh
```diff
@@ -0,0 +1,55 @@
...
+
+has_committed_table() {
+  find "${HOST_RESULTS_DIR}" -mindepth 1 -maxdepth 1 -type d -name 'q[0-9]*' \
+    -exec test -f '{}/.prestoSchema' \; -print -quit 2>/dev/null
+}
+
+has_inaccessible_results() {
+  [[ -n "$(find "${HOST_RESULTS_DIR}" -type d ! -writable -print -quit 2>/dev/null || true)" \
+    || -n "$(find "${HOST_RESULTS_DIR}" -type f -name '*.parquet' ! -readable -print -quit 2>/dev/null || true)" ]]
+}
```
paul-aiyedun, line 30: Please add comments that explain what the functions are doing and why they are needed.

paul-aiyedun (approved):

PR #375, opened 2026-07-16
[Wrong remote] Enable AWS S3 remote data source for Presto GPU benchmarks

kingcrimsontianyu:
This PR lets Presto GPU TPC-H/TPC-DS benchmarks and integration tests run against data in a remote object store (AWS S3) with no local copy of the dataset. Only the Hive metastore stays local.


kingcrimsontianyu:
Wrong remote used.

PR #376, opened 2026-07-16
Enable AWS S3 remote data source for Presto GPU benchmarks

kingcrimsontianyu:
This PR lets Presto GPU TPC-H/TPC-DS benchmarks and integration tests run against data in a remote object store (AWS S3) with no local copy of the dataset. Only the Hive metastore stays local.

The key changes include:
- Register tables for remote data: A new `presto/scripts/register_external_tables.sh` registers Hive tables at a URI base without generating or copying data. `create_hive_tables.py` gains `--external-location-base`. When set, each table's `EXTERNAL_LOCATION` is `<base>/<table>`. Otherwise it falls back to the local `file:` path.
- Update schema templates: The 8 TPC-H + 24 TPC-DS `.sql` files now use `EXTERNAL_LOCATION = '{location}'` instead of the hardcoded `'file:{file_path}'`.
- Add AWS credentials and region to docker files: The `docker-compose.common.yml` file passes `AWS_ACCESS_KEY_ID/SECRET_ACCESS_KEY/SESSION_TOKEN` and `AWS_DEFAULT_REGION`/`AWS_REGION` to the coordinator and native worker.
- Enable DuckDB S3: The new `duckdb_utils.ensure_remote_access` function configures httpfs and S3 `credential_chain` secret. `read_scale_factor` derives the scale factor from `metadata.json` (local via stdlib `json`, and S3 via DuckDB). This allows the integration tests to continue working for remote data source.
- Update the documentation.


simoneves (commented):
LGTM but will test properly

kingcrimsontianyu:
Sure! This has been tested on my laptop with S3 TPC-H/TPC-DS SF-1. The planned next step is to test on an EC2 instance with S3 SF-1K. 
Another thing is that I've only checked the cuDF/KvikIO data source path, and I'll need to test Velox's buffered input path (which uses AWS SDK for S3 access).

presto/testing/integration_tests/create_hive_tables.py
```diff
@@ -7,17 +7,19 @@
...
     for table_name, schema in schemas:
-        presto_cursor.execute(
-            schema.format(
-                file_path=f"/var/lib/presto/data/hive/data/{data_sub_directory}/{table_name}", schema=schema_name
-            )
-        )
+        # When external_location_base is set (e.g. s3://bucket/prefix/sf100), point the
+        # table at that base. Otherwise fall back to the local bind-mounted file path.
+        if external_location_base:
+            location = f"{external_location_base}/{table_name}"
```
shrshi, line 19: Do the trailing backslashes need to be stripped here similar to what you have in L38 in `generate_table_schemas.py`?

presto/docker/config/template/etc_worker/catalog/hive.properties
```diff
@@ -13,3 +13,7 @@ hive.metastore.catalog.dir=file:/var/lib/presto/data/hive/metastore
 # Allow DROP TABLE statements. Enabled to make smoke/perf tests able to reset
 # state and clean up artifacts without manual intervention.
 hive.allow-drop-table=true
+
+# AWS S3.
+hive.s3.ssl.enabled=true
+hive.s3.path-style-access=false
```
paul-aiyedun, line 19: Was `hive.s3.use-instance-credentials=false` intentionally not added here?

common/testing/integration_tests/test_utils.py
```diff
@@ -71,7 +71,10 @@ def write_query_engine_rows(output_dir, result_file_name, rows, columns, query_e
 
 
 def create_duckdb_table(table_name, data_path):
-    create_table(table_name, get_abs_file_path(__file__, data_path))
+    # Only resolve local path
+    if "://" not in data_path:
```
paul-aiyedun, line 75: Consider adding a direct flag parameter e.g. `is_s3_location` here instead of the `://` check.

benchmark_data_tools/generate_table_schemas.py
```diff
@@ -74,12 +96,42 @@ def get_column_definition(column_metadata):
...
         type=str,
-        required=True,
-        help="The name of the directory that contains the benchmark data.",
+        required=False,
+        default=None,
+        help="Path to a local directory that contains the benchmark data (one subdirectory per "
+        "table). Mutually exclusive with --external-location-base.",
+    )
+    parser.add_argument(
+        "--external-location-base",
```
paul-aiyedun, line 105: Consider a more specific option name like `s3-data-dir-path`.

Same comment applies to other files with a similar flag.

benchmark_data_tools/generate_table_schemas.py
```diff
@@ -74,12 +96,42 @@ def get_column_definition(column_metadata):
...
+        help="URI base whose per-table subdirectories hold the Parquet data (e.g. "
+        "s3://bucket/prefix/scale-1000). The schema is derived from that data instead of a local "
+        "directory. Requires --table-names.",
+    )
+    parser.add_argument(
+        "--table-names",
+        nargs="+",
+        default=None,
+        help="Table names to generate schemas for. Required with --external-location-base since a "
+        "remote prefix cannot be listed.",
```
paul-aiyedun, line 118: I believe S3 supports listing objects with a specified prefix?

presto/scripts/register_external_tables.sh
```diff
@@ -0,0 +1,166 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+print_help() {
+  cat << EOF
+
+Usage: $0 [OPTIONS]
+
+Registers benchmark external tables in a Presto Hive schema whose data lives at an
```
paul-aiyedun, line 13: I think this script is effectively doing the same thing as `setup_benchmark_tables.sh`. Can we extend `setup_benchmark_tables.sh` with S3 options (e.g. `--s3`/`--data-location {file|s3}`, `--s3-data-dir-path`, etc.) instead of adding a new script?

presto/testing/common/test_utils.py
```diff
@@ -3,39 +3,50 @@
...
 def get_table_external_location(schema_name, table, presto_cursor):
     create_table_text = presto_cursor.execute(f"SHOW CREATE TABLE hive.{schema_name}.{table}").fetchone()
-    test_pattern = r"external_location = 'file:/var/lib/presto/data/hive/data/integration_test/(.*)'"
-    user_pattern = r"external_location = 'file:/var/lib/presto/data/hive/data/user_data/(.*)'"
     assert len(create_table_text) == 1
-    test_match = re.search(test_pattern, create_table_text[0])
-    external_dir = ""
+    # Capture everything between two single quotes
+    location_match = re.search(r"external_location = '([^']*)'", create_table_text[0])
+    location = location_match.group(1) if location_match else ""
```
paul-aiyedun, line 23: Should we assert that `location_match` and `location` exist?

benchmark_data_tools/duckdb_utils.py
```diff
@@ -1,10 +1,55 @@
...
+def read_scale_factor(metadata_uri: str):
+    """Read the scale_factor field from a metadata.json at ``metadata_uri``."""
+    # For local data
+    if not str(metadata_uri).startswith("s3://"):
+        with open(metadata_uri) as file:
+            return _extract_scale_factor(json.load(file))
+    # For remote data
+    ensure_remote_access(metadata_uri)
+    raw = duckdb.sql(f"SELECT content FROM read_text('{metadata_uri}')").fetchone()[0]
+    return _extract_scale_factor(json.loads(raw))
```
paul-aiyedun, line 51: I don't think scale factor access/reading is DuckDB specific logic?

benchmark_data_tools/duckdb_utils.py
```diff
@@ -1,10 +1,55 @@
...
+import json
+import os
 import re
 
 import duckdb
 
+_s3_configured = False
+
+
+def ensure_remote_access(path) -> None:
```
paul-aiyedun, line 13: Consider something like `configure_data_access`, since this also applies to local files.

paul-aiyedun (commented):
Changes overall look good to me. However, I had comments about extending existing logic and using more specific options names.

presto/docker/docker-compose.common.yml
```diff
@@ -22,9 +22,14 @@ services:
       - 8080:8080
     environment:
       - SERVER_START_TIMESTAMP=${SERVER_START_TIMESTAMP:-}
+      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-}
```
mattgara, line 26: Should the AWS variables live on `presto-base-java`? I think moving them there is preferable so the coordinator and Java worker inherit the same settings.

AFAICT, they currently reach the coordinator and native worker, but `presto-java-worker` only gets `SERVER_START_TIMESTAMP`. S3 reads happen on the worker, so Java queries would not have the credentials or region they need.

presto/testing/common/test_utils.py
```diff
@@ -3,39 +3,50 @@
...
+    # For remote path
+    if location and not location.startswith("file:"):
```
mattgara, line 25: Consider either handling arbitrary `file:` locations here or limiting the new registration option and README to S3.

It looks like any `file:` URI outside the integration-test and user-data mounts falls through to `external_dir = ""` and raises. The documented `file:` use case therefore does not work for a general local path.

presto/scripts/register_external_tables.sh
```diff
@@ -0,0 +1,166 @@
...
+  --table-names "${TABLE_NAMES[@]}"
+
+# HOSTNAME/PORT are read by create_hive_tables.py to connect to the coordinator.
+HOSTNAME="$HOST_NAME" PORT="$PORT" "${SCRIPT_DIR}/../../scripts/run_py_script.sh" \
+  -p "$CREATE_TABLES_SCRIPT_PATH" \
+  -r "$REQUIREMENTS_PATH" \
+  -- \
+  --schema-name "$SCHEMA_NAME" \
+  --schemas-dir-path "$TEMP_SCHEMA_DIR" \
+  --external-location-base "$EXTERNAL_LOCATION_BASE"
```
mattgara, line 164: Please require exactly one of `--data-dir-name` and `--external-location-base` before connecting to Presto.

As written, omitting both seems to drop and recreate the schema with paths such as `.../user_data//t`, while supplying both silently ignores `--data-dir-name`. `argparse` can reject both cases before any schema changes.

mattgara (commented):
LGTM overall, just had some cleanup comments.

PR #377, opened 2026-07-17
Fix data generation script always sets Parquet v1

paul-aiyedun:


benchmark_data_tools/tests/parquet_file_metadata_test.py
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
+# SPDX-License-Identifier: Apache-2.0
+
+import pyarrow.parquet as pq
+from generate_data_files import generate_data_files
+
+from .common_fixtures import get_all_parquet_relative_file_paths
+
+
+def test_generated_files_use_v2_page_format(setup_and_teardown):
```
misiugodfrey, line 10: Nit:  This name and the error text specify v2, but the check is v2 or newer.

misiugodfrey (approved):

benchmark_data_tools/tests/parquet_file_metadata_test.py
paul-aiyedun: Updated the assertion to match the test name.

PR #382, opened 2026-07-22
Add nightly Presto benchmark script for NVL72 Slurm cluster

misiugodfrey:
## Summary

- Adds `presto/slurm/presto-nvl72/nightly-benchmark.sh`, a daemon script for running automated nightly Presto benchmarks on the NVL72 Slurm cluster
- Pulls the latest nightly-pinned coordinator and `gpu-cuda13.1` worker images from GHCR into a temporary directory (leaves the cluster's main `IMAGE_DIR` untouched), then removes them after both runs complete
- Runs two benchmark configurations each night:
  - 1-node / 4-GPU / TPC-H sf1000
  - 2-node / 8-GPU (4 per node) / TPC-H sf3000
- Posts results to the benchmarking DB via `post_results.py` after each run
- Designed to run in a persistent `tmux`/`screen` session; sleeps until 3 AM EST (08:00 UTC) each night using `date` arithmetic

## Test plan

- [x] Verify syntax: `bash -n nightly-benchmark.sh`
- [x] Test immediate trigger: `RUN_HOUR_UTC=$(date -u +%H) ./nightly-benchmark.sh` — should fire within seconds
- [x] Confirm `.sqsh` files appear in `${IMAGE_DIR}/nightly-temp/` during a run and are removed on completion
- [x] Confirm `post_results.py` is invoked with correct `--benchmark-name`, `--cache-state warm`, and storage config per run
- [ ] Kill one benchmark mid-run; verify cleanup still executes and the loop continues to the next night


presto/scripts/run_benchmark.sh
```diff
@@ -319,7 +319,8 @@ fi
 echo "Using PRESTO_IMAGE_TAG: $PRESTO_IMAGE_TAG"
 
 BENCHMARK_TEST_DIR=${TEST_DIR}/performance_benchmarks
-pytest -q -s ${BENCHMARK_TEST_DIR}/${BENCHMARK_TYPE}_test.py ${PYTEST_ARGS[*]}
+pytest_exit=0
+pytest -q -s ${BENCHMARK_TEST_DIR}/${BENCHMARK_TYPE}_test.py ${PYTEST_ARGS[*]} || pytest_exit=$?
```
misiugodfrey: Needed to avoid an early exit here on failure since that prevents the logs from being moved.  This is to enable the results to be pushed even if there was a pytest failure when running the benchmarks.

presto/slurm/presto-nvl72/launch-run.sh
```diff
@@ -212,12 +212,13 @@ echo ""
 echo "Output files:"
 ls -lh "${OUT_FILE}" "${ERR_FILE}" 2>/dev/null || echo "No output files found"
 show_job_output "${OUT_FILE}" "${ERR_FILE}" "logs/cli.log" "benchmark results"
-[[ "${JOB_STATE}" == "COMPLETED" ]] || exit 1
 
-if [[ -n "${OUTPUT_PATH}" ]]; then
+if [[ -n "${OUTPUT_PATH}" && -f "result_dir/benchmark_result.json" ]]; then
```
misiugodfrey: Only copy the results over if there are results to copy - this is instead of checking if the job state was "COMPLETED" (which means without failures).

misiugodfrey:
This is designed to run with PR #338 so it will need re-testing once that lands.

presto/slurm/presto-nvl72/nightly-benchmark.sh
```diff
@@ -0,0 +1,205 @@
...
+    echo "  Iterations:        ${ITERATIONS}"
+
+    pull_images
+
+    echo ""
+    echo "--- Run 1/2: 1-node, 4-GPU, tpch-sf1000 ---"
+    run_benchmark 1 4 1000 "${output_1k}" || true
+    if [[ -d "${output_1k}" ]]; then
+        echo "Posting results for tpch-1k..."
+        post_results "${output_1k}" "tpch-rs-1000" "${STORAGE_CONFIG_1K}"
```
patdevinwilson: `run_benchmark ... || true` is good, but `post_results` is **not** guarded. With `set -euo pipefail`, a posting failure (API down, bad SKU, missing result fields) will abort `run_nightly` before the sf3000 run ever starts.

Suggest:
```bash
post_results "${output_1k}" "tpch-rs-1000" "${STORAGE_CONFIG_1K}" \
  || echo_warning "Failed to post tpch-1k results"
```
(and the same for the 3k path).

presto/slurm/presto-nvl72/nightly-benchmark.sh
```diff
@@ -0,0 +1,205 @@
...
+    echo_success "=== Nightly benchmark complete: ${date_tag} ==="
+}
+
+# ------------------------------------------------------------------------------
+# Main loop
+# ------------------------------------------------------------------------------
+while true; do
+    [[ "${RUN_NOW}" -eq 1 ]] || sleep_until_next_run
+    RUN_NOW=0
+    (run_nightly) || echo_warning "Nightly run FAILED — will retry tomorrow"
```
patdevinwilson: `cleanup_images` runs after the subshell returns, which covers soft failures — nice. It will **not** run on SIGINT/SIGTERM/kill of the daemon (unchecked item in your test plan). Consider:
```bash
trap 'cleanup_images || true; exit 130' INT TERM
```
before the loop so a manual stop still removes the temp `.sqsh` files.

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -573,6 +573,7 @@ function run_queries {
     export PRESTO_DATA_DIR=/var/lib/presto/data/hive/data/user_data; \
     export MINIFORGE_HOME=/workspace/miniforge3; \
     export HOME=/workspace; \
+    export LOGS_DIR=/var/log/nsys; \
```
patdevinwilson: This conflicts with the direction in #338, which mounts `${LOGS}:/opt/presto-server/logs` and exports `LOGS_DIR=/opt/presto-server/logs` so provenance + server logs are visible to pytest.

`/var/log/nsys` is the nsys token/report mount, not the general logs dir used by `launch_presto_servers.sh` / `run_context.py`. If this merges before/without #338's `/opt/presto-server/logs` mount on the CLI container, log snapshotting / GPU detection / provenance may miss files depending on which `run_coord_image` path is used.

Prefer aligning with #338 (`LOGS_DIR=/opt/presto-server/logs` + the matching mount), or land #338 first and rebase this.

presto/scripts/run_benchmark.sh
patdevinwilson: Preserving `pytest_exit` so logs/validation still run on failure is the right call for nightlies that want to post partial results. Please keep this if/when rebasing onto #373 (that PR currently leaves bare `pytest` under `set -e`).

presto/slurm/presto-nvl72/nightly-benchmark.sh
```diff
@@ -0,0 +1,205 @@
...
+        --output "${COORD_SQSH}" \
+        --overwrite
+    echo "Pulling worker image (${WORKER_TAG})..."
+    "${SCRIPT_DIR}/pull_ghcr_image.sh" \
+        "${REGISTRY}:${WORKER_TAG}" \
+        --output "${WORKER_SQSH}" \
+        --overwrite
+}
+
+run_benchmark() {
```
patdevinwilson, line 128: Hardcoding `presto-latest-gpu-cuda13.1` is fine if that's the cluster pin, but it'll silently go stale when CUDA bumps. The `WORKER_TAG` / `COORD_TAG` overrides in the header help — maybe also log a reminder in `run_nightly` when using the default tags, or read the default from `cluster_config.env` / `defaults.env` if those already know the CUDA version.

presto/slurm/presto-nvl72/nightly-benchmark.sh
```diff
@@ -0,0 +1,205 @@
...
+        -c "${COORD_SQSH}" \
+        -w "${WORKER_SQSH}" \
+        -o "${output_dir}" \
+        -i "${ITERATIONS}"
+}
+
+post_results() {
+    local output_dir="$1" benchmark_name="$2" storage_config="$3"
+    "${VT_ROOT}/scripts/run_py_script.sh" \
+        -p "${VT_ROOT}/benchmark_reporting_tools/post_results.py" \
```
patdevinwilson, line 145: Nit: this prints an error per missing var but doesn't `exit 1`, so the daemon can enter the sleep loop and fail later inside `post_results`. Prefer `echo_error` that exits, or accumulate and exit after the loop.

patdevinwilson (changes requested):
Useful daemon for NVL72 nightlies. A couple of correctness issues around failure handling and `LOGS_DIR` (esp. relative to #338) before this should land. Details inline.

presto/slurm/presto-nvl72/functions.sh
misiugodfrey: I think we should wait for #338 to land and then I'll rebase this.

presto/slurm/presto-nvl72/nightly-benchmark.sh
misiugodfrey: I've added logging to specify which tag is being used, which would highlight a failure if cuda bumps, but otherwise I think we just change it when that happens as it's going to need an adjustment anyways.

presto/slurm/presto-nvl72/nightly-benchmark.sh
misiugodfrey: I may need some clarification as to what you mean here.  Which validation are you saying doesn't exit?  `post_results.py`?

patdevinwilson: Sorry for the confusion — I was wrong here, not \`post_results.py\`. I meant the env var validation loop right above \`post_results()\`:

\`\`\`bash
for req in SKU_NAME STORAGE_CONFIG_1K STORAGE_CONFIG_3K; do
    [[ -n "${!req:-}" ]] || echo_error "Required env var ${req} is not set"
done
\`\`\`

I'd assumed \`echo_error\` only printed (like \`echo_warning\`), but I just checked \`echo_helpers.sh\` and it does call \`exit 1\`. So this is correct as-is and my comment doesn't apply — no change needed here. Apologies for the noise.

presto/slurm/presto-nvl72/nightly-benchmark.sh
patdevinwilson: Fixed in 555869a — looks correct now: a posting failure for the sf1000 run warns and falls through to the sf3000 run instead of aborting the whole night.

presto/slurm/presto-nvl72/nightly-benchmark.sh
patdevinwilson: Fixed in 555869a, and placed correctly (before the main loop, so it's live during \`sleep_until_next_run\` too, not just mid-benchmark).

One thing worth confirming in testing (matches the still-unchecked "kill mid-run" item in the test plan): \`run_nightly\` executes inside a subshell \`( ... )\` at the bottom. If a SIGINT/SIGTERM arrives while that subshell is blocked deep inside \`launch-run.sh\` (e.g. waiting on \`srun\`/\`sbatch\`), double check the signal actually propagates up to the parent shell's trap promptly rather than only being handled once the subshell's current foreground command returns on its own. If it does propagate immediately (likely, since tmux/screen send the signal to the whole process group), this is exactly right as-is.

presto/slurm/presto-nvl72/nightly-benchmark.sh
patdevinwilson: Sounds good, the startup logging is a reasonable middle ground for now.

patdevinwilson:
Went through 555869a — replied inline, but summarizing:

- **post_results guard** and **SIGINT/TERM trap**: both fixed and look correct. Left one non-blocking testing note on the trap re: subshell signal propagation, matches your still-unchecked "kill mid-run" test plan item.
- **`echo_error` nit**: that one was my mistake — it does `exit 1`, no change needed. Sorry for the noise.
- **Worker tag hardcoding**: the startup logging is a reasonable call, agreed.
- **`functions.sh` LOGS_DIR conflict with #338**: this is the one remaining item, and per your comment it's intentionally deferred until #338 lands and you rebase. Makes sense to me — I'll take another pass once that's in. Everything else here looks good to merge as-is if you'd rather land this first and handle the rebase separately.

patdevinwilson (approved):

presto/slurm/presto-nvl72/nightly-benchmark.sh
```diff
@@ -0,0 +1,210 @@
...
+}
+
+# ------------------------------------------------------------------------------
+# Main loop
+# ------------------------------------------------------------------------------
+trap 'cleanup_images || true; exit 130' INT TERM
+while true; do
+    [[ "${RUN_NOW}" -eq 1 ]] || sleep_until_next_run
+    RUN_NOW=0
+    (run_nightly) || echo_warning "Nightly run FAILED — will retry tomorrow"
```
mattgara: Please run the subshell outside an `&&`/`||` conditional and capture its status explicitly with `set -e` active inside, or otherwise propagate unguarded failures from `run_nightly`. Otherwise we risk missing errors that the `set -e` would otherwise catch.

presto/slurm/presto-nvl72/launch-run.sh
```diff
@@ -238,12 +238,13 @@ echo ""
...
```
mattgara: `launch-run.sh` receives the per-run destination through `-o`/`--output-path`, but only copies to it when `benchmark_result.json` exists. Please consider copying any available artifacts to that directory after a failed job, then make this nightly caller test for the JSON file before posting so early-failure logs are preserved.

presto/slurm/presto-nvl72/nightly-benchmark.sh
```diff
@@ -0,0 +1,210 @@
...
+        "${output_dir}" \
+        --sku-name "${SKU_NAME}" \
+        --storage-configuration-name "${storage_config}" \
+        --cache-state "warm" \
+        --benchmark-name "${benchmark_name}" \
+        --label "nightly"
+}
+
+cleanup_images() {
+    rm -f "${COORD_SQSH}" "${WORKER_SQSH}"
```
mattgara: I think the cleanup should run through the same compute-side path used to create these files. `pull_ghcr_image.sh` explicitly runs its file checks and import under `srun` because `IMAGE_DIR` may not be mounted on the head node. In that configuration this head-node `rm -f` reports success against a missing path, leaving both large temporary images behind after every nightly run.

mattgara (commented):
LGTM, there are three items that may need another look.

presto/slurm/presto-nvl72/nightly-benchmark.sh
```diff
@@ -0,0 +1,222 @@
...
+
+# ------------------------------------------------------------------------------
+# Main loop
+# ------------------------------------------------------------------------------
+trap 'cleanup_images || true; exit 130' INT TERM
+while true; do
+    [[ "${RUN_NOW}" -eq 1 ]] || sleep_until_next_run
+    RUN_NOW=0
+    nightly_exit=0
+    (run_nightly) || nightly_exit=$?
```
mattgara: Can we please move `(run_nightly)` out of the OR-list and capture its status with `errexit` (`-e`) disabled only in the parent. I think the new assignment changes the right-hand side, but the subshell is still the left-hand side of `||`, so bash suppresses `errexit` throughout `run_nightly` (I assume this is not the intended case.)

presto/slurm/presto-nvl72/launch-run.sh
```diff
@@ -238,12 +238,13 @@ echo ""
 echo "Output files:"
 ls -lh "${OUT_FILE}" "${ERR_FILE}" 2>/dev/null || echo "No output files found"
 show_job_output "${OUT_FILE}" "${ERR_FILE}" "logs/cli.log" "benchmark results"
-[[ "${JOB_STATE}" == "COMPLETED" ]] || exit 1
 
-if [[ -n "${OUTPUT_PATH}" ]]; then
+if [[ -n "${OUTPUT_PATH}" && -d "result_dir" ]]; then
```
mattgara: I believe this still loses logs from failures before `collect_results`. Consider copying the available job and server logs into `${OUTPUT_PATH}` before returning the failed job status, or collecting them from an EXIT trap. AFAICT, `result_dir` is created unconditionally at startup, but populated from `logs/` only after `run_queries` succeeds.

presto/slurm/presto-nvl72/nightly-benchmark.sh
misiugodfrey: I think I better understand the ask here.  I've isolated this idiom and put it in it's own function `run_isolated()`.

presto/slurm/presto-nvl72/launch-run.sh
misiugodfrey: I think the overall logging for the slurm scripts needs a bit of an update.  For now I've moved the log collection into the same trap that handles metadata injection - so the logs should always get copied over if they exist.

mattgara (approved):
LGTM

misiugodfrey:
Holding out on merging until I can re-test the final version on the cluster.

misiugodfrey:
Testing has shown this to be in working order aside from some recent issues in the velox images described here: https://github.com/rapidsai/velox-testing/pull/427

I'll go ahead with the merge and then update as part of the linked issue.

PR #383, opened 2026-07-22
fix(workflow-status): prioritize failing symbol in related-PR search

Avinash-Raj:
## What
Make the nightly status-report's `"Related issues/PRs"` search surface the PR that actually fixes a failure instead of loosely-related noise.

## Why

In the Jul 22 nightly report, the TPCH Q22 GPU benchmark failed with `Unsupported expression: substr`, but the analyser tagged the tangential facebookincubator/velox#18154 and missed the real fix, facebookincubator/velox#18168.

Root cause: the search only used C++/source-file names, so it queried `AstExpressionUtils.h` (the file the error was *thrown from*) instead of `substr` (the symbol that actually failed) — and file-name searches return nothing useful.

## Changes

- Add `extract_error_phrases()`: generic extraction of the failing **symbol** +  **reason** from C++/Velox semantic errors (`Unsupported…`, `Unknown function…`, `… not registered`, `Cannot resolve…`, `'X' not supported`), ignoring enum codes like `INVALID_STATE`.

- Make `find_related_github_items()` two-tiered: search the semantic terms first and, if they hit, return only those — skipping the broad file/identifier search that fills the block with wrong URLs.

## Local Test 

```bash
export ANTHROPIC_API_KEY=...        # your key
export CLAUDE_MODEL=claude-opus-4-8 # optional; defaults to "opus"
python3 scripts/workflow_status/workflow_status.py \
  --run-id 29895314576 \
  --repo rapidsai/velox-testing
```

### DIFF
--- a/scripts/workflow_status/lib/upstream.py
+++ b/scripts/workflow_status/lib/upstream.py
@@ -24,6 +24,41 @@ _ERROR_INDICATOR = re.compile(r"error|undefined|unresolved|FAILED|fatal", re.IGN
 _UNDEFINED_REF = re.compile(r"undefined reference to\s+[`']([^`']+)[`']")
 _ARCHIVE_FILE = re.compile(r"([\w.-]{3,})\.(?:tar\.gz|tgz|tar\.bz2|tar\.xz|zip)\b")
 _FAILED_TARGET = re.compile(r"^FAILED:\s+(\S+)", re.MULTILINE)
+# Semantic runtime-error signals: the *symbol* that failed and the human reason.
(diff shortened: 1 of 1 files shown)


simoneves (approved):
LGTM if it works.

Do you have any before and after example CI runs?

mattgara:
> Do you have any before and after example CI runs?

Also would be interested in seeing this.

Avinash-Raj:
[Done](https://github.com/rapidsai/velox-testing/actions/runs/30067955732), pls check the nightly status slack channel

PR #387, opened 2026-07-27
Fall back to local compilation when sccache has no workers

kjmph:
The distributed scheduler can remain reachable while reporting zero compilation servers. The setup check treated any SchedulerStatus as available, so cache misses waited on 7,140-second requests and retries until GitHub Actions canceled the ARM jobs at six hours.

The existing local fallback was also exported from a child shell, so it could not reach compiler requests in the subsequent Docker build.

Require a nonempty server list, reuse one status snapshot, and source the setup script from both Presto and Velox image builds. This preserves S3 caching while bypassing distributed compilation whenever the scheduler is unavailable.

Add regression coverage for healthy, empty, missing, malformed, failed, explicitly disabled, and fallback-disabled scheduler states, and run it from PR checks.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -18,6 +18,7 @@ jobs:
       - devcontainer
       - detect-changes
       - test_benchmark_data_tools
+      - test_sccache_setup
     permissions:
       contents: read
     uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@main  # zizmor: ignore[unpinned-uses]
@@ -65,6 +66,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       benchmark_data_tools: ${{ steps.filter.outputs.benchmark_data_tools }}
+      sccache: ${{ steps.filter.outputs.sccache }}
     steps:
       - name: Checkout velox-testing
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
@@ -78,6 +80,11 @@ jobs:
           filters: |
             benchmark_data_tools:
               - "benchmark_data_tools/**"
+            sccache:
+              - "scripts/sccache/sccache_setup.sh"
+              - "scripts/sccache/sccache_setup_test.sh"
+              - "presto/docker/native_build.dockerfile"
+              - "velox/docker/adapters_build.dockerfile"
 
   test_benchmark_data_tools:
     needs: detect-changes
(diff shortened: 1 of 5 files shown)


scripts/sccache/sccache_setup_test.sh
```diff
@@ -0,0 +1,139 @@
...
+    actual_status_calls="$(grep -c '^--dist-status|' "${trace}" || true)"
+    [[ "${actual_status_calls}" == "${expected_status_calls}" ]] \
+        || fail "${name}: expected ${expected_status_calls} status calls, got ${actual_status_calls}"
+
+    if [[ "${expected_exit_code}" == "0" ]]; then
+        assert_contains "${output}" "final_no_dist=${expected_no_dist}"
+    fi
+}
+
+healthy_status='{"SchedulerStatus":["https://arm64.example",{"servers":[{"id":"worker-1"}]}]}'
```
simoneves, line 112: What about `x86_64`?

presto/docker/native_build.dockerfile
```diff
@@ -89,7 +89,8 @@ if [ "$ENABLE_SCCACHE" = "ON" ]; then
   if [ -n "${SCCACHE_NO_DIST_COMPILE:-}" ]; then
     export SCCACHE_NO_DIST_COMPILE=1;
   fi
-  bash /sccache_setup.sh;
+  # Source the setup so its automatic local-fallback export reaches the build.
+  source /sccache_setup.sh;
```
simoneves, line 93: Please explain the previous failing here in more detail

.github/workflows/pr.yaml
```diff
@@ -65,6 +66,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       benchmark_data_tools: ${{ steps.filter.outputs.benchmark_data_tools }}
+      sccache: ${{ steps.filter.outputs.sccache }}
```
simoneves, line 69: What does the `sccache` tag actually do here? (and below)

simoneves (commented):
Looks plausible although I am by no means a power-user of this mechanism

mattgara (approved):
LGTM, thanks!

.github/workflows/pr.yaml
kjmph: This is a boolean flag that is set when any of the sccache related files change. By declaring it, it is exposed as an output of detect-changes, and test_sccache_setup checks the output later to decide whether to run. If we don't want an sccache related test on PRs, we can remove it.

presto/docker/native_build.dockerfile
kjmph: Previously a child shell was executed which set a fallback environment variable (`SCCACHE_NO_DIST_COMPILE=1`), which is then lost when the process exits. Subsequent build commands didn't see it. By sourcing, it is executed in the current shell, so later steps inherit the fallback.

scripts/sccache/sccache_setup_test.sh
kjmph: This is a test, the URL isn't used. Would you prefer a different name just to make the arch neutrality clear in the test case? x86_64 does follow the same zero-worker path, it isn't branched based on arch, if that alleviates your concern.

PR #390, opened 2026-07-27
feat(presto/bench): config buffer tuning and benchmark robustness fixes

misiugodfrey:
## Summary

Low-risk correctness fixes and config tuning for Presto native benchmarks. Good first merge of this series.

Based on kjmph's bench33 POC patch.

- **run_benchmark.sh**: add `--maxfail=1` when `--profile` is set (avoids waiting for all queries when one fails during profiling runs); add `PRESTO_BENCHMARK_KEEP_VENV` escape hatch
- **common_functions.sh**: replace `jq`-based `wait_for_worker_node_registration` with `python3` JSON parse; fix broken while-loop that was not actually retrying correctly
- **common_fixtures.py**: fix error handling for non-Presto exceptions in `failed_queries_dict`; move REST metrics collection outside the profiler interval so profiling overhead doesn't skew results; add `PRESTO_SESSION_PROPERTIES` env var for per-run session overrides
- **hive.properties**: comment out `SOFT_AFFINITY` in base template; keep it in the CPU-specific override (`overrides/cpu/`) only — GPU experiments should opt in explicitly
- **Config overrides**: `exchange.max-response-size=64MB` for CPU coordinator; lower CPU worker buffer defaults to 32MB; add `local-exchange.max-buffer-size` and `largest-size-class-pages`; tune GPU exchange/sink buffer sizes
- **.gitignore**: add `result_dir_*/` and `latest_result_dir.txt` (introduced by PR #3 of this series)

## Dependencies

None. This PR is self-contained and should merge first.

> **Note:** `common_functions.sh`, `run_benchmark.sh`, and `common_fixtures.py` overlap with the open `dmakkar/qol-fixes` branch. The `common_fixtures.py` changes touch the same fixture function — whichever merges second will need a manual merge. Coordinate before landing.

## Verification

- [ ] Run a GPU TPC-H benchmark end-to-end on the NVL72 cluster
- [x] Run a CPU TPC-H benchmark with `--profile`; verify benchmark aborts after the first query failure instead of continuing through all 22 queries
- [x] Run `presto/scripts/run_benchmark.sh` and confirm the `PRESTO_BENCHMARK_KEEP_VENV=1` escape hatch skips venv teardown
- [x] Verify `wait_for_worker_node_registration` retries correctly when workers are slow to register (previously the loop exited immediately on the first iteration)
- [ ] Confirm `common_fixtures.py` metrics are collected after the profiler window closes (check that `QueryStats` REST calls don't appear inside the nsys/perf capture interval in logs)
- [x] Run `./ci/check_style.sh` (ruff + codespell) on changed Python files

### DIFF
--- a/presto/docker/config/template/etc_coordinator/catalog/hive.properties
+++ b/presto/docker/config/template/etc_coordinator/catalog/hive.properties
@@ -36,5 +36,6 @@ hive.optimize-parsing-of-partition-values-enabled=true
 hive.optimize-parsing-of-partition-values-threshold=200
 hive.split-loader-concurrency=16
 
-# Affinity scheduling
-hive.node-selection-strategy=SOFT_AFFINITY
+# Affinity scheduling is variant-specific. CPU enables soft affinity in its
+# override; GPU can opt in explicitly when placement experiments require it.
+# hive.node-selection-strategy=SOFT_AFFINITY
--- a/presto/docker/config/template/overrides/cpu/etc_worker/config_native.properties
+++ b/presto/docker/config/template/overrides/cpu/etc_worker/config_native.properties
@@ -11,7 +11,7 @@ exchange.max-response-size=64MB
 
 # CPU UCX row pages and local exchange were tuned together. The generator
 # reapplies these as overridable CPU_* values on every run.
-local-exchange.max-buffer-size=512MB
+local-exchange.max-buffer-size=536870912
 largest-size-class-pages=256
 
 # Increase concurrent exchange fetch threads and connection limits for heavy multi-node shuffles.
--- a/presto/docker/config/template/overrides/cpu/etc_worker/config_native.properties
+++ b/presto/docker/config/template/overrides/cpu/etc_worker/config_native.properties
@@ -2,10 +2,17 @@
 cudf.enabled=false
 
 exchange.http-client.enable-connection-pool=true
-exchange.max-buffer-size=1GB
-sink.max-buffer-size=1GB
+# Single-worker fallbacks. generate_presto_config.sh raises exchange/sink to
+# 512MB whenever the full cluster has more than one worker.
+exchange.max-buffer-size=32MB
+sink.max-buffer-size=32MB
 # Increase max bytes returned per exchange HTTP response to reduce round trips for large shuffles.
 exchange.max-response-size=64MB
+
+# CPU UCX row pages and local exchange were tuned together. The generator
+# reapplies these as overridable CPU_* values on every run.
+local-exchange.max-buffer-size=512MB
+largest-size-class-pages=256
 
 # Increase concurrent exchange fetch threads and connection limits for heavy multi-node shuffles.
 exchange.client-threads=40
--- a/presto/docker/config/template/overrides/cpu/etc_worker/config_native.properties
+++ b/presto/docker/config/template/overrides/cpu/etc_worker/config_native.properties
@@ -2,15 +2,9 @@
 cudf.enabled=false
 
 exchange.http-client.enable-connection-pool=true
-# Single-worker fallbacks. generate_presto_config.sh raises exchange/sink to
-# 512MB whenever the full cluster has more than one worker.
-exchange.max-buffer-size=32MB
-sink.max-buffer-size=32MB
-# Increase max bytes returned per exchange HTTP response to reduce round trips for large shuffles.
-exchange.max-response-size=64MB
-
-# CPU UCX row pages and local exchange were tuned together. The generator
-# reapplies these as overridable CPU_* values on every run.
+exchange.max-buffer-size=512MB
+sink.max-buffer-size=512MB
+# CPU UCX row pages and local exchange buffer tuned for heavy multi-node shuffles.
 local-exchange.max-buffer-size=536870912
 largest-size-class-pages=256
 
--- a/presto/docker/config/template/overrides/gpu/etc_coordinator/config_native.properties
+++ b/presto/docker/config/template/overrides/gpu/etc_coordinator/config_native.properties
@@ -3,3 +3,9 @@ optimizer.joins-not-null-inference-strategy=USE_FUNCTION_METADATA
 # Use a default filter factor to estimate the selectivity of filters in queries.
 optimizer.default-filter-factor-enabled=true
 cluster-tag=native-gpu
+
+# Match GPU workers' larger exchange/result buffers for final result fetch and
+# large distributed shuffles.
+exchange.max-buffer-size=512MB
+sink.max-buffer-size=512MB
+exchange.max-response-size=64MB
--- a/presto/docker/config/template/overrides/gpu/etc_coordinator/config_native.properties
+++ b/presto/docker/config/template/overrides/gpu/etc_coordinator/config_native.properties
@@ -8,4 +8,8 @@ cluster-tag=native-gpu
 # large distributed shuffles.
 exchange.max-buffer-size=512MB
 sink.max-buffer-size=512MB
+# Raise the HTTP client content-length cap above max-response-size so the
+# coordinator does not reject responses it requested. Default is 32MB which
+# is below the 64MB response size (48MB after 75% encoding allowance).
+exchange.http-client.max-content-length=128MB
 exchange.max-response-size=64MB
--- a/presto/docker/config/template/overrides/gpu/etc_worker/config_native.properties
+++ b/presto/docker/config/template/overrides/gpu/etc_worker/config_native.properties
@@ -11,7 +11,6 @@ cudf.memory_resource=async
 local-exchange.max-buffer-size=536870912
 exchange.max-buffer-size=512MB
 sink.max-buffer-size=512MB
-exchange.max-response-size=64MB
 
 # Disable JIT because it takes a few iterations to warm up jit cache in all workers
 cudf.jit_expression_enabled=false
--- a/presto/docker/config/template/overrides/gpu/etc_worker/config_native.properties
+++ b/presto/docker/config/template/overrides/gpu/etc_worker/config_native.properties
@@ -6,6 +6,13 @@ cudf.exchange=false
 cudf.exchange.server.port=0000
 cudf.memory_resource=async
 
+# Keep large distributed result and exchange paths from stalling on tiny default
+# buffers. Q15/Q11-style result delivery is sensitive to these at SF10K.
+local-exchange.max-buffer-size=512MB
+exchange.max-buffer-size=512MB
+sink.max-buffer-size=512MB
+exchange.max-response-size=64MB
+
 # Disable JIT because it takes a few iterations to warm up jit cache in all workers
 cudf.jit_expression_enabled=false
 # Turn on to use intra-node exchange optimization.
--- a/presto/docker/config/template/overrides/gpu/etc_worker/config_native.properties
+++ b/presto/docker/config/template/overrides/gpu/etc_worker/config_native.properties
@@ -8,7 +8,7 @@ cudf.memory_resource=async
 
 # Keep large distributed result and exchange paths from stalling on tiny default
 # buffers. Q15/Q11-style result delivery is sensitive to these at SF10K.
-local-exchange.max-buffer-size=512MB
+local-exchange.max-buffer-size=536870912
 exchange.max-buffer-size=512MB
 sink.max-buffer-size=512MB
 exchange.max-response-size=64MB
--- a/presto/scripts/common_functions.sh
+++ b/presto/scripts/common_functions.sh
@@ -31,8 +31,16 @@ function wait_for_worker_node_registration() {
   echo "Coordinator URL: $COORDINATOR_URL"
   local -r MAX_RETRIES=12
   local retry_count=0
-  until curl -s -f -o node_response.json ${COORDINATOR_URL}/v1/node && \
-        (( $(jq length node_response.json) > 0 )); do
+  local node_count=0
+  while true; do
+    if curl -s -f -o node_response.json "${COORDINATOR_URL}/v1/node"; then
+      node_count=$(python3 -c \
+        'import json, sys; print(len(json.load(open(sys.argv[1]))))' \
+        node_response.json 2>/dev/null || echo 0)
+      if [[ "${node_count}" =~ ^[0-9]+$ ]] && (( node_count > 0 )); then
+        break
+      fi
+    fi
     if (( $retry_count >= $MAX_RETRIES )); then
       echo "Error: Worker node not registered after 60s. Exiting."
       exit 1
--- a/presto/slurm/presto-nvl72/functions.sh
+++ b/presto/slurm/presto-nvl72/functions.sh
@@ -236,6 +236,12 @@ function run_worker {
     mkdir -p ${worker_data}/hive/data/user_data
     mkdir -p ${VT_ROOT}/.hive_metastore
 
+    # Provide a writable directory for libcudf's JIT/RTC cache. The container
+    # root filesystem is read-only squashfs, so /root/.libcudf is not writable
+    # without this mount.
+    local libcudf_cache="${SCRIPT_DIR}/libcudf_cache_${worker_id}"
+    mkdir -p "${libcudf_cache}"
+
     local vt_worker_env_file="/var/worker_env_file"
     local vt_cufile_log_dir="/var/log/cufile"
     local vt_cufile_log="${vt_cufile_log_dir}/cufile_worker_${worker_id}.log"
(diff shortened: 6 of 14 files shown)


presto/docker/config/template/overrides/cpu/etc_worker/config_native.properties
```diff
@@ -2,11 +2,18 @@
...
+# Single-worker fallbacks. generate_presto_config.sh raises exchange/sink to
+# 512MB whenever the full cluster has more than one worker.
+exchange.max-buffer-size=32MB
+sink.max-buffer-size=32MB
 # Increase max bytes returned per exchange HTTP response to reduce round trips for large shuffles.
 exchange.max-response-size=64MB
 
+# CPU UCX row pages and local exchange were tuned together. The generator
+# reapplies these as overridable CPU_* values on every run.
+local-exchange.max-buffer-size=536870912
```
misiugodfrey, line 8: Note that these values can't have units associated with them (MB for instance), so we need to leave this as a byte value.

presto/scripts/common_functions.sh
```diff
@@ -31,8 +31,16 @@ function wait_for_worker_node_registration() {
   echo "Coordinator URL: $COORDINATOR_URL"
   local -r MAX_RETRIES=12
   local retry_count=0
-  until curl -s -f -o node_response.json ${COORDINATOR_URL}/v1/node && \
-        (( $(jq length node_response.json) > 0 )); do
+  local node_count=0
+  while true; do
+    if curl -s -f -o node_response.json "${COORDINATOR_URL}/v1/node"; then
+      node_count=$(python3 -c \
+        'import json, sys; print(len(json.load(open(sys.argv[1]))))' \
```
misiugodfrey, line 38: replaced `jq` with python parsing since `jq` is not installed in the cluster/coordinator image by default.

presto/docker/config/template/overrides/gpu/etc_coordinator/config_native.properties
```diff
@@ -3,3 +3,9 @@ optimizer.joins-not-null-inference-strategy=USE_FUNCTION_METADATA
 # Use a default filter factor to estimate the selectivity of filters in queries.
 optimizer.default-filter-factor-enabled=true
 cluster-tag=native-gpu
+
+# Match GPU workers' larger exchange/result buffers for final result fetch and
+# large distributed shuffles.
+exchange.max-buffer-size=512MB
+sink.max-buffer-size=512MB
+exchange.max-response-size=64MB
```
kjmph, line 15: Could we pair this `exchange.max-response-size` increase with the coordinator’s HTTP content limit? With `64MB`, Presto can request roughly 48 MB after its 75% encoding allowance, while `exchange.http-client.max-content-length` remains at its 32 MB default. I received errors on SF=30K when we went over this limit.

The source benchmark attempted to set the HTTP limit to 128 MB through `--config-overrides`, but that helper was replace-only for coordinator properties. Because this key was absent, the override was silently applied only to native workers and never reached the Java coordinator. Now that we have this in the coordinator, we can override in testing. Alternatively, we could also extend `--config-overrides` to allow new propertise.

I can test once this PR is merged, and if we still have problems, the conservative fix is to use `exchange.max-response-size=32MB`. Yet, I'm pretty sure we are narrowing down why the prior test hetting `exchange.http-client.max-content-length` didn't work.

presto/docker/config/template/overrides/gpu/etc_worker/config_native.properties
```diff
@@ -6,6 +6,13 @@ cudf.exchange=false
 cudf.exchange.server.port=0000
 cudf.memory_resource=async
 
+# Keep large distributed result and exchange paths from stalling on tiny default
+# buffers. Q15/Q11-style result delivery is sensitive to these at SF10K.
+local-exchange.max-buffer-size=536870912
+exchange.max-buffer-size=512MB
+sink.max-buffer-size=512MB
+exchange.max-response-size=64MB
```
kjmph: I think this is a coordinator only setting.. Can we remove this in the GPU and CPU worker properties, so it isn't confusing in the future?

presto/docker/config/template/overrides/cpu/etc_worker/config_native.properties
```diff
@@ -2,11 +2,18 @@
...
 
 exchange.http-client.enable-connection-pool=true
-exchange.max-buffer-size=1GB
-sink.max-buffer-size=1GB
+# Single-worker fallbacks. generate_presto_config.sh raises exchange/sink to
+# 512MB whenever the full cluster has more than one worker.
+exchange.max-buffer-size=32MB
+sink.max-buffer-size=32MB
 # Increase max bytes returned per exchange HTTP response to reduce round trips for large shuffles.
 exchange.max-response-size=64MB
```
kjmph: Similarly, remove this here, as this pre-existed the commit, and I think what confused me.

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -347,7 +353,8 @@ ${DATA}:/var/lib/presto/data/hive/data/user_data,\
 ${VT_ROOT}/.hive_metastore:/var/lib/presto/data/hive/metastore,\
 ${WORKER_ENV_FILE}:${vt_worker_env_file},\
 ${LOGS}:${vt_cufile_log_dir},\
-${LOGS}:${vt_nsys_report_dir}${driver_mounts}${gds_mounts:+,${gds_mounts}}${worker_extra_mounts} \
+${LOGS}:${vt_nsys_report_dir},\
+${libcudf_cache}:/root/.libcudf${driver_mounts}${gds_mounts:+,${gds_mounts}}${worker_extra_mounts} \
```
kjmph: @lamarrr, @bdice, could you weigh in on the intended libcudf cache location here? The relevant cuDF version defaults `LIBCUDF_KERNEL_CACHE_PATH` to `$HOME/.cudf`, yes? This mounts the writable directory at `/root/.libcudf`.

Unless the environment variable is set elsewhere to `/root/.libcudf`, this mount will not cover cuDF’s default cache path. @lamarrr mentioned that a few different locations are tested, so maybe I'm mistaken.

Yet, I think this mount target should be `$HOME/.cudf`.. or should the launcher explicitly set `LIBCUDF_KERNEL_CACHE_PATH=xx` as in PR #398 ?

presto/docker/config/template/overrides/cpu/etc_worker/config_native.properties
```diff
@@ -2,11 +2,18 @@
 cudf.enabled=false
 
 exchange.http-client.enable-connection-pool=true
-exchange.max-buffer-size=1GB
-sink.max-buffer-size=1GB
+# Single-worker fallbacks. generate_presto_config.sh raises exchange/sink to
+# 512MB whenever the full cluster has more than one worker.
+exchange.max-buffer-size=32MB
```
kjmph: Was the corresponding `generate_presto_config.sh` change intentionally omitted? The comment says these are single-worker fallbacks and that multi-worker clusters are raised to `512MB`, but the generator in this PR never modifies `exchange.max-buffer-size` or `sink.max-buffer-size`.

Consequently, every worker in a newly generated multi-worker CPU run will use `32MB` for both buffers. Before this PR, each worker used `1GB`. The source patch, `bench33.patch` intended `512MB` for multi-worker runs. The coordinator also does not receive the intended `512MB` values.

Perhaps we port the buffer reconciliation from the bench33.patch after the numbered worker configurations have been created? Use `32MB` for a single worker and `512MB` for multiple workers? We can apply both properties with replace-or-append to the `coordinator` and every `etc_worker_*` configuration.

If that generator locic is intentionally deferred, perhaps we avoid lowering the static worker values in this PR and remove the comments promising behavior that will be present later?

presto/testing/performance_benchmarks/common_fixtures.py
```diff
@@ -124,20 +150,25 @@ def benchmark_query_function(query_id):
...
-                            query_name=str(query_id),
-                            hostname=hostname,
-                            port=port,
-                            output_dir=bench_output_dir,
-                        )
+            # Keep post-query REST collection outside a per-iteration
+            # profiler interval. Otherwise CPU samples after query
+            # completion describe metrics collection and worker idling,
+            # rather than the engine work the profile is meant to measure.
+            if metrics:
```
kjmph: I think this hunk was ported without the per-iteration profiling structure it depended on in the source patch.

Moving metrics collection outside the iteration loop means only the final iteration's Presto query ID is collected. With the default five iterations, the previous code produced five query-specific
metrics files. Yet, this produces one. If a later iteration fails, metrics are not collected for any earlier successful iteration.

It also does not currently move metrics outside the profiler interval. Profiling starts before the loop at line 133 and stops in the outer finally at line 176, after this call.

I suggest we either port the complete per-iteration profiling structure from the source patch, bench33.patch, or retain the current combined mode while recording every iteration's Presto query ID, stopping and waiting for the profiler after the loop. Then collect metrics for every recorded ID. We should verify that collection begins only after profiling has stopped.

presto/scripts/common_functions.sh
kjmph: Could we complete the `jq` dependency removal in the Slurm launcher as part of this change? `presto/slurm/presto-nvl72/functions.sh` still stages a static `jq` binary at lines 550–570 and prepends its directory to `PATH` at line 577. Thus, a clean run still downloads `jq` from GitHub before launching the benchmark. An environment without access can still fail even though `jq` is no longer used.

Hence, can we remove the `jq` download block and remove `/workspace/.cache/bin` PATH prefix from `run_coord_image`? The source patch, `bench33.patch` removed both pieces.

presto/docker/config/template/etc_coordinator/catalog/hive.properties
```diff
@@ -36,5 +36,6 @@ hive.optimize-parsing-of-partition-values-enabled=true
 hive.optimize-parsing-of-partition-values-threshold=200
 hive.split-loader-concurrency=16
 
-# Affinity scheduling
-hive.node-selection-strategy=SOFT_AFFINITY
+# Affinity scheduling is variant-specific. CPU enables soft affinity in its
+# override; GPU can opt in explicitly when placement experiments require it.
+# hive.node-selection-strategy=SOFT_AFFINITY
```
kjmph, line 41: Upon review, could we preserve the existing Java benchmark behavior here? This shared coordinator template is also used by the Java variant. Commenting out SOFT_AFFINITY here and restoring it only in the CPU override changes Java from SOFT_AFFINITY to Presto's default NO_PREFERENCE.

The stated intent is to disable affinity for GPU while retaining it for CPU.. Yet, maybe we want to preserve the pre-PR behavior for Java but adding a Java coordinator override containing `hive.node-selection-strategy=SOFT_AFFINITY`?

If we are intentionally changing Java, perhaps it should be called out and validated explicitly..

presto/testing/performance_benchmarks/common_fixtures.py
```diff
@@ -66,7 +84,14 @@ def presto_cursor(request):
...
     user = request.config.getoption("--user")
     schema = request.config.getoption("--schema-name")
-    conn = prestodb.dbapi.connect(host=hostname, port=port, user=user, catalog="hive", schema=schema)
+    conn = prestodb.dbapi.connect(
+        host=hostname,
+        port=port,
+        user=user,
+        catalog="hive",
+        schema=schema,
+        session_properties=_session_properties_from_env(),
```
kjmph, line 176: Since this feature is intended for per-run optimizer and execution tuning, could we record the effective session properties in the benchmark results? They are currently passed to the Presto connection but are not included in `run_context` or the generated configuration snapshot.

Two runs using different `PRESTO_SESSION_PROPERTIES` can produce materially different plans and timings while appearing to have identical configurations in `benchmark_result.json`. `post_results.py` also filters context fields that are not declared in `BenchmarkMetadata`.

Perhaps we parse the properties once, and pass the resulting map to the connection while storing the same normalized map under a `session_properties` context field? We can add the corresponding field to `BenchmarkMetadata` so it survives result posting. Maybe the env var format should also be documented alongside the benchmark options?

presto/slurm/presto-nvl72/functions.sh
```diff
@@ -236,6 +236,12 @@ function run_worker {
     mkdir -p ${worker_data}/hive/data/user_data
     mkdir -p ${VT_ROOT}/.hive_metastore
 
+    # Provide a writable directory for libcudf's JIT/RTC cache. The container
+    # root filesystem is read-only squashfs, so /root/.libcudf is not writable
+    # without this mount.
+    local libcudf_cache="${SCRIPT_DIR}/libcudf_cache_${worker_id}"
```
kjmph, line 263: If these per-worker cache directories remain after resolving the cache-path question above, could we keep them out of the working tree status? Each Slurm run creates `libcudf_cache_<worker_id>` beneath the tracked `presto-nvl72` directory, but the current `.gitignore` covers `worker_data*` and not these new directories.

Shall we add `presto/slurm/presto-nvl72/libcudf_cache_*/` to `.gitignore`, or place the cache beneath an existing ignored per-worker runtime directory. If the earlier cache-path change removes these separate directories, no additional change is needed.

kjmph (commented):
In addition to the inline review comments, I see a few broader or lower-priority items:

- The PR description says `result_dir_<jobid>/` and `latest_result_dir.txt` were verified, but that behavior does not appear to be implemented by this PR's current head. The diff only adds related `.gitignore` entries. It looks like some validation may have included stacked changes from `MultiProfile` or `bench33.patch`. Could we rerun the stated validation from a clean checkout of this exact PR head, then either include the missing dependencies or update the checklist?

- `--iterations=0 --metrics` reaches the metrics collection path without ever assigning a query cursor. It would be good to reject non-positive iteration counts during argument validation.

- `PRESTO_SESSION_PROPERTIES` currently accepts an empty property name such as `=value`. Could we reject that locally and report which entry is malformed?

- The CPU template comments advertise `CPU_*` overrides for the local-exchange and page-size settings, but the corresponding generator logic is not present in this PR. We should either port that plumbing or revise the comments so they describe what this patch supports.

The inline comments contain the changes I consider most important. My main overall feedback is ensuring that the final patch is self-contained.

kjmph (changes requested):

presto/docker/config/template/overrides/gpu/etc_coordinator/config_native.properties
misiugodfrey: I'll go with the 128MB pairing for `max-content-length` for now.  Unless you think we should leave both as default and refine it further once we do some explicit testing?

presto/docker/config/template/overrides/gpu/etc_worker/config_native.properties
misiugodfrey: Removed from both cpu/gpu worker properties.  It looks like the workers silently ignore this option

presto/slurm/presto-nvl72/functions.sh
misiugodfrey: I'll go with `LIBCUDF_KERNEL_CACHE_PATH` for now, but will update if we get further feedback.

presto/docker/config/template/etc_coordinator/catalog/hive.properties
misiugodfrey: I'll add it to java for now, although I'm pretty confident that we never confirmed if we want SOFT_AFFINITY for java runs or not.  I think it makes sense not to change more behavior than intended - but I also want to flag that adding an explicit option here implies that this option's affect on the java worker is more empirically measured than it is.

presto/testing/performance_benchmarks/common_fixtures.py
misiugodfrey: I've added `sessions_properties` in the benchmark_results.json's context and the format is now documented.

presto/docker/config/template/overrides/cpu/etc_worker/config_native.properties
misiugodfrey: Most of the generator logic is deferred to the next PR (#391), but some of it leaked over.  For now I'll have these switch to the 512MB values by default (the intended multi-worker values) and the subsequent PR can add the single-worker configuration.

presto/scripts/common_functions.sh
misiugodfrey: I removed the PATH prefix and the noted jq dependencies from the slurm paths.  There is still one instance of `jq` that runs on the host which is harder to remove.  If that ends up being problematic we can put in more effort to strip it out.

presto/slurm/presto-nvl72/functions.sh
misiugodfrey: Added to `.gitignore` for now.

presto/testing/performance_benchmarks/common_fixtures.py
```diff
@@ -167,47 +207,75 @@ def benchmark_query(request, presto_cursor, benchmark_queries, benchmark_result_
...
+                        )
+                    else:
+                        query = "--" + str(benchmark_type) + "_" + str(query_id) + "--" + "\n" + benchmark_queries[query_id]
+
+                    cursor = presto_cursor.execute(query)
+
+                    if ctas_results is not None:
+                        # CTAS returns only its update count to the client. Consuming it
+                        # ensures all worker writes have completed before recording stats.
+                        cursor.fetchall()
```
misiugodfrey, line 257: There's some complication here between how we want to measure things during profiling, CTAS, etc...  Right now the profiling covers CTAS table creation and teardown - which may not be of interest, but I expect it's still ok for now?  If we need to prune that out we can re-write when the profiling starts/stops.

Issue #164, opened 2026-01-15
Generated Data should be tracked in the Benchmarks database


felipeblazing:
If we want to standardize on benchmark runs we also need to standardize on the sources of data that we are using for various runs. To make sure that we can compare apples to apples we should register the datasets that we generate with the benchmarking database so that all tools trying to run benchmarks can run them on a dataset that has been registered with the database.

There is certain information about datasets that would be very convenient to be able to quickly ascertain via queries against the database for each generated dataset:

- Data Storage Location - where is this data being stored. This should be some kind of ID that we attach to each of the storage devices that we are using. Vast on Coreweave, S3 bucket(s) on AWS, etc .
- Benchmark - what benchmark is this data for
- Scale Factor
- Generator - what actually made these files? arrow, duckdb, cudf, etc.
- Compression type - what kind of compression is being used on the row groups, zstd, snappy, etc.
- decimal 128/64, float32/64 - what are we using to encode numbers naively hoping that people generally tend to pick one for these benchmarks
- row group size type - are we limiting by rows or bytes
- row group size - the amount we are limiting our row groups to
- number of row groups
- file format
- file size average
- user defined name
- user defined notes
- file_path_root - could be a file path, could be an s3 bucket, I am assuming within these we can agree each table gets its own directory and these names should be standard across benchmarks

This would allow us to compare performance across different settings for generating these files and both help us find the data we need to run benchmarks and understand the performance impacts of different configurations for data generation.

PR #396, opened 2026-07-31
Add enriched dataset metadata and storage config registration

### LINKED ISSUE #388: Automate post_results options
Currently the `post_results.py` script takes many parameters that are input manually for the user to post details about a given benchmark run, this includes the storage-configuration, the commits the images were built with, the harware SKU, etc...

This process is unfortunately quite prone to failure, since the user can easily input an option that does not match what the benchmark was actually doing.  For instance, there is no way for the script to verify if the storage-configuration that the user claims they ran with is what the benchmark actually ran on.

We should be able to automate the inputs for many of these options (or at least automate verification)

### LINKED ISSUE #389: Have benchmark generation include more info in metadata.json
Currently we store many benchmark-configurations in the benchmarking database which the `post-results.py` script matches via the `--storage-configuration` option.  We should have our data generation scripts add information to the metadata.json file about how the data were generated and what benchmark configuration they should match; that way we can have the script automatically detect, or otherwise verify, the storage configuration for a given run instead of blindly trusting the user's input.

This will also hopefully help us reduce the number of potential duplicate storage configurations if it is clear how they differ or not.

misiugodfrey:
## Motivation

Closes #389, closes #388, partially addresses #164.

When running benchmarks and posting results via `post_results.py`, the `--storage-configuration-name` flag has to be supplied manually by the user. There is no mechanism to verify that the storage configuration matches the actual dataset the benchmark ran on — a mismatch is silent and corrupts the database record. Similarly, `metadata.json` written by `generate_data_files.py` only recorded `scale_factor` and `approx_row_group_bytes`, making it impossible to fully reproduce a dataset or trace what configuration produced it.

## Summary

- **Enriched `metadata.json`**: `generate_data_files.py` now writes a complete record of how a dataset was generated — benchmark type, generator (tpchgen/duckdb) and version, scale factor, parquet version, codec definitions, row group settings, and the absolute data path. The schema is self-contained enough to fully replicate a dataset from metadata alone.

- **New `register_storage_config.py`**: Standalone CLI that registers a dataset as a storage configuration in the benchmarking database (`POST /api/storage-configs/`). Auto-detects storage system type (via `findmnt` + `lsblk`) and GDS status (via `lsmod`); derives compression from codec definitions; auto-generates the config name as `{machine}-{storage_system}-{benchmark_type}-sf{N}`. Does a pre-flight `GET` to prevent duplicate names. Writes `storage_configuration_name` back into `metadata.json` after successful registration. Also callable inline via a `--register` flag group on `generate_data_files.py` for simple generate-and-register workflows — datasets that need modification before registration can call the script separately.

- **Auto-detection and validation in `post_results.py`**: `--storage-configuration-name` is now optional. The script reads `storage_configuration_name` from `metadata.json` in the dataset directory (via `data_dir` in `benchmark_result.json` context). When a name is provided explicitly, it is validated against the registered name in `metadata.json`:
  - Matches registered name → proceeds silently
  - Differs from registered name → warning (mismatch)
  - No registered name in metadata → warning (could not validate)
  - Not provided and registered name found → auto-detected
  - Not provided and no registered name → error pointing to `register_storage_config.py`

## Test plan

- [x] Generate TPC-H dataset via tpchgen and verify all new fields in `metadata.json`
- [x] Generate TPC-H dataset via DuckDB and verify tpchgen-only fields are absent
- [x] Generate TPC-DS dataset and verify `benchmark_type` and labels are correct
- [x] `--convert-decimals-to-floats` reflected correctly in `metadata.json`
- [x] `register_storage_config.py --dry-run` on tpchgen dataset — verify payload shape, auto-detected fields, auto-generated name, and `codec_definitions` embedded in `table_metadata`
- [x] `register_storage_config.py --dry-run` on DuckDB dataset — verify tpchgen-only fields absent from `table_metadata`
- [x] `--name` override on `register_storage_config.py` — verified name used as-is without auto-generation
- [x] `--register` inline flag on `generate_data_files.py` with `--register-dry-run`
- [x] `post_results.py --dry-run` without `--storage-configuration-name` — verified auto-detection from `metadata.json` via `data_dir` in benchmark context
- [x] `post_results.py` with mismatched `--storage-configuration-name` — warning printed, provided value used
- [x] `post_results.py` with manually provided `--storage-configuration-name` and no registered name in `metadata.json` — warning printed
- [x] `post_results.py` with no `metadata.json` and no flag — helpful error message printed
- [x] `_resolve_storage_configuration` returns `None` with helpful message when `storage_configuration_name` absent from `metadata.json`
- [x] `_resolve_storage_configuration` returns `None` with helpful message when no `metadata.json` found
- [x] `BenchmarkMetadata.data_dir` correctly populated from benchmark result context
- [ ] `register_storage_config.py` with a live API key — verify pre-flight collision check and `storage_configuration_name` written back to `metadata.json`


benchmark_data_tools/generate_data_files.py
```diff
@@ -99,12 +101,43 @@ def generate_data_files(args):
             print("generating with duckdb")
         generate_data_files_with_duckdb(args)
 
+    if args.register:
+        if args.verbose:
+            print("registering storage configuration")
+        rc = register_storage_config(
+            data_dir=Path(args.data_dir_path),
+            machine=args.register_machine,
+            name=args.register_name,
```
misiugodfrey, line 111: Most of these should be automatically determined, but they can be manually overwritten via flag if desired.

benchmark_data_tools/generate_data_files.py
```diff
@@ -417,5 +463,68 @@ def build_default_codec_defs():
...
+        "For datasets that will be modified before registration, run register_storage_config.py "
+        "separately instead.",
+    )
+    reg.add_argument(
+        "--register",
+        action="store_true",
+        default=False,
+        help="Register the dataset as a storage configuration after generation.",
+    )
+    reg.add_argument(
```
misiugodfrey, line 480: It's debatable if we want to include these overrides; they should usually not be used.

benchmark_reporting_tools/post_results.py
```diff
@@ -734,6 +774,32 @@ async def _process_benchmark_dir(
...
+        if resolved_storage_config is None:
+            print(
+                "  Error: --storage-configuration-name was not provided and could not be "
+                "auto-detected. Run register_storage_config.py on the dataset directory first.",
+                file=sys.stderr,
+            )
+            return 1
+        storage_configuration_name = resolved_storage_config
+    elif resolved_storage_config and resolved_storage_config != storage_configuration_name:
+        print(
```
misiugodfrey, line 819: Right now we give some deference to the user when there is a mismatch - assuming the user wants to submit using an older storage configuration and they know what they are doing.  We could make this a proper error, but right now I don't want to block older workflows.

Perhaps we strengthen this once we have generated new datasets and hardened their usage?

benchmark_data_tools/register_storage_config.py
```diff
@@ -0,0 +1,438 @@
...
+    except (FileNotFoundError, Exception):
+        pass
+
+    return "local"
+
+
+def _detect_gds_enabled() -> bool:
+    """Return True if the nvidia-fs kernel module (GPU Direct Storage) is loaded."""
+    try:
+        result = subprocess.run(["lsmod"], capture_output=True, text=True)
```
TomAugspurger, line 111: Just a note that I think this is reflecting an issue with the database's data model: whether some dataset is over accessible over GDS is less about the dataset itself and more about the compute node and its path to the storage system. This check doesn't touch the dataset or its storage system: just the system that `lsmod` is running on.

Nothing to do for now, but LMK if this felt awkward to you too and maybe we can find a better way to model that.

misiugodfrey: That's a fair point.  I think for now there should not be a difference since the system should always be running on the same nodes that access the data (this is true even on our slurm systems); but it's something I'll confirm in testing.

Avinash-Raj:
@misiugodfrey Can we have a custom decimal dataset generation script for testing decimal arithmetic errors such as ADD/SUB/MUL/MOD overflows and DivideByZero?

misiugodfrey:
> @misiugodfrey Can we have a custom decimal dataset generation script for testing decimal arithmetic errors such as ADD/SUB/MUL/MOD overflows and DivideByZero?

I think that's a reasonable ask - but outside the focus of this PR.  You should open a separate issue detailing this request.

bdice (approved):
Looks fine. Seems like there was good discussion and good questions raised, but I think there are none that should block this.

misiugodfrey:
Going to re-test this on the cluster before merging

benchmark_data_tools/generate_data_files.py
```diff
@@ -99,12 +102,43 @@ def generate_data_files(args):
             print("generating with duckdb")
         generate_data_files_with_duckdb(args)
 
+    if args.register:
+        if args.verbose:
+            print("registering storage configuration")
+        rc = register_storage_config(
+            data_dir=Path(args.data_dir_path),
+            machine=args.register_machine,
```
paul-aiyedun: I don't think the data generation script should be concerned about the host machine, GDS being enabled, region, etc. This seems like something that should be happening outside the script and using the generated `metadata.json` file.

misiugodfrey:
Re-tested everything now that the cluster is available.  Looks to be in order.

PR #398, opened 2026-08-03
Set libcudf kernel cache path

bdice:
## Description

Set `LIBCUDF_KERNEL_CACHE_PATH=/var/lib/presto/data/libcudf-cache` in the native Presto image.

Without an explicit path, cuDF uses the root user cache directory (`/root/.cache/libcudf`). The Presto container filesystem is read-only, so that directory cannot be created. Older cuDF versions silently continued without a file cache. [rapidsai/cudf#22975](https://github.com/rapidsai/cudf/pull/22975) makes failure to select or create a writable kernel-cache directory an error, which turns this into a startup failure.

The Presto data directory is writable, so configure it as the kernel-cache location.

### DIFF
--- a/presto/docker/native_build.dockerfile
+++ b/presto/docker/native_build.dockerfile
@@ -112,6 +112,8 @@ RUN mkdir /usr/lib64/presto-native-libs && \
 
 COPY velox-testing/presto/docker/launch_presto_servers.sh velox-testing/presto/docker/presto_profiling_wrapper.sh /opt
 
+ENV LIBCUDF_KERNEL_CACHE_PATH=/var/lib/presto/data/libcudf-cache
+
 ARG PRESTO_SHA
 ARG PRESTO_BRANCH
 ARG PRESTO_REPOSITORY


simoneves (approved):

presto/docker/native_build.dockerfile
```diff
@@ -112,6 +112,8 @@ RUN mkdir /usr/lib64/presto-native-libs && \
 
 COPY velox-testing/presto/docker/launch_presto_servers.sh velox-testing/presto/docker/presto_profiling_wrapper.sh /opt
 
+ENV LIBCUDF_KERNEL_CACHE_PATH=/var/lib/presto/data/libcudf-cache
```
bdice, line 115: This sets the environment variable for both GPU and CPU workers, because I didn't see an easy way to set it for only GPU workers. If you have alternative ideas, please feel free to push to this PR.

kjmph:
Note, @misiugodfrey is working on a fix by making a mountable write point, yet it has a different name than I expected. I was digging into this, and was going to provide a comment on PR #390:
https://github.com/rapidsai/velox-testing/pull/390/changes#diff-449962442591b98025be548738403996b56879760d531c6cd425816020d6cec6R239-R241

lamarrr (approved):
Changes LGTM. Thanks for working on this!

mattgara (approved):

PR #399, opened 2026-08-04
Fix validation of decimal benchmark results

shrshi:
## Summary

Fix benchmark validation for Presto DECIMAL result columns.                                                                                                       
                                                                                                                                                                       
The Presto Python client returns DECIMAL values as strings. Benchmark results were therefore written as Parquet string columns, causing the validator to compare DECIMAL values lexicographically during `ORDER BY` validation.

## Changes:
                                                                                                                                                 
- Capture Presto result types from cursor.description.                                                                                                            
- Convert declared DECIMAL columns to Python Decimal values before writing benchmark Parquet results.                                                             
- Store the complete Presto result types in per-query qN.types.json sidecar files.                                                                                
- Load the sidecar during benchmark validation.                                                                                                                   
- Fall back to Parquet decimal logical types when sidecar metadata is unavailable.                                                                                
- Normalize only columns explicitly declared as decimal(p,s).                                                                                                     
- Preserve numeric-looking VARCHAR columns as strings.                                                                                                            
- Pass result types through the shared comparison path so DECIMAL ordering and comparison use numeric values.                                                     
- Add regression coverage for numeric DECIMAL ordering and VARCHAR preservation.

### DIFF
--- a/common/testing/result_comparison.py
+++ b/common/testing/result_comparison.py
@@ -12,6 +12,7 @@ parquet files). See compare_result_frames for full comparison semantics.
 import datetime
 import re
 import warnings
+from decimal import Decimal
 from typing import Literal
 
 import numpy as np
@@ -160,6 +161,23 @@ def _normalize_to_expected(actual: pd.DataFrame, expected: pd.DataFrame) -> pd.D
             # parse the strings and yield datetime.date to match.
             out.isetitem(i, pd.to_datetime(a_col).dt.date)
     return out
+
+
+def normalize_decimal_columns(frame: pd.DataFrame, column_types: list[str] | None) -> pd.DataFrame:
+    """Restore DECIMAL values using engine types rather than value inference."""
+    if not column_types:
+        return frame
+
+    frame = frame.copy()
+    for i, type_name in enumerate(column_types):
+        if i >= frame.shape[1] or _decimal_abs_tolerance(type_name) is None:
+            continue
+        column = frame.iloc[:, i]
+        frame.isetitem(
+            i,
+            column.map(lambda value: value if pd.isna(value) else Decimal(str(value))),
+        )
+    return frame
 
 
 # ---------------------------------------------------------------------------
@@ -418,7 +436,7 @@ def compare_result_frames(
 
     Steps:
       1. Validate column count and capture expected column names.
-      2. Normalize dtypes / value types.
+      2. Restore declared DECIMAL columns and normalize dtypes / value types.
       3. Validate row count.
       4. Parse ORDER BY / LIMIT from SQL.
       5. Validate each engine respected ORDER BY (per-frame, strict equality).
@@ -436,7 +454,10 @@ def compare_result_frames(
     expected_col_names = list(expected.columns)
     abs_tolerances = [_decimal_abs_tolerance(type_name) or ABS_TOL for type_name in actual_column_types or []]
 
-    # 2. Coerce actual's dtypes/value types to match expected's
+    # 2. Restore typed decimal values before generic dtype alignment. Presto's
+    # Python client and benchmark Parquet files represent DECIMAL as strings.
+    actual = normalize_decimal_columns(actual, actual_column_types)
+    expected = normalize_decimal_columns(expected, actual_column_types)
     actual = _normalize_to_expected(actual, expected)
 
     # 3. Row count
@@ -492,6 +513,7 @@ def validate_query_result(
     actual: pd.DataFrame,
     expected: pd.DataFrame,
     query_sql: str,
+    actual_column_types: list[str] | None = None,
 ) -> tuple[ValidationStatus, str | None]:
     """
     Compare actual vs expected for one query. Returns (status, message).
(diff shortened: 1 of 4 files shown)


common/testing/result_comparison.py
```diff
@@ -162,6 +163,23 @@ def _normalize_to_expected(actual: pd.DataFrame, expected: pd.DataFrame) -> pd.D
...
+        return frame
+
+    frame = frame.copy()
+    for i, type_name in enumerate(column_types):
+        if i >= frame.shape[1] or _decimal_abs_tolerance(type_name) is None:
+            continue
+        column = frame.iloc[:, i]
+        frame.isetitem(
+            i,
+            column.map(lambda value: value if pd.isna(value) else Decimal(str(value))),
```
misiugodfrey, line 178: The actual Decimal objects here are created based on the value in the DataFrame, not the column_type; so the column_type passed in is only used to determine if a type is a decimal - the precision/scale values are not relevant as they will be determined by the DataFrame's value instead.

If that's the case (and we want to keep it that way, i.e. use the DataFrame as the source of truth) then it should not matter if we pass type info from DuckDB or Presto for the column_types as only their precision/scale should differ.

presto/testing/performance_benchmarks/common_fixtures.py
```diff
@@ -116,13 +118,17 @@ def benchmark_query_function(query_id):
...
+                    column_types = [desc[1] for desc in cursor.description]
                     df = pd.DataFrame(rows, columns=columns)
+                    df = normalize_decimal_columns(df, column_types)
 
                     # Save to Parquet format to match expected results
                     results_dir = Path(f"{bench_output_dir}/query_results")
                     results_dir.mkdir(parents=True, exist_ok=True)
                     parquet_path = results_dir / f"{query_id.lower()}.parquet"
                     df.to_parquet(parquet_path, index=False)
+                    types_path = results_dir / f"{query_id.lower()}.types.json"
```
misiugodfrey, line 130: If the `df` is now being written with the decimal types preserved I'm not sure we need the additional `types.json` file?  I understand that this provides us with the additional precision/scale information that Presto has calculated, but I don't think we use that information anywhere (Decimal comparisons are exact).  If we think that information is necessary in the validation, then we should incorporate it into the comparison, but right now I think we are writing this new file to track information (is_decimal) that can be identified from the expected_results file, or from the Presto column types when written.

common/testing/validate_results.py
```diff
@@ -27,6 +29,22 @@
 # ---------------------------------------------------------------------------
 
 
+def _get_column_types(result_file: Path) -> list[str] | None:
+    types_file = result_file.with_suffix(".types.json")
+    if types_file.exists():
+        return json.loads(types_file.read_text())
+
+    inferred_types = []
```
misiugodfrey, line 37: So from what I understand, we have three sources of type information:
- Presto types (now written to the `types.json` file).  These should contain the decimal types and will have precison/scale values calculated by Presto - these may differ from the DuckDB prevision/scale specified in the expected_results files for computed expressions but should be the same for table columns).
- DuckDB types.  These are the types written in the expected_results files and will not have changed after this PR.
- Result file types.  These are what we are writing when doing benchmarks and prior to this PR were being written without preserving the decimal types causing the comparison errors.

If I understand the issue correctly, the main problem was that the result file types were being written without preserving the decimal type, and therefore we were getting ordering errors because decimals were being compare lexicographically.  However, I don't think that adding this file tracks new information (at least not in a way that it is used).

common/testing/validate_results.py
```diff
@@ -27,6 +29,22 @@
 # ---------------------------------------------------------------------------
 
 
+def _get_column_types(result_file: Path) -> list[str] | None:
```
misiugodfrey, line 32: I think it's a little misleading to have a function called `_get_column_types()` return an empty list if none of the columns are decimals.  If we do need this function to have this shape, then maybe we rename it?

common/testing/validate_results.py
```diff
@@ -86,6 +104,7 @@ def validate(
 
         actual = pd.read_parquet(result_file)
         expected = pd.read_parquet(expected_file)
+        actual_column_types = _get_column_types(result_file)
```
misiugodfrey, line 107: The `actual_column_types` we get here is only used in `validate_query_results` to determine if a column is decimal or not (the precision values are not taken into account).  To that end, I think we can use the types from the expected_file instead of the results_file (and throw an error if the types do not align with the expected file).  This eliminates the need for the sidecar file.

misiugodfrey (changes requested):
I think the main fix to make sure the decimal types are preserved in the results files should work.  However, some of the other additions here are - to my eye - either not necessary, or need further changes to refine our decimal comparisons.

Specifically, we are adding a new type.json file to track the calculated Presto types which could have different precision values than DuckDB - but we aren't using that information to refine our comparison, we are just using it to know if a type is a decimal, which is information we can already get from the expected_results file or from the results file itself now that the types are being written correctly.

This brings up that right now our decimal comparisons are absolute.  Any differences in the calculated decimal types between Presto and DuckDB are not reflected in our evaluation - and that may be fine?  If it is - then I think we can avoid the need to write a type.json file and keep things simpler.  If it is not fine, then we can write the sidecar, but also incorporate the precision differences into our validation.

PR #412, opened 2026-09-02
Enable EFA-backed Presto GPU workers

devavret:
## Summary

- bundle checksum-pinned UCX 1.22.0 into native GPU worker images, including explicit CUDA and EFA support
- verify the UCX version, library origin, plugins, transports, and requested devices during image build and worker startup
- add an opt-in `--ucx-efa` launch mode with host networking, EFA device access, and per-worker `UCX_NET_DEVICES`
- keep EFA-specific UCXX settings in an opt-in configuration overlay and preserve CPU-only worker behavior
- enable native cuDF exchange only for GPU benchmark queries, after metadata discovery
- add a benchmark-independent g7e.48xlarge example that discovers GPU/EFA topology and starts an eight-worker cluster

## Motivation

Presto GPU workers need explicit EFA/SRD device and network configuration to move CUDA exchange buffers between GPUs that cannot use PCIe peer access. Keeping UCX inside the GPU worker image removes the need for a host UCX checkout or mounted runtime and prevents mismatches between build headers, linked libraries, and dynamically loaded plugins.


bdice (approved):

presto/docker/native_build.dockerfile
```diff
@@ -7,6 +7,31 @@ RUN rpm --import https://developer.download.nvidia.com/compute/cuda/repos/ubuntu
     dnf install -y nsight-systems-cli-2026.3.1 numactl
 
 ARG GPU=ON
+ARG UCX_VERSION=1.22.0
```
kjmph: The dependency base image already defines ENV UCX_VERSION=1.19.0. Docker gives an ENV precedence over an ARG with the same name during RUN, so the wget below still downloads UCX 1.19.0 even when Compose passes 1.22.0.

Could we use a non-colliding build argument, for example:

```dockerfile
ARG BUNDLED_UCX_VERSION=1.22.0
ENV UCX_VERSION=${BUNDLED_UCX_VERSION}
```

and use `${BUNDLED_UCX_VERSION}` in the download URL? In `docker-compose.common.yml`, the corresponding argument could be:

```yaml
- BUNDLED_UCX_VERSION=${UCX_VERSION:-1.22.0}
```

An assertion such as `/usr/local/bin/ucx_info -v | grep -F "${BUNDLED_UCX_VERSION}"` after installation would also prevent this from silently recurring.

presto/docker/native_build.dockerfile
```diff
@@ -7,6 +7,31 @@ RUN rpm --import https://developer.download.nvidia.com/compute/cuda/repos/ubuntu
...
+      | tar -C /tmp/ucx-src --strip-components=1 -xz && \
+    cd /tmp/ucx-build && \
+    /tmp/ucx-src/contrib/configure-release \
+      --prefix=/usr/local \
+      --with-sysroot \
+      --enable-cma \
+      --enable-mt \
+      --with-gnu-ld \
+      --with-rdmacm \
+      --with-verbs \
```
kjmph, line 28: Since this build specifically promises EFA/SRD support, should we pass `--with-efa` explicitly here?

UCX defaults EFA support so that f the required EFA headers or library are unavailable, configuration succeeds but silently omits the SRD transport. The failure would then occur only when a worker starts with `UCX_TLS=srd`. Perhaps we enforce this?

```bash
      --with-verbs \
      --with-efa \
```

With the explicit option, `configure-release` fails when EFA support cannot be compiled.

presto/scripts/start_presto_helper_parse_args.sh
```diff
@@ -275,3 +291,15 @@ if [[ -n $GPU_IDS ]]; then
     exit 1
   fi
 fi
+
+if [[ "$UCX_EFA" == true ]]; then
+  : "${PRESTO_WORKER_INTERNAL_ADDRESS:?set PRESTO_WORKER_INTERNAL_ADDRESS for --ucx-efa}"
+  IFS=',' read -ra ucx_gpu_ids <<< "${GPU_IDS:-0}"
```
kjmph: When `GPU_IDS` is omitted, the rendered default is `0..NUM_WORKERS-1`, but this fallback validates only GPU 0. For example, `--ucx-efa --num-workers 2` currently succeeds when `UCX_NET_DEVICES_GPU_1` is missing, then renders worker 1 with an empty device selector.

Could this derive the same default range as the renderer?

```bash
  if [[ -n ${GPU_IDS:-} ]]; then
    ucx_gpu_ids=("${GPU_ID_ARRAY[@]}")
  else
    mapfile -t ucx_gpu_ids < <(seq 0 "$((NUM_WORKERS - 1))")
  fi
```

presto/docker/config/template/overrides/gpu/etc_worker/config_native.properties
```diff
@@ -11,13 +11,17 @@ cudf.jit_expression_enabled=false
...
+ucxx.error_handling=false
+ucxx.blocking_polling=false
 
-# Use 100M rows per chunk for cudf partitioned output.
+# Bound partition output for SF3K Q18 memory pressure.
 # NOTE: This is not yet propagated to the worker properly because only a fixed set of query configs are supported.
 #       https://github.com/prestodb/presto/blob/a62672886152c8c6b61cf301d246f217d850e357/presto-native-execution/presto_cpp/main/PrestoToVeloxQueryConfig.cpp#L106-L224
 #       As a result, this needs to be hardcoded right now.
-cudf.partitioned_output_batch_rows=100000000
+cudf.partitioned_output_batch_rows=10000000
```
kjmph: These settings affect every GPU launch. Irrespective of `--ucx-efa`. The UCXX flags are described as EFA/SRD-specific, while the batch size is explicitly tuned for SF3K Q18. The consistent-hashing settings added to the shared coordinator template also affect every native CPU/GPU benchmark.

This means merging the opt-in EFA feature changes existing non-EFA benchmark and failure-handling behavior. Could we leave the shared defaults unchanged and apply these through an explicit EFA/Q18
configuration profile selected by the new runner?

I suggest separating:

- EFA transport requirements (ucxx.error_handling and polling) into an overlay enabled by `--ucx-efa`.
- Q18-specific batching and consistent hashing into a named benchmark profile selected by `run_g7e48_local_nvme_q18.sh`.

Gating all of these solely on `--ucx-efa` would still make the workload-specific tuning affect unrelated EFA benchmarks.

presto/docker/docker-compose/template/docker-compose.native-gpu.yml.jinja
```diff
@@ -30,19 +30,14 @@ x-presto-native-worker-gpu: &gpu_worker_base
...
-    #   - /dev/infiniband/uverbs3
-    #   - /dev/infiniband/uverbs4
-    #   - /dev/infiniband/uverbs5
-    #   - /dev/infiniband/uverbs6
-    #   - /dev/infiniband/uverbs7
-    #   - /dev/infiniband/uverbs8
-    #   - /dev/infiniband/uverbs9
+{% if ucx_efa %}
+  devices:
+    - /dev/infiniband/rdma_cm
```
kjmph, line 41: This check requires the four-card topology used by the benchmark, but the runner does not explain how that topology must be provisioned. A default `g7e.48xlarge` launch typically assigns an EFA interface only to network-card index 0. Interfaces for indices 1–3 must be explicitly requested through the launch configuration/CLI. Such an instance reaches this check with only `uverbs0` and reports a missing device without explaining the prerequisite.

Since this script is presented as a reproducer, could we add either:

- exact AWS CLI/launch-template instructions for creating one EFA interface on each network-card index 0–3
- OR, a link to those instructions plus an actionable error such as: "Expected four EFA network cards; launch the instance with EFA interfaces assigned to NetworkCardIndex 0–3."

We can keep provisioning outside the benchmark script, but document and validate it here.

presto/scripts/run_g7e48_local_nvme_q18.sh
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env bash
```
kjmph: The new script is missing the repository-required SPDX header, which currently causes verify-copyright and check-style to fail.

```bash
#!/usr/bin/env bash
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
# SPDX-License-Identifier: Apache-2.0
```

kjmph (changes requested):

devavret:
@kjmph Thanks for the quick review. can you please take another look.

presto/docker/docker-compose/template/docker-compose.native-gpu.yml.jinja
```diff
@@ -6,6 +6,8 @@ x-presto-native-worker-gpu: &gpu_worker_base
   build:
     args:
       - GPU=ON
+      - BUNDLED_UCX_VERSION=${BUNDLED_UCX_VERSION:-1.22.0}
```
karthikeyann, line 9: Is this the latest version that is used by us today?

presto/docker/docker-compose/template/docker-compose.native-gpu.yml.jinja
karthikeyann: There is a launch template in our account that can do this.

presto/docker/docker-compose/template/docker-compose.native-gpu.yml.jinja
karthikeyann: @kjmph

kjmph: This is the version we've been testing, yes!

PR #419, opened 2026-09-08
feat(presto): add --cache-mode for reproducible cold/hot/lukewarm benchmark runs

karthikeyann:
## What

Adds `--cache-mode {lukewarm,hot,cold}` to the Presto TPC-H/TPC-DS pytest
benchmark harness (`run_benchmark.sh`), so cache-reset behavior for cold-IO
benchmarking is explicit and correct, instead of an accidental side effect
of a single session-start OS-cache drop.

## Why

The harness reported a `Lukewarm(ms)` column (iteration 0's raw time) next
to `Hot(ms)` stats (iterations 1..N-1), but the only automatic cache reset
was a single OS page-cache drop at the very start of the whole pytest
session — presto-native's own Velox memory/SSD cache and the Hive
connector's file-handle cache were never cleared at all. So "Lukewarm" was
only meaningfully cold for the very first query of a run; every later
query's "Lukewarm" number silently rode on whatever cache warmth earlier
queries left behind — not comparable across queries, query order, or runs.

## The modes

Adapted from @kjmph's [c5a43f9](https://github.com/kjmph/velox-testing/commit/c5a43f9fcb75b309c21f78ec3375c040d9554984), which contributed the `cold-once` shape, the idle-cluster check, exact success-body matching, and concurrent clears.

| Mode | Reset schedule | What gets reported |
|---|---|---|
| **`off`** (default) | Legacy behavior: OS page-cache drop once, **no worker cache clearing**. The only mode Java workers support. | `Lukewarm(ms)` + `Hot(ms)` — unchanged from today |
| **`lukewarm`** | Full reset **once**, before the first measured query. | `Lukewarm(ms)` + `Hot(ms)` |
| **`cold-once`** | Full reset before **each query's** iteration group. | `Cold(ms)` (iteration 0) + `Hot(ms)` (the rest) |
| **`cold`** | Full reset before **every iteration**. | `Cold(ms)` aggregates only |
| **`hot`** | **No resets.** First `--warmup-iterations` prime the cache. | `Hot(ms)` aggregates only — warm-ups never reported |

```
                    Q1 iter0    Q1 iter1    Q1 iter2    Q2 iter0    Q2 iter1
off/lukewarm  reset─┬───────────────────────────────────────────────────────▶
                    └─ once per run

cold-once     reset─┬─────────────────────▶ reset─┬─────────────────────────▶
                    └ Cold   └ Hot  └ Hot         └ Cold   └ Hot

cold          reset─┼─ reset─┼─ reset─┼─ reset─┼─ reset─┼─▶
                    └ Cold   └ Cold   └ Cold   └ Cold   └ Cold

hot           ──[ warm-up ][ measured ][ measured ]─────────────────────────▶
                 not reported    └ Hot      └ Hot
```

Cache clearing hits presto-native's `/v1/operation/{connector,server}/clearCache`
REST endpoints (GET, per `PrestoServer.cpp`'s `registerGet` — not POST, which
404s) on every worker discovered via the coordinator's `/v1/node`, concurrently,
followed by the OS page-cache drop.

**Resets fail closed.** The cluster must be idle before and after
(`AsyncDataCache::clear()` keeps *pinned* entries, so clearing mid-query is only
partly cold), every worker must be cleared, and responses are matched against the
exact success body — the endpoint answers `200` with `No memory cache set on
server` for no-ops, so trusting the status code reports warm numbers as cold.

A tier that simply isn't configured is *reported*, not failed: this repo's GPU
worker ships without an AsyncDataCache, so cold modes there control the
file-handle and OS page caches only. The run metadata records the layers actually
cleared, so it can't claim a reset that never happened.

**Known limitation:** `drop_cache()` writes `/proc/sys/vm/drop_caches` on the host
running pytest only. That covers workers sharing that kernel (the single-node
Docker setup this was tested on), but does nothing for remote workers — so `cold`
numbers from a multi-node cluster are not truly cold. A real per-worker drop would
be a follow-up.

## New flags

- `--cache-mode {off,lukewarm,cold-once,cold,hot}` (default `off`)
- `--warmup-iterations N` (default `1`) — leading iterations per query that prime the cache and are excluded from the aggregate stats. Validated at startup: must be `>= 0` and `< --iterations`.
- `--connector-id ID` (default `hive`) — the *catalog* to clear. The connector **type** is always `hive`: native's `clearConnectorCache` only implements `name == "hive"` and `VELOX_USER_FAIL`s on anything else.

## Compatibility

- The default is now `off`, which is behaviorally identical to today: OS page-cache drop once, no worker clearing, same reported columns, no `Cache Mode:` line. Existing baselines and Java clusters are unaffected.
- Opting into any other mode changes timings by design (they clear caches the old code never touched). `lukewarm`/`cold-once`/`cold` require native workers; the run fails fast with a clear message on `presto-java`.
- `Cache Mode:` / `cache_mode` / `cache_reset_layers` appear in the report and JSON context only for non-`off` modes (`post_results.py` filters context to known fields, so the extra keys are ignored there).
- `spark_gluten` never registers `--cache-mode`, so its report stays byte-identical.

## Testing

- `common/testing/cache_mode_reporting_test.py` (new, 16 cases): header/stat shape invariants across all modes and iteration counts, per-mode column expectations, the `spark_gluten` unregistered-`--cache-mode` fallback, and degenerate inputs (failed queries, `--warmup-iterations >= --iterations`). Matches the existing `*_test.py` convention in `common/testing/`; like its siblings there it is run manually, not by CI.
- Live runs against a presto-native GPU cluster (TPC-H SF10) for all three modes: reset cadence (1 reset for lukewarm, 0 for hot, one-per-iteration for cold), successful worker REST calls, and correct per-mode report columns.
- Failure path verified live with a bogus `--connector-id`: the run now aborts with a `RuntimeError` and reports `NULL` instead of printing plausible-looking `Cold(ms)` numbers from caches that were never cleared.
- `presto/testing/cache_reset_test.py` (new, 3 cases): all-workers-cleared succeeds and strips the `/v1/status` suffix, partial clear raises, no-nodes raises.

### DIFF
--- a/common/testing/cache_mode_reporting_test.py
+++ b/common/testing/cache_mode_reporting_test.py
@@ -1,0 +1,82 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for --cache-mode aware reporting in conftest.py."""
+
+import pytest
+
+from common.testing.performance_benchmarks.benchmark_keys import BenchmarkKeys
+from common.testing.performance_benchmarks.conftest import _agg_headers, compute_aggregate_timings
+
+CACHE_MODES = ["off", "lukewarm", "cold-once", "cold", "hot"]
+
+
+def _aggregate(timings, cache_mode="off", warmup_iterations=1):
+    """Run compute_aggregate_timings over one query and return (headers, stats)."""
+    results = {BenchmarkKeys.RAW_TIMES_KEY: {"Q1": timings}}
+    compute_aggregate_timings(results, cache_mode, warmup_iterations)
+    iterations = len(timings) if timings else 0
+    return _agg_headers(cache_mode, iterations), results[BenchmarkKeys.AGGREGATE_TIMES_KEY]["Q1"]
+
+
+@pytest.mark.parametrize("cache_mode", CACHE_MODES)
+@pytest.mark.parametrize("iterations", [1, 2, 5])
+def test_header_and_stats_shapes_match(cache_mode, iterations):
+    """Every mode/iteration combination must produce one stat per column.
+
+    pytest_terminal_summary and build_and_write_benchmark_result both assert this, so a
+    mismatch fails the benchmark run at reporting time.
+    """
+    headers, stats = _aggregate([100.0 + i for i in range(iterations)], cache_mode)
+    assert len(headers) == len(stats)
+
+
+def test_default_and_unregistered_modes_match_historical_layout():
+    """off is the default and the spark_gluten fallback; both must reproduce the
+    pre-cache-mode column layout exactly."""
+    headers, stats = _aggregate([100.0, 101.0, 102.0])
+    assert headers == ["Avg Hot(ms)", "Min Hot(ms)", "Max Hot(ms)", "Median Hot(ms)", "GMean Hot(ms)", "Lukewarm(ms)"]
+    assert len(stats) == 6
+
+
+def test_lukewarm_reports_first_iteration_separately():
+    headers, stats = _aggregate([200.0, 100.0, 102.0], "lukewarm")
+    assert headers[-1] == "Lukewarm(ms)"
+    assert stats[-1] == 200.0  # iteration 0, reported alone
+    assert stats[0] == pytest.approx(101.0)  # avg over iterations[1:]
+
+
+def test_hot_excludes_warmup_iterations():
+    """Warm-up iterations prime the cache and are not reported at all under hot."""
+    headers, stats = _aggregate([200.0, 300.0, 100.0, 102.0], "hot", warmup_iterations=2)
+    assert "Lukewarm(ms)" not in headers
+    assert stats[0] == pytest.approx(101.0)  # avg over iterations[2:] only
+    assert stats[2] == 102.0  # max excludes the 300.0 warm-up
+
+
+def test_cold_once_reports_cold_first_sample_then_hot():
+    """cold-once resets per query, so iteration 0 is a real cold sample and the rest are hot."""
+    headers, stats = _aggregate([200.0, 100.0, 102.0], "cold-once")
+    assert headers[-1] == "Cold(ms)"
+    assert headers[0] == "Avg Hot(ms)"
+    assert stats[-1] == 200.0  # the cold sample
+    assert stats[0] == pytest.approx(101.0)  # hot aggregate excludes it
+
+
+def test_cold_aggregates_all_iterations():
+    """Every cold iteration was independently reset, so none are excluded."""
+    headers, stats = _aggregate([100.0, 102.0, 200.0], "cold")
+    assert headers == ["Avg Cold(ms)", "Min Cold(ms)", "Max Cold(ms)", "Median Cold(ms)", "GMean Cold(ms)"]
+    assert stats[0] == pytest.approx(134.0)
+    assert stats[2] == 200.0
+
+
+@pytest.mark.parametrize("cache_mode", CACHE_MODES)
+def test_degenerate_inputs_do_not_break_reporting(cache_mode):
+    """A failed query (None timings) and warmup >= iterations must not raise."""
+    results = {BenchmarkKeys.RAW_TIMES_KEY: {"Q1": None}}
+    compute_aggregate_timings(results, cache_mode, 1)
+    assert results[BenchmarkKeys.AGGREGATE_TIMES_KEY]["Q1"] is None
+
+    headers, stats = _aggregate([100.0, 102.0], cache_mode, warmup_iterations=5)
+    assert len(headers) == len(stats)
--- a/common/testing/performance_benchmarks/conftest.py
+++ b/common/testing/performance_benchmarks/conftest.py
@@ -18,10 +18,43 @@ class DataLocation:
     key: str
 
 
+def _first_sample_label(cache_mode: str) -> str | None:
+    """Label for the standalone first-iteration column, or None if the mode has none.
+
+    off/lukewarm report iteration 0 as "Lukewarm"; cold-once resets before each query so
+    its iteration 0 is a genuine "Cold" sample followed by hot ones. cold (every
+    iteration is reset) and hot (warm-ups are never reported) have no first-sample
+    column. Single source of truth for _agg_headers, AGG_KEYS and
+    compute_aggregate_timings.
+    """
+    if cache_mode == "cold-once":
+        return "Cold"
+    if cache_mode in ("cold", "hot"):
+        return None
+    return "Lukewarm"
+
+
+def _agg_headers(cache_mode: str, iterations: int) -> list[str]:
+    """Column headers for the terminal report, per --cache-mode (see cache_reset.py for
+    the mode definitions and _first_sample_label for what each mode reports)."""
+    label = "Cold" if cache_mode == "cold" else "Hot"
+    first = _first_sample_label(cache_mode)
+    if iterations == 1:
+        return [f"{first}(ms)"] if first else [f"{label}(ms)"]
+    headers = [f"Avg {label}(ms)", f"Min {label}(ms)", f"Max {label}(ms)", f"Median {label}(ms)", f"GMean {label}(ms)"]
+    if first:
+        headers.append(f"{first}(ms)")
+    return headers
+
+
 def pytest_terminal_summary(terminalreporter, exitstatus, config):
     text_report = []
     iterations = config.getoption("--iterations")
     tag = config.getoption("--tag")
+    # None when the engine doesn't register --cache-mode (spark_gluten): keep its report
+    # byte-identical by omitting the mode line and falling back to lukewarm's layout.
+    cache_mode_opt = config.getoption("--cache-mode", default=None)
+    cache_mode = cache_mode_opt or "off"
     if not hasattr(terminalreporter._session, "benchmark_results"):
         return
 
@@ -38,22 +71,17 @@ def pytest_terminal_summary(terminalreporter, exitstatus, config):
 
         _write_line(terminalreporter, text_report, "")
         _write_line(terminalreporter, text_report, f"Iterations Count: {iterations}")
+        if cache_mode_opt and cache_mode_opt != "off":
+            mode_line = f"Cache Mode: {cache_mode_opt}"
+            if cache_mode_opt != "cold":
+                mode_line += f" (warmup iterations: {config.getoption('--warmup-iterations', default=1)})"
+            _write_line(terminalreporter, text_report, mode_line)
         _write_line(terminalreporter, text_report, f"{data_location_display_name} Name: {data_location_name}")
         if tag:
             _write_line(terminalreporter, text_report, f"Tag: {tag}")
         _write_line(terminalreporter, text_report, "")
 
-        if iterations > 1:
-            AGG_HEADERS = [
-                "Avg Hot(ms)",
-                "Min Hot(ms)",
-                "Max Hot(ms)",
-                "Median Hot(ms)",
-                "GMean Hot(ms)",
-                "Lukewarm(ms)",
-            ]
-        else:
-            AGG_HEADERS = ["Lukewarm(ms)"]
+        AGG_HEADERS = _agg_headers(cache_mode, iterations)
         width = max([len(agg_header) for agg_header in AGG_HEADERS])
         width = max(width, result[BenchmarkKeys.FORMAT_WIDTH_KEY]) + 2  # Additional padding on each side
         header = " Query ID "
@@ -112,27 +140,32 @@ def build_and_write_benchmark_result(session, json_result):
     before calling this function.
     """
     iterations = session.config.getoption("--iterations")
+    cache_mode = session.config.getoption("--cache-mode", default="off")
+    warmup_iterations = session.config.getoption("--warmup-iterations", default=1)
 
     bench_output_dir = get_output_dir(session.config)
     bench_output_dir.mkdir(parents=True, exist_ok=True)
 
-    if iterations > 1:
+    first_label = _first_sample_label(cache_mode)
+    first_key = BenchmarkKeys.COLD_KEY if first_label == "Cold" else BenchmarkKeys.LUKEWARM_KEY
+    if iterations == 1:
+        AGG_KEYS = [first_key] if first_label else [BenchmarkKeys.AVG_KEY]
+    else:
         AGG_KEYS = [
             BenchmarkKeys.AVG_KEY,
             BenchmarkKeys.MIN_KEY,
             BenchmarkKeys.MAX_KEY,
             BenchmarkKeys.MEDIAN_KEY,
             BenchmarkKeys.GMEAN_KEY,
-            BenchmarkKeys.LUKEWARM_KEY,
         ]
-    else:
-        AGG_KEYS = [BenchmarkKeys.LUKEWARM_KEY]
+        if first_label:
+            AGG_KEYS.append(first_key)
 
     if not hasattr(session, "benchmark_results"):
         return
 
     for benchmark_type, result in session.benchmark_results.items():
-        compute_aggregate_timings(result)
+        compute_aggregate_timings(result, cache_mode, warmup_iterations)
         json_result[benchmark_type] = {
             BenchmarkKeys.AGGREGATE_TIMES_KEY: {},
             BenchmarkKeys.RAW_TIMES_KEY: result[BenchmarkKeys.RAW_TIMES_KEY],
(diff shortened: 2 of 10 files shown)


presto/testing/performance_benchmarks/cache_reset.py
```diff
@@ -0,0 +1,69 @@
...
+        if not worker_uri:
+            print(f"Warning: worker node has no 'uri' in /v1/node response ({node}); skipping cache clear for it.")
+            continue
+        # /v1/node's "uri" field is the node's status-heartbeat URL (it includes a
+        # trailing /v1/status), not a bare base URI - strip that suffix before
+        # appending the /v1/operation/... admin paths, or every request 404s.
+        worker_uri = worker_uri.removesuffix("/v1/status")
+        fetch_text(f"{worker_uri}/v1/operation/connector/clearCache?name=hive&id={connector_id}")
+        fetch_text(f"{worker_uri}/v1/operation/server/clearCache?type=memory")
+        fetch_text(f"{worker_uri}/v1/operation/server/clearCache?type=ssd")
```
patdevinwilson: These `fetch_text` results are discarded. `fetch_text` already prints on failure and returns `None`, so a Java worker, bad URI suffix, or down node only becomes a log line — then `cold` still reports `Cold(ms)` as if the reset happened.

For `cold` (and arguably `lukewarm`’s one-time reset), please treat a failed clear as a failed reset: require at least one successful worker, and prefer failing the query/session if *all* three endpoints fail on a given worker. Soft-fail is reasonable for mixed Java/native discovery noise, but not if it can silently warm the numbers this flag exists to isolate.

Also, `name=hive` is hardcoded while `--connector-id` is configurable. Per the native `connector/clearCache` API, `name` is the connector type and `id` is the catalog. That’s correct for the default Hive catalog, but a non-hive catalog with `--connector-id` still hits `name=hive`. Either document that `name` is always the Hive connector type, or take a connector-name flag too.

Minor: `removesuffix("/v1/status")` misses a trailing slash (`/v1/status/`). `run_context.py` already documents that `/v1/node` is workers-only, so filtering by role is probably unnecessary.

common/testing/performance_benchmarks/conftest.py
```diff
@@ -43,17 +65,7 @@ def pytest_terminal_summary(terminalreporter, exitstatus, config):
             _write_line(terminalreporter, text_report, f"Tag: {tag}")
         _write_line(terminalreporter, text_report, "")
```
patdevinwilson, line 82: Please print `--cache-mode` (and warmup) next to `Iterations Count`. Column headers change (`Hot` vs `Cold`, presence of `Lukewarm`), and `benchmark_result.txt` is easy to mis-compare across runs without the mode in the saved report.

Same for `benchmark_result.json` context if that’s what post_results consumes.

presto/testing/performance_benchmarks/common_fixtures.py
```diff
@@ -84,6 +85,27 @@ def ctas_results(request, benchmark_queries):
...
+    No-op under hot (no resets) or cold (each iteration resets itself).
+    """
+    if request.config.getoption("--cache-mode") != "lukewarm":
+        return
+    hostname = request.config.getoption("--hostname")
+    port = request.config.getoption("--port")
+    connector_id = request.config.getoption("--connector-id")
+    skip_drop_cache = request.config.getoption("--skip-drop-cache")
+    cold_run_reset(hostname, port, connector_id, skip_os_drop=skip_drop_cache)
+    print("[CacheMode=lukewarm] Reset worker + OS caches once at session start.")
```
patdevinwilson: This session autouse fixture runs *before* `run_context_collector` and `verify_tables_analyzed` (both also session autouse, defined below). Those issue `SHOW TABLES` / `SHOW STATS` and a host-side `metadata.json` read, which is probably metastore-only — but anything that touches worker file handles after this reset will warm caches before Q1 iter0, which is exactly the “lukewarm isn’t comparable” bug this PR is fixing.

Safer: do the lukewarm reset immediately before the first measured iteration (same `cold_run_reset()` call, one-shot flag), after setup queries, so “once before the first query” is literally true.

Related pre-existing limit worth documenting on `--cache-mode=cold`: `drop_cache()` only writes `drop_caches` on the Docker host running pytest (`cache_utils.py`). That is a genuine OS drop when workers share that kernel; it does nothing to remote workers’ page cache.

patdevinwilson (commented):
Thanks for making cache-reset behavior explicit — the three-mode model and the GET `/v1/operation/...` usage look right, and skipping `drop_cache_once` when `--cache-mode` is registered is the correct split vs spark_gluten.

A few gaps between the PR write-up and the diff, plus one correctness issue for `--cache-mode=cold`:

1. **Claimed tests are not in this PR.** The description says a 17-case standalone suite for `compute_aggregate_timings` / `_agg_headers`. The file list is only harness/reporting changes — those tests should land here (or the Testing section should be updated).
2. **Default Presto behavior is not unchanged.** `lukewarm` now also clears Velox memory/SSD and Hive file-handle caches, not just the OS page cache. Reporting columns match the old default, but Q1 times (and anything that used to ride on a warm Velox cache) will move. Worth saying that in Compatibility rather than “unchanged.”
3. **Failed worker cache clears still produce “cold” numbers.** `fetch_text` already warns and returns `None`; this code ignores that. Under `cold`, a 404/timeout on every worker still proceeds, so the report can look like a successful cold run. I’d fail the iteration (or at least count successes and abort if zero workers cleared).

Inline notes below on the clear-cache loop and making the report self-describing.

karthikeyann:
**TL;DR** — all 8 points addressed in `dbc84214`. 7 fixed; 1 documented instead of auto-detected, with reasoning.

| # | Point | Action |
|---|---|---|
| 1 | Failed clears → fake "cold" numbers | **Fixed** — raises if no worker cleared; warns on partial |
| 2 | Lukewarm raced session fixtures | **Fixed** — one-shot before first measured iteration; fixture deleted |
| 3 | Report didn't record mode | **Fixed** — `Cache Mode:` line + JSON context keys |
| 4 | "Unchanged" compatibility claim | **Fixed** — PR description corrected; timings do move |
| 5 | Tests not in PR | **Added** — `cache_mode_reporting_test.py`, 16 cases |
| 6 | `name=hive` hardcoded | **Documented** — native only accepts `"hive"`; a flag can't help |
| 7 | Trailing-slash in `removesuffix` | **Fixed** |
| 8 | `drop_cache()` is pytest-host-only | **Documented, no auto-warning** — see #8 below for why |

<details>
<summary>Details</summary>

**1. Failed clears silently producing "cold" numbers** — fixed, and you were right that this was the important one. `clear_worker_data_caches` now tracks per-worker success and returns the count: it raises if *no* worker could be cleared, and warns loudly on partial success. I kept partial as a warning rather than a hard failure for the mixed-discovery case you mentioned.

Worth noting this bug was live during my own testing and I nearly missed it: before I found that `/v1/operation` is GET-only, every endpoint 404'd on every iteration and the run still printed a confident `Cold(ms)` table. Verified the new path with a bogus `--connector-id`:

```
Warning: cache clear incomplete on worker http://172.19.0.3:10000; its data may still be warm.
RuntimeError: Failed to clear caches on any of the 1 worker(s); refusing to report these timings as reset.
```
Exit 1, and the table reports `NULL` rather than plausible-looking numbers.

**2. Lukewarm racing the other session fixtures** — good catch, same bug class I thought I'd already eliminated. Moved the reset out of the session fixture to a one-shot immediately before the first measured iteration, so it lands after `run_context_collector` / `verify_tables_analyzed` do their metadata queries and "once, before the first query" is literally true. This also deleted the `cache_mode_worker_setup` fixture entirely — both resets now live in the same loop, differing only in cadence.

**3. Report is now self-describing** — `Cache Mode: cold` / `Cache Mode: hot (warmup iterations: 1)` next to `Iterations Count`, plus `cache_mode` / `warmup_iterations` in the JSON context. `post_results.py` filters context to known dataclass fields so the extra keys are ignored; wiring them through to the results DB would need a dataclass field + column, which I left out of scope.

**4. Compatibility claim** — you're right, "unchanged" was wrong. `lukewarm` now also clears the Velox and file-handle caches, so timings move even though columns don't. PR description corrected.

**5. Tests** — added `common/testing/cache_mode_reporting_test.py` (16 cases): shape invariants across modes/iterations, per-mode columns, the spark_gluten fallback, and degenerate inputs. Placed alongside the existing `*_test.py` files rather than in `performance_benchmarks/`, because that directory's `conftest.py` registers `pytest_sessionfinish` hooks needing engine-only options, so a bare unit test there fails at teardown (pre-existing — `pytest common/testing/` is already broken on `main` for this reason). Like its siblings it's manual, not CI-run; CI only runs `benchmark_data_tools/tests`.

**6. `name=hive`** — documented rather than made configurable: native's `clearConnectorCache` only implements `name == "hive"` and `VELOX_USER_FAIL`s on anything else, so a flag couldn't do anything useful today. The docstring now spells out that `name` is the connector *type* and `id`/`--connector-id` selects the catalog.

**7. Trailing slash** — now `rstrip("/").removesuffix("/v1/status")`.

**8. `drop_cache()` is pytest-host-only** — good point, documented in `cache_reset.py`'s module docstring: it covers workers sharing that kernel (single-node Docker) but does nothing for remote workers, so multi-node "cold" numbers aren't truly cold. I deliberately did *not* add an automatic warning — the only signals I could key off (worker count > 1, or non-loopback worker IP) both false-positive on the standard single-node Docker setup, where worker URIs are container IPs like `172.19.0.3` on the same kernel. I'd rather document the limitation than ship a check that cries wolf on the common path. Happy to do a real per-worker drop as a follow-up if you think multi-node cold runs are imminent.


</details>

patdevinwilson (commented):
I would approve if partial clears on cold are an accepted tradeoff for mixed clusters; otherwise I’d want cleared == len(nodes) (or equivalent) before treating this as merge-ready for reproducible cold-IO numbers.

**Nits (non-blocking)**
- Lukewarm’s one-shot reset sits inside the profile span on Q1; only cold warns about reset overhead in traces.
- Commit message still says lukewarm “preserves prior behavior”; the PR body is the accurate one.
- Tests cover reporting only, not cache_reset.py (you covered that live — fine if they stay manual).

karthikeyann:
**TL;DR** — went strict: `cleared == len(nodes)` or raise. All 3 nits fixed. `0328b117`.

| Point | Action |
|---|---|
| Partial clears on cold | **Strict now** — raises unless every worker cleared |
| Lukewarm reset inside profile span | **Fixed** — warning now covers lukewarm too |
| Commit message "preserves prior behavior" | **Fixed** — now says columns preserved, timings change |
| No tests for `cache_reset.py` | **Added** — `presto/testing/cache_reset_test.py`, 3 cases |

On the strict-vs-partial call: I don't think the mixed-cluster tradeoff applies here. `start_presto_helper.sh` sets a single `DOCKER_COMPOSE_FILE` (`java` | `native-cpu` | `native-gpu`), so a cluster is homogeneous — there's no supported way to get mixed Java/native workers. That makes a partial clear a symptom of something being wrong, not a cluster shape to tolerate, and soft-failing it contradicted the reasoning behind the fix in the first round. So: raise unless all workers cleared, with the count in the message. If a mixed-worker setup does become real, an explicit opt-out flag is the better answer than a silent default.

The new tests pin exactly that semantic (all-cleared succeeds, partial raises, no-nodes raises) so it can't quietly regress.

patdevinwilson (approved):
LGTM

presto/testing/performance_benchmarks/cache_reset.py
```diff
@@ -0,0 +1,96 @@
...
+            fetch_text(f"{worker_uri}/v1/operation/server/clearCache?type=memory"),
+            fetch_text(f"{worker_uri}/v1/operation/server/clearCache?type=ssd"),
+        ]
+        if all(response is not None for response in responses):
+            cleared += 1
+        else:
+            print(f"Warning: cache clear failed on worker {worker_uri}; its data is still warm.")
+
+    if cleared != len(nodes):
+        raise RuntimeError(
```
kjmph: Can we make this engine aware? Maybe default `--cache-mode` to `None`, as currently the default of `lukewarm` will raise here. This way, we preserve the existing Java/OS-drop-once path, and match the project goals of maintaining Java support.

presto/testing/performance_benchmarks/conftest.py
```diff
@@ -47,6 +48,28 @@ def pytest_addoption(parser):
...
+        default="lukewarm",
+        help=(
+            "Cache reset schedule (default: lukewarm); never restarts Presto. lukewarm: "
+            "reset once, before the first query, then run continuously. hot: no automatic "
+            "resets; --warmup-iterations primes the cache. cold: reset before every "
+            "iteration. See presto/testing/performance_benchmarks/cache_reset.py."
+        ),
+    )
+    parser.addoption(
+        "--warmup-iterations",
```
kjmph, line 66: Does the benchmark DB assume only execution 1 is the warmup? If so, --warmup-iterations > 1 cannot be represented because post_results.py drops this value.

common/testing/performance_benchmarks/conftest.py
```diff
@@ -189,25 +222,46 @@ def get_output_dir(config):
...
+                        f"Warning: {query_id} has only 1 iteration under --cache-mode=hot "
+                        f"(--warmup-iterations={warmup_iterations}); the reported 'Hot' value is "
+                        "actually an unwarmed/priming run, not a true steady-state measurement."
+                    )
+                # The lone iteration IS the sample under every mode: cold uses all of
+                # timings, hot/lukewarm's warmup-slice-or-fallback both reduce to
+                # timings[0] too, so there's nothing mode-specific to branch on here.
+                stats = (timings[0],)
+            else:
+                if cache_mode != "cold" and not timings[warmup_iterations:]:
```
kjmph, line 264: Can this fail validation instead of falling back? When `warmup_iterations >= iterations`, `timings[-1]` is still a warmup but is reported as `Hot`; negative counts should also be rejected.

presto/testing/performance_benchmarks/cache_reset.py
```diff
@@ -0,0 +1,96 @@
...
+    data on an un-cleared worker is still warm, and reporting that as cold defeats the
+    point of the reset. Clusters here are homogeneous (start_presto_helper.sh picks one
+    of java/native-cpu/native-gpu), so a partial clear means something is wrong rather
+    than a mixed-worker cluster.
+
+    `name=hive` is the connector *type*, not the catalog: presto-native's
+    clearConnectorCache only implements "hive" (anything else VELOX_USER_FAILs), while
+    `id` / --connector-id selects which catalog of that type to clear.
+    """
+    nodes = get_nodes(hostname, port)
```
kjmph, line 68: Can we retain the pre/post `/v1/query` idle checks from c5a43f9f? `AsyncDataCache::clear()` preserves pinned entries, so an active query can leave a nominally cold reset partly warm.

common/testing/performance_benchmarks/conftest.py
```diff
@@ -170,6 +193,16 @@ def pytest_sessionfinish(session, exitstatus):
     if tag:
         json_result[BenchmarkKeys.CONTEXT_KEY][BenchmarkKeys.TAG_KEY] = tag
 
+    # Only present for engines that register --cache-mode; the reported columns depend
+    # on it, so a saved report is ambiguous without it.
+    cache_mode = session.config.getoption("--cache-mode", default=None)
+    if cache_mode:
+        json_result[BenchmarkKeys.CONTEXT_KEY][BenchmarkKeys.CACHE_MODE_KEY] = cache_mode
```
kjmph: Can we also record `skip_drop_cache` or the effective reset layers? `--cache-mode cold --skip-drop-cache` currently produces the same saved metadata as a full cold reset.

common/testing/cache_mode_reporting_test.py
```diff
@@ -0,0 +1,73 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for --cache-mode aware reporting in conftest.py."""
```
kjmph, line 4: Should these new tests be wired into PR CI? Neither cache-mode test file is exercised by the current pytest job.

kjmph (changes requested):
Can we also consider a cold-once mode? This allows for cold to populate the cache, then subsequent queries use the populated cache. It is good for capturing cold on the first iteration, but letting this warm the cache for hot runs later.

karthikeyann:
**TL;DR** — adapted c5a43f9 into the PR (you're a co-author on the commit). All 7 points addressed in `fea4d865`. Your exact-body check found a real bug in my version: **this repo's GPU worker has no AsyncDataCache at all**, so every "cold" run I'd reported was clearing nothing on the Velox side.

| # | Point | Action |
|---|---|---|
| 1 | cold-once mode | **Added** — `--cache-mode=cold-once`, resets per query → `Cold(ms)` + `Hot(ms)` |
| 2 | Engine-aware default | **Fixed** — default is now `off` (legacy path); non-`off` modes fail fast on `presto-java` |
| 3 | `warmup_iterations` dropped by post_results | **Answered below** — not fixed here |
| 4 | Fail validation instead of falling back | **Fixed** — validated at startup; negatives and `warmup >= iterations` rejected |
| 5 | Retain `/v1/query` idle checks | **Adopted** from your commit, with a bounded wait |
| 6 | Record `skip_drop_cache` / reset layers | **Fixed** — records the layers *actually* cleared |
| 7 | Wire tests into CI | **Done** — new `test_benchmark_harness` job |

<details>
<summary>Details</summary>

**What your commit caught that mine didn't.** I was only checking HTTP status, so `No memory cache set on server` counted as success. With exact-body matching, live runs show this cluster's worker returns that for both `type=memory` and `type=ssd` — there is no Velox cache configured, so my earlier cold numbers came purely from the OS page-cache drop. I kept that from being fatal (it would make cold modes unusable on the default GPU config) but it is now reported, warned about once per run, and reflected in metadata:

```
Warning: no worker has a Velox memory cache configured, so there was none to clear.
"cache_reset_layers": ["hive-file-handle", "os-page-cache"]
```

Note the ssd request also answers with the *memory*-null body when no cache exists, since the null check precedes the type switch — worth knowing if you match on bodies too.

**Differences from c5a43f9, and why.** I kept the `--cache-mode` enum rather than two booleans, since it now spans five cadences and made `off` expressible. I also clear the SSD and file-handle tiers plus the OS page cache rather than memory only — the OS page cache is what actually dominates here given the above. And the once-per-query/session resets run *before* `start_profiler`, matching your ordering, so only `cold`'s per-iteration reset lands inside a trace.

**On #3 (post_results).** Confirmed: `BenchmarkMetadata.from_parsed` filters context to known dataclass fields, so `warmup_iterations` is dropped. Per-iteration rows *are* posted with `execution_number`, so the raw data survives — but nothing records which of those were warm-ups. Representing it needs a `BenchmarkMetadata` field and almost certainly a DB column, and there's an adjacent question of whether the required `--cache-state` arg (`cold/warm/hot/lukewarm`) should be derived from `cache_mode` instead of passed by hand. Both feel like a separate PR against the reporting tool rather than scope creep here — happy to file it.

Live-verified all five modes on TPC-H SF10; `cold-once` gives a clean split (Q6: 329 ms cold vs 131 ms hot avg).


</details>

kjmph (approved):

PR #422, opened 2026-09-10
feat(ci): Auto-fetch Velox staging PRs from github project board and add nightly staging workflow

Avinash-Raj:
## Summary

Switches staging PR auto-fetch from GitHub labels to the NVIDIA project board, and adds a `staging`
workflow that refreshes the Velox and Presto staging branches nightly.

Previously `create_staging_branch.sh` collected PRs by the `cudf` label, so staging contained whatever
happened to be labelled. There was no way to stage a PR without labelling it upstream, and no single
place to see what was queued. The board at [NVIDIA/projects/306](https://github.com/orgs/NVIDIA/projects/306/views/4) is already the team's source
of truth, so staging now reads it directly.

Image builds from the staging branch are **not** part of this PR; they live in #424.

## Where the staging branches are created

`staging.yml` runs at 03:00 UTC (plus `workflow_dispatch` / `workflow_call`) and fans out to two independent jobs with no dependency between them, so a Velox conflict does not block Presto:

| Job | Base | Target | Branches pushed | Script | Token |
|-----|------|--------|-----------------|--------|-------|
| `velox-staging` | `facebookincubator/velox@main` | `rapidsai/velox` | `staging`, `staging_MM-DD-YYYY` | `velox/scripts/create_staging.sh` | `VELOX_FORK_PAT` |
| `presto-staging` | `prestodb/presto@master` | `karthikeyann/presto` | `staging`, `staging_MM-DD-YYYY` | `presto/scripts/create_staging.sh` | `PRESTO_FORK_PAT` |

**Call chain for each job:**

```
staging.yml ──► velox-create-staging.yml ──► create-staging-composite.yml ──► [rapidsai/velox@staging]   

          └─► presto-create-staging.yml ──► create-staging-composite.yml ──► [karthikeyann/presto@staging]
```

`create-staging-composite.yml` holds the actual step sequence, and every step is one invocation of the
shared `scripts/create_staging_branch.sh`: clone the target repo, reset `staging` to the upstream base
commit, fetch the PR list from the board, test merges, merge, write `.staging-manifest.yaml`, and push.
Both `staging` and the dated snapshot go out in a single `git push` with two refspecs, so they cannot
diverge. `force_push: true` is required because `staging` is rebuilt from upstream every run and is
therefore never a fast-forward over the previous branch.

> [!IMPORTANT]
> `presto-staging` currently targets `karthikeyann/presto` for testing. This must be changed back to
> `rapidsai/presto` before merge, otherwise the nightly pushes to a personal fork — and
> `PRESTO_FORK_PAT` has no write access there, so the push step would fail.

### PR discovery (`scripts/create_staging_branch.sh`)

- Adds `fetch_prs_from_project()`, which reads the board with a single paginated `gh api graphql` call
  and keeps items whose `Velox Staging` field is `Staging`.
- The board is addressed by its **node ID** (`PVT_kwDOABpemM4BgGAW`) rather than owner + number, per
  review feedback. The ID survives the board being renamed, moved between owners, or renumbered, and
  it skips the owner-type lookup `gh project item-list` performs — which is what required `read:org`
  on top of `read:project` and produced the misleading `unknown owner type` failure in CI.
- The GraphQL response carries `state`, `isDraft`, and `repository` per PR, so closed and draft cards
  are filtered from the same response (no second `gh pr list` call) and the skip log reports the actual
  reason, e.g. `Skipping project PR #28398 (state=OPEN, draft=true)`.
- One board serves both projects: cards for either repository share the `Velox Staging` field, so
  results are separated by `--base-repository`. Constants are named `STAGING_PROJECT_*` to reflect
  that the script is shared rather than Velox-specific.
- Keeps label-based fetching: `--pr-labels` overrides the board for one-off runs.

### Empty board is a no-op, not a failure

Auto-fetch previously aborted with `No PRs found to merge` whenever nothing was marked `Staging`, so a
quiet night failed the nightly. `fetch_pr_list` now emits `PR_COUNT=0` and returns cleanly, and the
composite skips test-merge, test-pairwise, merge, manifest, and push on that count — the target branch
is left exactly as the last successful run produced it. An explicitly supplied PR list that resolves to
empty still fails, since that is a misconfiguration.

### Token handling

- Adds an optional `project_board_pat` secret, wired from `PROJECT_BOARD_TOKEN` and applied as
  `GH_TOKEN` on the Fetch PR list step only, so board scopes stay away from the clone, merge, and push
  steps. Only `read:project` is needed now that the board is addressed by node ID.
- Drops `GITHUB_TOKEN` to `contents: read` across the staging chain. Every git operation runs with the
  fork PAT, so the write grant was unused; it had to be repeated at each level only because a called
  workflow cannot be granted more than its caller.

### Workflow plumbing

- `velox-create-staging.yml` and `presto-create-staging.yml` each gain a `workflow_call` trigger so
  `staging.yml` can call them. Inputs are declared once and aliased with YAML anchors, matching
  `velox.yml`, so there is no duplicated input block to keep in sync.
- `target_repository` moves from `required: true` to `required: false` on both; it keeps its default,
  so the dispatch forms are unchanged.
- `auto_fetch_prs` now defaults to `true` for Presto as well as Velox, and the Presto wrapper script no
  longer forces `--auto-fetch-prs false`.

## Conflict handling

Conflicts are detected up front, never auto-resolved, and abort the run before anything is pushed.
Three gates:

1. **Each PR against base** (`test-merge`) — every PR is merged with `--no-commit --no-ff`, then
   aborted and reset to `BASE_COMMIT` so tests cannot contaminate each other. All PRs are checked
   before failing, so the log lists every conflicting PR with its URL rather than just the first.
2. **Every pair together** (`test-pairwise`) — all A+B combinations are tried; on failure the log
   includes a compatibility matrix plus a table of the involved PRs with author, title, and URL.
3. **The real merge** (`merge`) — a joint conflict among three or more PRs can survive both gates
   above, so the sequential merge aborts and exits non-zero as well.

Any of these fails the step and therefore the job. "Push branches" is the last step, so it never
executes: nothing reaches the target repo and `staging` remains as the last successful run left it. The
failure propagates up through `create-staging-composite` and the per-project workflow to the
`velox-staging` or `presto-staging` job.

The trade-off is bluntness: one conflicting PR costs that project's whole nightly refresh. The escape
hatch is `exclude_pr_numbers`, reachable by dispatching the per-project workflow directly, or by
removing the card from the board.

## Deployment note

`PROJECT_BOARD_TOKEN` must exist as a repo secret (classic PAT with `read:project`, from an account
that can read project 306) before the nightly runs. 

## Testing

- Board query verified against project 306 by node ID: returns the three `Staging` cards, skips the
  draft (#28398), and keeps #27719 and #27975.
- Per-repo separation verified from the same board read: `facebookincubator/velox` resolves to `[]`
  and `prestodb/presto` to `[27719 27975]`, so neither path can see the other's cards.
- Empty-board handling verified both ways: auto-fetch exits 0 with `PR_COUNT=0`, while an empty manual
  list still exits 1.
- `staging.yml` cannot be dispatched until it is on `main`, since `workflow_dispatch` resolves workflow
  names against the default branch.

### DIFF
--- a/scripts/create_staging_branch.sh
+++ b/scripts/create_staging_branch.sh
@@ -30,6 +30,12 @@ MODE="local"
 STEP_NAME=""
 declare -A PR_SHA
 
+# NVIDIA libcudf project board used by Velox auto-fetch.
+# https://github.com/orgs/NVIDIA/projects/306/views/4
+VELOX_STAGING_PROJECT_OWNER="NVIDIA"
+VELOX_STAGING_PROJECT_NUMBER="306"
+VELOX_STAGING_PROJECT_QUERY="velox-staging:Staging is:pr is:open"
+
 log() { echo "$@" >&2; }
 die() { log "ERROR: $*"; exit 1; }
 STEP=0
@@ -121,10 +127,11 @@ Options:
   --base-branch branch             Base branch (default: ${BASE_BRANCH})
   --target-branch branch           Target branch (default: ${TARGET_BRANCH})
   --work-dir path                  Directory to clone target repo (default: ${WORK_DIR})
-  --auto-fetch-prs true|false      Auto-fetch non-draft PRs with label (default: ${AUTO_FETCH_PRS})
+  --auto-fetch-prs true|false      Auto-fetch open non-draft PRs from the NVIDIA project board
+                                   (https://github.com/orgs/NVIDIA/projects/306/views/4)
   --manual-pr-numbers "1,2,3"      Comma-separated PR numbers to merge (disables auto-fetch)
   --exclude-pr-numbers "4,5,6"    Comma-separated PR numbers to exclude from auto-fetch results
-  --pr-labels labels               Comma-separated PR labels to auto-fetch (default: ${PR_LABELS})
+  --pr-labels labels               If set, auto-fetch by GitHub labels instead of the project board
   --manifest-template path         Manifest template path (default: repo template)
   --force-push true|false          Force push to target branch (default: ${FORCE_PUSH})
   --additional-repository repo     Additional repository to merge from (e.g., rapidsai/cudf)
@@ -149,8 +156,7 @@ Examples:
   ./scripts/create_staging_branch.sh \\
     --target-path ../velox \\
     --base-repository facebookincubator/velox \\
-    --base-branch main \\
-    --pr-labels "cudf"
+    --base-branch main
 
   # Manual PR list (auto-fetch disabled automatically):
   ./scripts/create_staging_branch.sh \\
@@ -162,7 +168,6 @@ Examples:
   ./scripts/create_staging_branch.sh \\
     --target-path ../velox \\
     --base-repository facebookincubator/velox \\
-    --pr-labels "cudf" \\
     --additional-repository rapidsai/cudf \\
     --additional-branch velox-exchange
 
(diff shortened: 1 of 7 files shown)


karthikeyann (approved):
approving. 
I will wait for @bdice 's reviews and  approval too.

scripts/create_staging_branch.sh
```diff
@@ -30,6 +30,12 @@ MODE="local"
 STEP_NAME=""
 declare -A PR_SHA
 
+# NVIDIA libcudf project board used by Velox auto-fetch.
+# https://github.com/orgs/NVIDIA/projects/306/views/4
+VELOX_STAGING_PROJECT_OWNER="NVIDIA"
+VELOX_STAGING_PROJECT_NUMBER="306"
```
bdice, line 36: I think there's a unique ID for the project board that you can get with the GitHub API. That might be better to use as a long-term stable reference.

bdice (approved):

scripts/create_staging_branch.sh
Avinash-Raj: Thanks, Updated the relevant staging script with board specific unique id for fetching it's contents.

Issue #420, opened 2026-09-08
Automated staging pipeline: staging branches → published images → SF1K benchmark


karthikeyann:
## Goal

Get nightly integrated signal on in-flight Velox and Presto PRs, before they merge upstream — so we find breakage and measure performance on the combined set early instead of after the fact.

Three pieces:

**1. Staging branch creation in CI** — *Avinash*

Select the PRs we care about from a GitHub project board https://github.com/orgs/NVIDIA/projects/306/views/4 (a field on the board marks a PR as "Velox Staging"), merge them into a staging branch for Velox and Presto, and run it nightly. On merge conflict, fail and report.

We want the board as the control surface: moving a card is how someone adds or removes a PR from staging. No label access is needed on upstream repos.

use rapidsai/velox for staging branch
For presto, use karthikeyann/presto for now. IF zero staging PRs on presto, then use prestodb/presto master.

**2. Presto and Velox image build and publish** — *Matt*

Build from those staging branches and publish images under a stable staging tag, so anyone can pull the current staging state without rebuilding.

**3. SF1K benchmark automation** — *Karthik*

On a dedicated GPU machine, pull the published staging images and run TPC-H at SF1000 on a schedule, and publish the results so regressions are visible.

## Why

Today staging branches are created manually and nothing builds or benchmarks them. Integration problems between in-flight PRs surface late, and we have no continuous SF1K performance signal.

Avinash-Raj:
I’m currently blocked by this issue: https://github.com/rapidsai/ops/issues/4534#issuecomment-5589747974. The current `VELOX_FORK_PAT` secret doesn’t have the necessary scopes to read the Presto GPU board: https://github.com/orgs/NVIDIA/projects/306/views/4. 

It would be great if someone from the Ops team could help me with this. @rockhowse

Avinash-Raj:
Managed to overcome the above read issue by creating a new PAT token (owned) with `project::read` and `org::read` scopes and setting it as a secret (`PROJECT_BOARD_TOKEN`) on `velox-testing` repo.

Avinash-Raj:
1. Staging branch creation via velox-testing CI -> https://github.com/rapidsai/velox-testing/pull/422

PR #424, opened 2026-09-11
ci: integrate staging image builds with shared workflows

mattgara:
* Add an optional staging image workflow that builds Velox, the Presto native worker, and the Presto coordinator from configurable repositories and branches, tags, or commits using existing build workflows.

* Update existing Velox and Presto build workflows to optionally build one architecture and CUDA version, use staging tags, and add provenance without changing their default behavior.

* Publish staging images to the existing private ghcr.io/rapidsai/velox-testing-images package under the staging-images- prefix, and use existing 30-day cleanup to remove old staging builds.

* Store the source repositories, resolved full commits, and workflow details in every final image. Produce a build manifest and digest-based image references.

* Verify image package privacy, provenance, and optional GPU startup before updating stable staging tags.

---

## Notes

This implements the image build and publish portion of #420 and is intended to consume the staging branches created by #422.


karthikeyann:
Thanks @mattgara

karthikeyann:
I will wait for other reviewers to review the github workflow.

.github/workflows/staging-images.yml
```diff
@@ -0,0 +1,724 @@
...
+          PRESTO_SHORT_SHA: ${{ needs.resolve-sources.outputs.presto_short_sha }}
+          VELOX_SHORT_SHA: ${{ needs.resolve-sources.outputs.velox_short_sha }}
+        run: |
+          velox_image="${REGISTRY}/${IMAGE_NAME}:${TAG_PREFIX}velox-${VELOX_SHORT_SHA}-gpu-cuda${CUDA_VERSION}-${BUILD_DATE}-${BUILD_VARIANT}-${GITHUB_RUN_ID}"
+          presto_worker_image="${REGISTRY}/${IMAGE_NAME}:${TAG_PREFIX}presto-${PRESTO_SHORT_SHA}-velox-${VELOX_SHORT_SHA}-gpu-cuda${CUDA_VERSION}-${BUILD_DATE}-${BUILD_VARIANT}-${GITHUB_RUN_ID}"
+          presto_coordinator_image="${REGISTRY}/${IMAGE_NAME}:${TAG_PREFIX}presto-coordinator-${PRESTO_SHORT_SHA}-${BUILD_DATE}-${BUILD_VARIANT}-${GITHUB_RUN_ID}"
+
+          velox_digest="sha256:$(docker buildx imagetools inspect --raw "${velox_image}" | sha256sum | cut -d' ' -f1)"
+          presto_worker_digest="sha256:$(docker buildx imagetools inspect --raw "${presto_worker_image}" | sha256sum | cut -d' ' -f1)"
+          presto_coordinator_digest="sha256:$(docker buildx imagetools inspect --raw "${presto_coordinator_image}" | sha256sum | cut -d' ' -f1)"
```
patdevinwilson: This is the registry digest of the merged tag only if `--raw` is bit-identical to the stored blob (no trailing newline, no reformatting). `sha256sum` of that stream is a common source of `sha256:...` values that GHCR will not accept, and those values are then used for pull-by-digest, provenance checks, attestations, and the uploaded manifest.

Please take the digest from the registry instead, for example:

```bash
digest=$(docker buildx imagetools inspect --format '{{.Descriptor.Digest}}' "${velox_image}")
```

If that template is empty on the runner's buildx, parse the first `Digest:` line from the default inspect output — anything that echoes GHCR's descriptor rather than hashing CLI output. Then `docker pull "${image}@${digest}"` in the next step will actually prove the pin.

.github/workflows/velox-build.yml
```diff
@@ -312,29 +361,40 @@ jobs:
...
         uses: ./.github/actions/delete-arch-tags
         with:
           registry: ${{ env.REGISTRY }}
           image-name: ${{ env.IMAGE_NAME }}
           base-tags: |
-            velox-${{ inputs.velox_short_sha }}-gpu-cuda12.9-${{ inputs.date }}-${{ inputs.build_variant }}-${{ github.run_id }}
-            velox-${{ inputs.velox_short_sha }}-gpu-cuda13.1-${{ inputs.date }}-${{ inputs.build_variant }}-${{ github.run_id }}
-            velox-${{ inputs.velox_short_sha }}-cpu-${{ inputs.date }}-${{ inputs.build_variant }}-${{ github.run_id }}
+            ${{ inputs.tag_prefix }}velox-${{ inputs.velox_short_sha }}-gpu-cuda12.9-${{ inputs.date }}-${{ inputs.build_variant }}-${{ github.run_id }}
+            ${{ inputs.tag_prefix }}velox-${{ inputs.velox_short_sha }}-gpu-cuda13.1-${{ inputs.date }}-${{ inputs.build_variant }}-${{ github.run_id }}
```
patdevinwilson: `create_manifest` now honors `manifest_cuda_versions` / `manifest_architectures`, but cleanup still always deletes only the cuda12.9 and cuda13.1 (and cpu) arch tags. Staging currently sets `cleanup_arch_tags: false`, so this is not on the staging path, but a later caller that customizes the matrix with cleanup enabled will either miss tags or try to delete images that were never built.

Please drive these `base-tags` from the same CUDA/arch lists (same issue in `presto-build.yml`).

.github/workflows/staging-images.yml
```diff
@@ -0,0 +1,724 @@
...
+        run: |
+          docker buildx imagetools create \
+            --tag "${REGISTRY}/${IMAGE_NAME}:${TAG_PREFIX}velox-latest-gpu-cuda${CUDA_VERSION}-${ARCHITECTURE}" \
+            "${VELOX_IMAGE}"
+          docker buildx imagetools create \
+            --tag "${REGISTRY}/${IMAGE_NAME}:${TAG_PREFIX}presto-latest-gpu-cuda${CUDA_VERSION}-${ARCHITECTURE}" \
+            "${PRESTO_WORKER_IMAGE}"
+          docker buildx imagetools create \
+            --tag "${REGISTRY}/${IMAGE_NAME}:${TAG_PREFIX}presto-coordinator-latest-${ARCHITECTURE}" \
+            "${PRESTO_COORDINATOR_IMAGE}"
```
patdevinwilson: These three `imagetools create` calls are each tag-atomic, but not atomic as a set. If the coordinator retag fails, Velox/worker `...-latest-...` already point at this run and SF1K (or anyone pulling the three latest aliases) can mix builds.

Worth either retagging into a single shared stamp first and flipping the three aliases only after all `create`s succeed, or documenting that consumers should use the digest manifest artifact rather than the three latest tags as a bundle.

.github/workflows/staging-images.yml
mattgara: Updated this to read the manifest digest directly from imagetools inspect and reject malformed values. Each returned digest was also verified to resolve against GHCR.

.github/workflows/velox-build.yml
mattgara: Updated both workflows so cleanup follows the supplied CUDA version. An empty input retains the existing CUDA 12.9, CUDA 13.1, and CPU cleanup.

.github/workflows/staging-images.yml
mattgara: Documented the publish_stable_tags input as independent convenience aliases and directed consumers that need a matched set to the digest manifest artifact.

presto/docker/provenance_labels.dockerfile
```diff
@@ -19,3 +19,39 @@ RUN mkdir -p /opt/velox-testing && \
...
+      velox-testing.workflow.repository=${VELOX_TESTING_REPOSITORY} \
+      velox-testing.workflow.file=${WORKFLOW_FILE} \
+      velox-testing.workflow.run-id=${WORKFLOW_RUN_ID} \
+      velox-testing.workflow.run-attempt=${WORKFLOW_RUN_ATTEMPT}
+
+RUN python3 -c 'import json, sys; path = "/opt/velox-testing/provenance.json"; data = json.load(open(path)); keys = ["image_role", "architecture", "cuda_version", "build_created", "presto_ref", "velox_ref", "velox_testing_sha", "velox_testing_ref", "velox_testing_repo", "workflow_file", "workflow_run_id", "workflow_run_attempt", "workflow_run_url"]; data.update(zip(keys, sys.argv[1:])); data["schema_version"] = "1"; json.dump(data, open(path, "w"), indent=2, sort_keys=True)' \
+    "${IMAGE_ROLE}" "${ARCHITECTURE}" "${CUDA_VERSION}" "${BUILD_CREATED}" \
+    "${PRESTO_BRANCH}" "${VELOX_BRANCH}" "${VELOX_TESTING_SHA}" \
+    "${VELOX_TESTING_REF}" "${VELOX_TESTING_REPOSITORY}" "${WORKFLOW_FILE}" \
+    "${WORKFLOW_RUN_ID}" "${WORKFLOW_RUN_ATTEMPT}" "${WORKFLOW_RUN_URL}"
```
patdevinwilson, line 57: `ARG PRESTO_BRANCH` and `ARG VELOX_BRANCH` are declared only in the `basic` stage. They are not visible here, so these expansions are empty even when the action passes `--build-arg PRESTO_BRANCH=...`.

This `data.update(...)` then overwrites `presto_ref` / `velox_ref` in `provenance.json` with `""`. `verify-images` requires `.presto_ref` / `.velox_ref` to equal the requested refs, so worker and Velox images will fail verification.

Redeclare the source ARGs in this stage (at least `PRESTO_BRANCH` and `VELOX_BRANCH`; keep the SHA/repo fields from `basic` as you do now).

.github/workflows/staging-images.yml
```diff
@@ -0,0 +1,437 @@
...
+                and .presto_repo == $presto_repo
+                and .presto_ref == $presto_ref
+                and .presto_sha == $presto_sha
+                and .velox_repo == $velox_repo
+                and .velox_ref == $velox_ref
+                and .velox_sha == $velox_sha
+                and .velox_testing_repo == env.GITHUB_REPOSITORY
+                and .velox_testing_ref == env.GITHUB_REF_NAME
+                and .velox_testing_sha == env.GITHUB_SHA
+                and .workflow_file == env.GITHUB_WORKFLOW_REF
```
patdevinwilson, line 307: The image bakes `WORKFLOW_FILE=${{ github.workflow_ref }}` from inside the called `velox-build.yml` / `presto-build.yml` jobs. This check runs in `staging-images.yml` and compares against the parent `GITHUB_WORKFLOW_REF`.

Those strings are `.../velox-build.yml@...` vs `.../staging-images.yml@...`. Either pass the caller’s `github.workflow_ref` into the reusable workflows, or assert a prefix/contains check instead of equality.

patdevinwilson (changes requested):
Re-reviewed 953786b. The three earlier comments are addressed: registry digest via `imagetools inspect --format`, cleanup keyed off `architecture`/`cuda_version`, and latest tags documented as independent aliases.

Please fix the extended provenance stage before merge. `FROM basic AS extended` does not inherit `ARG PRESTO_BRANCH` / `ARG VELOX_BRANCH` from the previous stage, so the Python update overwrites `presto_ref`/`velox_ref` with empty strings. `verify-images` then checks those fields against the requested refs and will fail on the first real staging run. While you are there, confirm `workflow_file` baked from `github.workflow_ref` inside the called `velox-build`/`presto-build` workflows matches `GITHUB_WORKFLOW_REF` in the parent `staging-images` job — those are different files.

.github/workflows/staging-images.yml
patdevinwilson: Thanks — this addresses it. Digest now comes from `imagetools inspect --format '{{.Manifest.Digest}}'` and is rejected unless it matches `sha256:` + 64 hex.

.github/workflows/velox-build.yml
patdevinwilson: Thanks — this addresses it for the current input model. Cleanup runs only on the default dual-arch path, and `cuda_version` empty still deletes 12.9 / 13.1 / CPU to match the default matrix.

.github/workflows/staging-images.yml
patdevinwilson: Thanks — documenting `publish_stable_tags` as independent convenience aliases and pointing matched-set consumers at the digest manifest is enough.

karthikeyann (approved):
Approving this  to unblock.
I will wait for @bdice reviews too.

.github/actions/apply-provenance-labels/action.yml
```diff
@@ -56,6 +65,17 @@ runs:
...
+          CUDA_VERSION=${{ inputs.cuda_version }}
+          BUILD_CREATED=${{ inputs.build_created }}
+          VELOX_TESTING_SHA=${{ github.sha }}
+          VELOX_TESTING_REPOSITORY=${{ github.repository }}
+          VELOX_TESTING_REF=${{ github.ref_name }}
+          WORKFLOW_FILE=${{ github.workflow_ref }}
+          WORKFLOW_RUN_ID=${{ github.run_id }}
+          WORKFLOW_RUN_ATTEMPT=${{ github.run_attempt }}
+          WORKFLOW_RUN_URL=${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        outputs: type=registry,compression=zstd,force-compression=${{ inputs.build_created == '' }},compression-level=15
```
bdice, line 80: Why does `force-compression` vary here?

.github/workflows/presto-build.yml
```diff
@@ -41,9 +41,30 @@ on:
...
         required: false
         default: 'manual'
+      tag_prefix:
+        description: 'Optional prefix applied to image and cache tags'
+        type: string
+      architecture:
+        description: 'Optional architecture; empty builds amd64 and arm64'
+        type: string
+      cuda_version:
+        description: 'Optional CUDA version; empty builds CUDA 12.9, CUDA 13.1, and CPU images'
```
bdice:
```suggestion
        description: 'Optional CUDA version; empty builds CUDA 12, CUDA 13, and CPU images'
```

.github/workflows/presto-build.yml
```diff
@@ -58,15 +79,8 @@ jobs:
...
-          - arch: amd64
-            cuda_version: &cuda12 '12.9'
-          - arch: arm64
-            cuda_version: *cuda12
-          - arch: amd64
-            cuda_version: &cuda13 '13.1'
-          - arch: arm64
-            cuda_version: *cuda13
+        arch: &architecture_matrix ${{ fromJSON(inputs.architecture != '' && format('["{0}"]', inputs.architecture) || '["amd64","arm64"]') }}
+        cuda_version: ${{ fromJSON(inputs.cuda_version != '' && format('["{0}"]', inputs.cuda_version) || '["12.9","13.1"]') }}
```
bdice: We need to reduce the number of places we hardcode the CUDA versions to a minimum.

That will help us with future CUDA minor version bumps.

presto/docker/provenance_labels.dockerfile
mattgara: Thanks for flagging this. I double-checked it against the documented Docker scope rules and the built image. Because `extended` uses `FROM basic AS extended`, it is based on `basic`; Docker documents that an `ARG` declared in a build stage is inherited by stages based on it: https://docs.docker.com/reference/dockerfile/#scope

I also built this Dockerfile locally with non-empty `PRESTO_BRANCH` and `VELOX_BRANCH`, and both refs were present in the final JSON. The exact current-commit run then passed the strict ref equality checks for each image: https://github.com/rapidsai/velox-testing/actions/runs/34666758132/job/103489486702. Those checks would fail if either expected ref were empty.

Given that, I believe the current inheritance is working as intended and keeping it avoids duplicate declarations. Please let me know if I am missing a different stage boundary here.

.github/workflows/staging-images.yml
mattgara: Thanks, I checked this as well. GitHub documents that the `github` context in a reusable workflow is associated with the caller workflow: https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#github-context

In the exact run, the baked value and uploaded manifest recorded `rapidsai/velox-testing/.github/workflows/staging-images.yml@refs/heads/feat/staging-image-integration-420`. The strict `.workflow_file == env.GITHUB_WORKFLOW_REF` check passed for all three images, and the generated attestations also identify `staging-images.yml`: https://github.com/rapidsai/velox-testing/actions/runs/34666758132/job/103489486702.

I believe exact equality is therefore appropriate here without another input. Please let me know if there is a different reusable-workflow context case I have missed.

patdevinwilson (approved):
Re-reviewed with the dispatch run and docs you linked. I was wrong on both remaining comments.

Docker does inherit `ARG`s into stages based on the declaring stage ([ARG scope](https://docs.docker.com/reference/dockerfile/#scope)), and the `github` context in a reusable workflow is the caller’s ([reusable workflows](https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#github-context)). That matches [the verify job](https://github.com/rapidsai/velox-testing/actions/runs/34666758132/job/103489486702) succeeding, including digest pins, provenance equality, and GPU smoke.

LGTM.

presto/docker/provenance_labels.dockerfile
patdevinwilson: You’re right — I had the ARG scope wrong. Docker inherits an `ARG` declared in a stage into later stages based on it ([docs](https://docs.docker.com/reference/dockerfile/#scope)), and the passing verify job is the proof. Thanks for checking it against a real build.

.github/workflows/staging-images.yml
patdevinwilson: Agreed. The caller’s `github` context is what the reusable workflow sees ([docs](https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#github-context)), and the verify job recorded `staging-images.yml`. Equality is the right check.

.github/workflows/velox-build.yml
```diff
@@ -29,9 +29,27 @@ on:
...
         required: false
         default: 'manual'
+      tag_prefix:
+        description: 'Optional prefix applied to image and cache tags'
+        type: string
+      architecture:
+        description: 'Optional architecture; empty builds amd64 and arm64'
+        type: string
+      cuda_version:
+        description: 'Optional CUDA version; empty builds CUDA 12.9, CUDA 13.1, and CPU images'
```
bdice:
```suggestion
        description: 'Optional CUDA version; empty builds CUDA 12, CUDA 13, and CPU images'
```

bdice (approved):

.github/actions/apply-provenance-labels/action.yml
mattgara: Good question. The basic path keeps the existing behavior. The extended path is new, and its base image has already been exported with zstd, so I believe forcing it again would only recompress those layers. I added a short comment here.

.github/workflows/presto-build.yml
mattgara: Agreed. I now reuse one build list per workflow and derive the cleanup tags from the manifests that were created. The architecture checks only use the CUDA major.

PR #427, opened 2026-09-18
Presto Slurm: personal-account fixes — fast-fail exit, mmap-allocator override, quieter logs

misiugodfrey:
## Summary

The main goal here was to get the presto slurm scripts working with the new user accounts on the NVL72.  There were some QOL improvements added as well.

Fixes found while switching the Presto NVL72 Slurm scripts over from an
always-`prestouser` assumption to support personal accounts, and while
debugging benchmark runs on that path:

- **`launch-run.sh`**: the "Resolving first node IP" loop used to poll
  `squeue` for up to 5 minutes even when the job had already failed and left
  the queue (e.g. a fast failure during `setup` — missing Hive schema, etc).
  It now checks job state each iteration and breaks immediately once the job
  is gone, instead of waiting out the full timeout before reporting failure.
- **New `CLUSTER_GPU_USE_MMAP_ALLOCATOR` / `CLUSTER_CPU_USE_MMAP_ALLOCATOR`
  cluster-config override**, wired through `launcher_common.sh` →
  `build_common_export_vars` → `generate_presto_config.sh`, which now sets
  `use-mmap-allocator` in the generated native worker config. Some
  GB200/ARM nodes use a 64KB kernel page size, and Velox's `MmapAllocator`
  (the Presto C++ default) asserts at startup on those hosts
  (`isPageSizeSupported()`, see `MmapAllocator.cpp:76`). Setting the new
  variable to `false` switches to `MallocAllocator` without touching any
  launcher scripts. Defaults to `true` everywhere else, so clusters with
  normal 4KB pages are unaffected.
- **Quieter job logs**: dropped `-x` shell tracing from the two job
  entrypoints (`run-presto-benchmarks.sh`, `run-analyze-tables.sh`), which
  was dumping every executed command into `.err` logs. Also silenced
  miniforge/venv/conda/pip setup noise (`run_py_script.sh --quiet`,
  captured-and-dumped-on-failure for the miniforge installer) and stray
  `pushd`/`popd` directory-stack echoes.
- **Color consistency**: centralized `echo_success`/`RED` color helpers into
  `launcher_common.sh` and applied them to the main pass/fail lines
  (`Job completed` / `Job FAILED`, `Analyze tables complete!`,
  `Benchmark complete!`, etc.), matching the existing green "Configs were
  generated successfully" convention.
- **Docs/comments fix**: every script comment and the README referenced
  `~/.cluster_config.env`, but the actual default (`defaults.env`) has
  always been `~/presto_cluster_config.env` — a stale, misleading path.
  Fixed everywhere, and added guidance to copy a cluster-wide default
  (e.g. `/scratch/nvidia/presto_cluster_config.env`) instead of starting
  from the blank template when one has already been published.

## Outstanding issues

Testing showed that this worked for existing tpch-float workloads, but we ran into errors when testing on decimal.  This does not seem to be an issue with the scripts, as it's a segfault in velox.  Further investigation will be needed on that end, but otherwise we are working as expected.

The default data directory still points at the existing data, where new users have read-only access - this is as intended for now.  If we want to create new "official" data for now we will use `prestouser`, otherwise you can creat your own data and point the configs at it.

## Test plan

- [x] Reproduced the original "hangs for 5 minutes after a fast failure"
      behavior, confirmed `launch-run.sh` now exits promptly once the job
      leaves the queue.
- [x] Reproduced the `MmapAllocator` startup crash on a 64KB-page GB200 node;
      confirmed `CLUSTER_GPU_USE_MMAP_ALLOCATOR=false` avoids it and workers
      register successfully.
- [x] Verified the `use-mmap-allocator` `sed` substitution against the real
      template file.
- [x] Full end-to-end TPC-H SF1000 benchmark run (blocked on a separate,
      unrelated GPU-exchange correctness bug under investigation — not part
      of this PR).
- [x] Ran launch-analyze.sh and launch-run.sh on tpch-float and tpch-decimal workloads (1k, 3k).  Worked as expected with the exception of the decimal benchmarks segfaulting in velox (as noted above).


presto/docker/config/template/etc_worker/config_native.properties
```diff
@@ -33,3 +33,9 @@ system-mem-shrink-gb=20
 single-node-execution-enabled=true
 
 async-data-cache-enabled=false
+
+# Use Velox's mmap-based allocator (the Presto C++ default). MmapAllocator
```
misiugodfrey, line 37: This was a velox error that I ran into while testing with newer images.  I added this config as a workaround that should only apply to the relevant cluster - we can remove it once it's sorted, but I think it's necessary right now to run modern images.

presto/scripts/analyze_tables.sh
```diff
@@ -111,4 +111,4 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 ANALYZE_TABLES_SCRIPT_PATH="${SCRIPT_DIR}/../testing/integration_tests/analyze_tables.py"
 REQUIREMENTS_PATH="${SCRIPT_DIR}/../testing/requirements.txt"
 
-"${SCRIPT_DIR}/../../scripts/run_py_script.sh" -p "$ANALYZE_TABLES_SCRIPT_PATH" -r "$REQUIREMENTS_PATH" "${SCRIPT_ARGS[@]}"
+"${SCRIPT_DIR}/../../scripts/run_py_script.sh" --quiet -p "$ANALYZE_TABLES_SCRIPT_PATH" -r "$REQUIREMENTS_PATH" "${SCRIPT_ARGS[@]}"
```
misiugodfrey, line 114: Small QoL improvement to clean up the logs as they were full of noise.

presto/slurm/presto-nvl72/launch-run.sh
```diff
@@ -204,6 +206,12 @@ ERR_FILE="${ERR_FMT//%j/${JOB_ID}}"
 echo "Resolving first node IP..."
 for i in {1..60}; do
     STATE=$(squeue -j "$JOB_ID" -h -o "%T" 2>/dev/null || true)
+    if [[ -z "${STATE}" ]]; then
```
misiugodfrey, line 209: This was a cleanup to stop the script from waiting for a long time if the job failed at an early state.

presto/slurm/presto-nvl72/launch-run.sh
```diff
@@ -245,5 +253,5 @@ if [[ -n "${OUTPUT_PATH}" ]]; then
     echo "Copying results to ${OUTPUT_PATH}..."
     mkdir -p "${OUTPUT_PATH}"
     cp -r result_dir/. "${OUTPUT_PATH}/"
-    echo "Results copied to ${OUTPUT_PATH}"
+    echo_success "Results copied to ${OUTPUT_PATH}"
```
misiugodfrey, line 255: These are some small QoL changes to make the logs/results more quickly parse-able with major successes in green, and major failures in red.

presto/slurm/presto-nvl72/run-analyze-tables.sh
```diff
@@ -2,7 +2,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-set -exuo pipefail
```
misiugodfrey, line 5: Removed -x here because it was making the logs extremely noisy from all the "source" commands.

scripts/py_env_functions.sh
```diff
@@ -10,10 +10,16 @@ function install_miniforge3() {
   TEMP_DIR=$(mktemp -d)
   trap "rm -rf $TEMP_DIR" EXIT
 
-  pushd $TEMP_DIR
-  curl -L -O "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh"
-  bash Miniforge3-$(uname)-$(uname -m).sh -b -p "$MINIFORGE_HOME"
-  popd
+  local log_file="$TEMP_DIR/install.log"
+  pushd $TEMP_DIR >/dev/null
+  if ! { curl -fsSL -O "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh" \
```
misiugodfrey, line 15: Cleanup the miniforge logging so that it only prints the logs  if something goes  wrong.  Otherwise we have hundreds of "installing..., linking..." messages that are not useful.

scripts/run_py_script.sh
```diff
@@ -103,7 +103,7 @@ fi
 STAMP_FILE=".venv/.requirements_stamp"
 if [[ ! -f "$STAMP_FILE" ]] || ! diff -q "$REQUIREMENTS_FILE_PATH" "$STAMP_FILE" &>/dev/null; then
   log "Running pip install for requirements file: $REQUIREMENTS_FILE_PATH"
-  pip install -q -r $REQUIREMENTS_FILE_PATH
+  pip install -q --root-user-action=ignore -r $REQUIREMENTS_FILE_PATH
```
misiugodfrey, line 106: Suppresses a known warning that should not be relevant since we are running in a container.

presto/slurm/presto-nvl72/README.md
```diff
@@ -22,11 +22,22 @@ print the command needed to satisfy it.
 
 ## First-time setup
 
-1. **Cluster config.** Copy the template and fill in your cluster's values:
+1. **Cluster config.** If someone has already published a cluster-wide default
+   (e.g. `/scratch/nvidia/presto_cluster_config.env` on this cluster), start
```
misiugodfrey, line 26: This is probably the most important bit.  Everyone using the shared account was using a shared `~/.cluster_config.env` and now each user needs to create their own.  They can just grab an updated shared one with all the defaults that the given location, or they can create their own.

presto/docker/config/template/etc_worker/config_native.properties
bdice: See discussions on https://github.com/facebookincubator/velox/pull/18618 for more info on this.

presto/slurm/presto-nvl72/cluster_config.env.example
```diff
@@ -97,6 +102,12 @@ CLUSTER_GPU_USE_NUMA=1
 # Example: 2 for a node with 4 GPUs across 2 NUMA domains.
 CLUSTER_GPU_NUMA_GPUS_PER_NODE=<gpus-per-numa-node>
 
+# Velox native worker allocator: true = MmapAllocator (Velox/Presto C++ default),
+# false = MallocAllocator. MmapAllocator asserts at startup on hosts with a
+# non-4KB kernel page size (e.g. 64KB pages on some ARM/GB200 systems) — set to
+# false there. See "system page size is 65536 bytes" in a worker's crash log.
```
bdice, line 108: Maybe cross-link https://github.com/facebookincubator/velox/pull/18618 here.

bdice (approved):

presto/docker/config/template/etc_worker/config_native.properties
misiugodfrey: Added — linked both the assertion PR (facebookincubator/velox#18618) and the tracking issue for the real fix (facebookincubator/velox#18617) in a comment above this line.

presto/slurm/presto-nvl72/cluster_config.env.example
misiugodfrey: Done, linked both facebookincubator/velox#18618 and facebookincubator/velox#18617 here as well.

